PR #1, opened 2018-09-10
setup installer manifest content

bparees:


bparees:
@smarterclayton ptal

manifest/imagestream.yaml
```diff
@@ -0,0 +1,8 @@
+kind: ImageStream
```
smarterclayton: rename to `image-references` (entire filename)

bparees:
@legionus fyi

smarterclayton:
Let's try this out

/lgtm

PR #2, opened 2018-09-10
Fix case

bparees:
@smarterclayton as per your request


smarterclayton:
https://github.com/openshift/origin/pull/20926 correctly failed because of this (yay!).

/lgtm

PR #3, opened 2018-09-12
fix hack script to use makefile to build

bparees:


hack/deploy.sh
```diff
@@ -4,7 +4,7 @@ cd "$(dirname "$0")/.."
 CURRENT_CONTEXT=$(oc config current-context)
 SYSTEM_ADMIN_CONTEXT=${CURRENT_CONTEXT%/*}/system:admin
 
-operator-sdk build docker.io/openshift/cluster-image-registry-operator:latest
```
bparees: fyi this was misaligned w/ the deploy/operator.yaml which uses "origin-cluster-image-registry-operator" as the image name now.

bparees:
@legionus ptal

PR #6, opened 2018-09-18
Prepare repo for distgit conventions

dmage:


bparees:
@legionus @dmage we should not need a spec file, we will not be building RPMs for the operators.

@gabemontero did you run into this?

PR #8, opened 2018-09-18
deploy: add ordering to manifests that end up in updatepayload

abhinavdahiya:
the current update payload looks like
```console
ls -lah /tmp/release-manifests/release-manifests
99_cluster-image-registry-operator_crd.yaml
99_cluster-image-registry-operator_namespace.yaml
99_cluster-image-registry-operator_operator.yaml
99_cluster-image-registry-operator_rbac.yaml
99_cluster-samples-operator_crd.yaml
99_cluster-samples-operator_namespace.yaml
99_cluster-samples-operator_operator.yaml
99_cluster-samples-operator_rbac.yaml
...
image-references
```

ClusterVersionOperator is waiting on `99_cluster-image-registry-operator_operator.yaml` to succeed,
but the pods crashloop because rbac is not setup and
`99_cluster-image-registry-operator_rbac.yaml` is next in order to the deployment...

This should create an update payload like:
```console
ls -lah /tmp/release-manifests/release-manifests
99_cluster-image-registry-operator_00-crd.yaml
99_cluster-image-registry-operator_01-namespace.yaml
99_cluster-image-registry-operator_02-rbac.yaml
99_cluster-image-registry-operator_03-operator.yaml
...
image-references
```


abhinavdahiya:
cc @smarterclayton

bparees:
@abhinavdahiya can you fix the references in hack/deploy.sh as well?

Dockerfile
```diff
@@ -9,5 +9,5 @@ RUN useradd cluster-image-registry-operator
 USER cluster-image-registry-operator
 
 COPY --from=0 /go/src/github.com/openshift/cluster-image-registry-operator /usr/bin
-COPY deploy/image-references deploy/crd.yaml deploy/namespace.yaml deploy/operator.yaml deploy/rbac.yaml /manifests/
+COPY deploy/image-references deploy/00crd.yaml deploy/01namespace.yaml deploy/03operator.yaml deploy/02rbac.yaml /manifests/
```
bparees: can we make these 00-crd.yaml, etc for better readability?

abhinavdahiya: https://github.com/openshift/origin/blob/256f7eaa7c5a718b95f4a2c1384a540eda67f8d6/pkg/oc/cli/admin/release/new.go#L488

is looking for `<number>_<whatever>` to decide if `99_<operator-name>_` should be prepended to filename. So i think `00-crd.yaml` should be fine.

abhinavdahiya: done

smarterclayton:
This seems like the wrong ordering.  I expected the other ordering - by doing this bens things go first, even before Kube operators.

If you want internal ordering you need to do “prefix_numbers_rest”

smarterclayton:
Reverting temporarily so we can discuss.

bparees:
> This seems like the wrong ordering. I expected the other ordering - by doing this bens things go first, even before Kube operators.

i read @abhinavdahiya's comment as saying these names are getting prefixed with "99_", no?

PR #11, opened 2018-09-19
Add simple completion of GCS configuration

legionus:


bparees:
@legionus please split your vendoring commit out to make this reviewable.

@coreydaley fyi, you should compare notes w/ what you're doing for s3 to make sure we have the same general pattern/flow.

legionus:
@bparees Done

pkg/operator/bootstrap.go
```diff
@@ -54,5 +62,25 @@ func (h *Handler) bootstrap() error {
...
 	}
 
+	if len(cr.Spec.HTTPSecret) == 0 {
+		var secretBytes [randomSecretSize]byte
+		if _, err := rand.Read(secretBytes[:]); err != nil {
+			return fmt.Errorf("could not generate random bytes for HTTP secret: %s", err)
+		}
+		cr.Spec.HTTPSecret = fmt.Sprintf("%x", string(secretBytes[:]))
+
+		logrus.Warn("No HTTP secret provided - generated random secret")
```
dmage: How should it be provided?

legionus: @dmage Thanks! Removed.

coreydaley:
@bparees will do

pkg/operator/bootstrap.go
```diff
@@ -54,5 +60,23 @@ func (h *Handler) bootstrap() error {
 		Replicas: 1,
 	}
 
+	if len(cr.Spec.HTTPSecret) == 0 {
+		var secretBytes [randomSecretSize]byte
+		if _, err := rand.Read(secretBytes[:]); err != nil {
+			return fmt.Errorf("could not generate random bytes for HTTP secret: %s", err)
+		}
+		cr.Spec.HTTPSecret = fmt.Sprintf("%x", string(secretBytes[:]))
```
bparees: what is this used for?

pkg/storage/gcs/gcs.go
```diff
@@ -33,5 +38,25 @@ func (d *driver) Volumes() ([]corev1.Volume, []corev1.VolumeMount, error) {
...
+		}
+
+		ctx := context.Background()
+
+		client, err := storage.NewClient(ctx)
+		if err != nil {
+			return err
+		}
+
+		err = client.Bucket(d.Config.Bucket).Create(ctx, projectID, nil)
```
bparees: do you not need to check if the bucket already exists?

pkg/operator/bootstrap.go
legionus: https://docs.docker.com/registry/configuration/#http

> A random piece of data used to sign state that may be stored with the client to protect against tampering. For production environments you should generate a random piece of data using a cryptographically secure random generator. If you omit the secret, the registry will automatically generate a secret when it starts. If you are building a cluster of registries behind a load balancer, you MUST ensure the secret is the same for all registries.

bparees: ok thanks

pkg/storage/gcs/gcs.go
legionus: That is why it's a simple gcs completion. I do not know how to do it better ))

bparees: ok. So you're still investigating how to check if a bucket exists?

legionus: @bparees Yes, I wanted to leave it as follow up. This PR is just a preview to understand how to update the configuration, when to do it, and how the function might looks like.

bparees: that's reasonable

PR #12, opened 2018-09-21
Log operator version

dmage:


Dockerfile
```diff
@@ -1,13 +1,10 @@
...
+COPY . .
+RUN make build
 
 FROM centos:7
-
 RUN useradd cluster-image-registry-operator
 USER cluster-image-registry-operator
-
-COPY --from=0 /go/src/github.com/openshift/cluster-image-registry-operator /usr/bin
-COPY deploy/image-references deploy/00-crd.yaml deploy/01-namespace.yaml deploy/03-operator.yaml deploy/02-rbac.yaml /manifests/
```
bparees: why is this removed?  it's necessary for the installer to process our operator image correctly.

Makefile
```diff
@@ -9,7 +9,7 @@ generate:
 	operator-sdk generate k8s
 
 build:
-	go build "./cmd/$(PROG)"
+	go build -ldflags "-X github.com/openshift/cluster-image-registry-operator/version.Version=$$(git describe --tags --always --dirty)" "./cmd/$(PROG)"
```
bparees: i'm not sure if this will be viable when this is being built out of dist-git, but we can cross that bridge when we come to it.

Dockerfile
dmage: Because these files aren't used by the operator. I added a note that these files are for the installer.

PR #13, opened 2018-09-24
Save immutable config parameters into configmap

legionus:


Dockerfile-dev
```diff
@@ -4,6 +4,6 @@ RUN useradd cluster-image-registry-operator
 USER cluster-image-registry-operator
 
 COPY cluster-image-registry-operator /usr/bin
-COPY deploy/image-references deploy/crd.yaml deploy/namespace.yaml deploy/operator.yaml deploy/rbac.yaml /manifests/
+COPY deploy/image-references deploy/*.yaml /manifests/
```
bparees: does the dev image need this at all?  fundamentally if you're not going to use the image w/ the installer, the /manifests dir doesn't matter.

bparees:
/lgtm

you can remove the COPY command in a follow up if you want, or leave it there for consistency.

Dockerfile-dev
legionus: @bparees I added this just in case. Suddenly someday in the future it will be needed. This is not difficult to copy.

bparees: ok, that's fine

PR #14, opened 2018-09-25
Add stub for end-to-end tests

dmage:


Makefile
```diff
@@ -9,16 +9,19 @@ generate:
...
-build-devel-image: build
-	docker build -t "$(IMAGE)" -f Dockerfile-dev .
+build-devel-image:
+	operator-sdk build "$(IMAGE)"
 
 test:
 	go test ./...
 
+test-e2e:
+	operator-sdk test --test-location=./test/e2e
```
bparees: what does this actually require to run?  does it need a docker socket?

cmd/cluster-image-registry-operator/main.go
```diff
@@ -53,7 +54,7 @@ func main() {
 	watch(corev1.SchemeGroupVersion.String(), "Secret", namespace, 0)
 	watch(kappsapi.SchemeGroupVersion.String(), "Deployment", namespace, 0)
 	watch(appsapi.SchemeGroupVersion.String(), "DeploymentConfig", namespace, 0)
-	watch(regopapi.SchemeGroupVersion.String(), "OpenShiftDockerRegistry", namespace, 5)
+	watch(regopapi.SchemeGroupVersion.String(), "OpenShiftDockerRegistry", namespace, 5*time.Second)
```
bparees: these values should be higher, like 10 minutes.

legionus:
LGTM

Makefile
dmage: It requires kubeconfig with admin capabilities (it is going to deploy the operator and create namespaces).

Also it requires [deploy/sa.yaml, deploy/rbac.yaml and deploy/operator.yaml](https://github.com/operator-framework/operator-sdk/blob/d23344453ec6124f1555fafc2e66230f98410bbb/commands/operator-sdk/cmd/test.go#L58) or a special file with resources that should be created in test namespaces.

docker socket isn't required.

bparees: ok

cmd/cluster-image-registry-operator/main.go
legionus: @bparees it will break operator logic.

legionus: @bparees it's not about finalizer. it's about objects re-creation and upgrade.

bparees: We get updates when objects change, we get delete events when they are deleted, we can look for objects at startup.  What is your scenario that requires you to poll every 5s?

PR #24, opened 2018-09-27
deploy: add namespace for the CR.

abhinavdahiya:
`imageregistries.imageregistry.operator.openshift.io` CRD is Namespaced.
But the cr.yaml doesn't have namespace setup. Cluster Version Operator
will not have context to decide which namespace the CR shouldbe installed in.

### DIFF
--- a/deploy/cr.yaml
+++ b/deploy/cr.yaml
@@ -3,6 +3,7 @@ apiVersion: imageregistry.operator.openshift.io/v1alpha1
 kind: ImageRegistry
 metadata:
   name: image-registry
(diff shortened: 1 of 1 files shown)


bparees:
@abhinavdahiya we  shouldn't be including the CR in our image /manifests, our operator creates its own default CR when it starts.

bparees:
i'll merge this since it's not wrong, per se, but i'm going to remove the cr from our manifest here:
https://github.com/openshift/cluster-image-registry-operator/pull/25

/lgtm

PR #25, opened 2018-09-27
don't include cr in manifests

bparees:


bparees:
@legionus @dmage fyi, we dont want the installer to install the cr.

PR #26, opened 2018-09-28
Watch for all the necessary resources and re-create them if necessary

legionus:


cmd/cluster-image-registry-operator/main.go
```diff
@@ -22,6 +25,8 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
 
+var logLevel = flag.String("loglevel", "info", "sets the sensitivity of logging output.")
```
bparees: `log-level` please, so we're consistent with everywhere else.

pkg/generate/route.go
```diff
@@ -59,7 +59,7 @@ func Route(cr *regopapi.ImageRegistry, route *regopapi.ImageRegistryConfigRoute,
 			},
 		},
 	}
-	if cr.Spec.TLS {
+	if len(route.SecretName) > 0 {
```
bparees: as discussed this morning, i think this was correct as is.  If the backend registry is using TLS, the route should be "re-encrypt" and if the backend registry is not using TLS, the route should be "Edge termination".

Whether the route is given a certificate or not is up to the user (whether they want to supply one or not).

bparees: well maybe not correct as is, but the logic should be:

1) if backend TLS, create reencrypt route
2) if backend !TLS, create edge termination route
3) either way, if user provided a route.SecretName, add TLS certificates to the route

pkg/operator/bootstrap.go
```diff
@@ -24,44 +24,57 @@ const randomSecretSize = 64
...
-				VolumeSource: corev1.VolumeSource{
-					PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
-						ClaimName: "image-registry-pvc",
+		Spec: regopapi.ImageRegistrySpec{
+			OperatorSpec: operatorapi.OperatorSpec{
+				ManagementState: operatorapi.Managed,
+				Version:         "none",
+				ImagePullSpec:   "docker.io/openshift/origin-docker-registry",
+			},
+			Storage: regopapi.ImageRegistryConfigStorage{
```
bparees: so as discussed this morning, please change this to an empty configstorage and make sure the operator reports a reasonable error when it sees an empty configstorage.

And when we actually have enough information to make a good decision about what the initial storage configuration should be, we'll do that.

bparees: (you can do that in a followup, I don't want to confuse this PR)

pkg/operator/handler.go
```diff
@@ -137,33 +141,77 @@ func conditionDeployment(cr *regopapi.ImageRegistry, status operatorapi.Conditio
...
+	if statusChanged {
+		logrus.Debugf("resource %s is re-recreated", tmpl.Name())
+		conditionResourceApply(cr, operatorapi.ConditionTrue, "all resources applied", &statusChanged)
 	}
 
-	return h.reRollout(cr), nil
+	return cr, statusChanged, nil
+}
+
+func (h *Handler) reRollByEvent(event sdk.Event, gen generate.Generator) (*regopapi.ImageRegistry, bool, error) {
```
bparees: "reroll"?  I don't understand what that means.

pkg/operator/handler.go
```diff
@@ -195,6 +311,7 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 				return err
 			}
 
+			logrus.Debugf("resource %s is re-recreated", tmpl.Name())
```
bparees: re-recreated? recreated?

pkg/operator/handler.go
```diff
@@ -225,6 +342,7 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
```
bparees: you mean recreated?

bparees:
is this supposed to be the finalizer implementation or is that in another PR?  Because i don't see a finalizer here.

cmd/cluster-image-registry-operator/main.go
dmage: oc cli uses `loglevel`: https://github.com/openshift/origin/blob/e004e6513ec755a7a106dc97516703847e470d0b/pkg/cmd/flagtypes/glog.go#L16

bparees: hm. i swear i saw --log-level elsewhere, but ok.

dmage: installer and cluster-operator use `--log-level`.

cmd/cluster-image-registry-operator/main.go
```diff
@@ -52,9 +66,15 @@ func main() {
 	}
 
 	k8sutil.AddToSDKScheme(appsapi.AddToScheme)
+	k8sutil.AddToSDKScheme(authapi.AddToScheme)
+
+	watch("rbac.authorization.k8s.io/v1beta1", "ClusterRole", "", 0)
+	watch("rbac.authorization.k8s.io/v1beta1", "ClusterRoleBinding", "", 0)
```
dmage: Shouldn't it be authapi.SchemeGroupVersion.String()?

cmd/cluster-image-registry-operator/main.go
bparees: that must be where i saw it. anyway --loglevel is fine since origin is using it, sorry for the noise.

cmd/cluster-image-registry-operator/main.go
```diff
@@ -37,6 +42,15 @@ func watch(apiVersion, kind, namespace string, resyncPeriod time.Duration) {
 func main() {
 	flag.Parse()
 
+	if logLevel == nil {
+		*logLevel = os.Getenv("LOG_LEVEL")
+	}
+	lvl, err := logrus.ParseLevel(*logLevel)
```
dmage: I would expect command line arguments to have precedence over environment variables. Though, the default value for the command line argument shouldn't be taken into account.

pkg/generate/route.go
```diff
@@ -27,13 +27,13 @@ func DefaultRoute(cr *regopapi.ImageRegistry, p *parameters.Globals) (Template,
...
-	if cr.Spec.TLS {
-		// TLS certificates are served by the front end of the router, so they must be configured into the route,
-		// otherwise the router's default certificate will be used for TLS termination.
-		r.Spec.TLS = &routeapi.TLSConfig{
-			Termination: routeapi.TLSTerminationReencrypt,
-		}
+
+	// TLS certificates are served by the front end of the router, so they must be configured into the route,
+	// otherwise the router's default certificate will be used for TLS termination.
+	r.Spec.TLS = &routeapi.TLSConfig{
```
dmage: So, this change makes impossible to deploy the registry that doesn't secured by TLS?

dmage: Perhaps we should always create cluster signed certificates and remove this option.

bparees:
> We should either remove the TLS option from our CR or create insecure routes.

The logic here was:

1) by default we should be using TLS for the registry service, using service serving certificate generation
2) we want to be able to disable TLS so that when we go to adopt an existing registry that has TLS turned off, we have a way to represent that configuration setting in the registry config resourse

dmage:
@bparees can we just enable TLS for them? It shouldn't be a breaking change, an insecure registry and a registry with self-signed certificates should be treated roughly the same.

cmd/cluster-image-registry-operator/main.go
legionus: NO!

legionus:
@bparees @dmage please review again

cmd/cluster-image-registry-operator/main.go
bparees: ?

pkg/generate/route.go
```diff
@@ -17,23 +17,25 @@ func DefaultRoute(cr *regopapi.ImageRegistry, p *parameters.Globals) (Template,
...
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      p.DefaultRoute.Name,
+			Name:      cr.ObjectMeta.Name + "-default-route",
 			Namespace: p.Deployment.Namespace,
 		},
 		Spec: routeapi.RouteSpec{
 			To: routeapi.RouteTargetReference{
 				Kind: "Service",
-				Name: p.Deployment.Name,
+				Name: cr.ObjectMeta.Name,
```
bparees: we expect this service to a have a fixed name, right?  (currently docker-registry)

pkg/generate/route.go
```diff
@@ -55,15 +57,21 @@ func Route(cr *regopapi.ImageRegistry, route *regopapi.ImageRegistryConfigRoute,
 			Host: route.Hostname,
 			To: routeapi.RouteTargetReference{
 				Kind: "Service",
-				Name: p.Deployment.Name,
+				Name: cr.ObjectMeta.Name,
```
bparees: service name should be constant?

pkg/generate/route.go
```diff
@@ -17,23 +17,25 @@ func DefaultRoute(cr *regopapi.ImageRegistry, p *parameters.Globals) (Template,
...
+	// TLS certificates are served by the front end of the router, so they must be configured into the route,
+	// otherwise the router's default certificate will be used for TLS termination.
 	if cr.Spec.TLS {
-		// TLS certificates are served by the front end of the router, so they must be configured into the route,
-		// otherwise the router's default certificate will be used for TLS termination.
-		r.Spec.TLS = &routeapi.TLSConfig{
-			Termination: routeapi.TLSTerminationReencrypt,
-		}
+		r.Spec.TLS.Termination = routeapi.TLSTerminationReencrypt
 	}
```
bparees: should be setting the termination to edge otherwise?

legionus: @bparees As we discussed today for the default route, we always use reencrypt. no ?

pkg/generate/route.go
legionus: Yep, but only for legacy deployments in the "default" namespace. no ?

pkg/generate/route.go
bparees: no, since it's still possible for someone to turn off TLS on the registry service, so we still have to be able to use edge termination on the default route.

pkg/generate/route.go
bparees: I would expect it to be a fixed name for new installs too (though possibly a new name, like "image-registry")?

bparees: I don't want to worry that someone manages to create the CR with some new name before the operator creates the CR, and now we have to deal with a cluster that has a random name for the registry service/etc.  Is there a good reason *not* to hardcode the name of these things, given, as you note, that we're only allowing one instance of them anyway?

legionus: > Is there a good reason not to hardcode the name of these things, given, as you note, that we're only allowing one instance of them anyway?

I wanted to remove hardcode the name from everywhere. I talked with @dmage about tests. This should help us in testing. After this commit we have only one global collection - images, but if you use random generated images with no blob reuse this should not be a problem.

But the service can be called the same everywhere. I will hardcode this name again.

dmage: @bparees to solve this problem we can hardcode names for the CR (default/docker-registry and openshift-image-registry/image-registry). In this case modifications for adoption will be minimal. Old installations will still have `docker-registry.default.svc`, the new one will have `image-registry.openshift-image-registry.svc`.

The pattern $NAMESPACE-$CRNAME-resource will also allow us to run tests for the operator in parallel.

pkg/generate/service.go
```diff
@@ -19,7 +19,7 @@ func Service(cr *v1alpha1.ImageRegistry, p *parameters.Globals) (Template, error
 			Kind:       "Service",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      p.Deployment.Name,
+			Name:      "image-registry",
```
dmage: :-1:
https://github.com/openshift/cluster-image-registry-operator/pull/26#discussion_r221916729

cmd/cluster-image-registry-operator/main.go
bparees: why are we not using rbac.authorization.k8s.io/v1 ?

bparees: (and same for the resource definitions)

/cc @gabemontero as the samples operator is also creating v1beta1 rbac resources.

legionus:
> is this supposed to be the finalizer implementation or is that in another PR? Because i don't see a finalizer here.

@bparees  I'm sorry. I miss your question. No, finalizer is already in the repository:

https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/handler.go#L356
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/handler.go#L514-L545

bparees:
>  @bparees I'm sorry. I miss your question. No, finalizer is already in the repository:

that's not a finalizer.  That's just handling a delete event.

what happens if you miss the delete event?  that's the purpose of the finalizer... it ensures the resource is not deleted until your operator has had a chance to handle it.

you can see @gabemontero's implementation here: https://github.com/openshift/cluster-samples-operator/pull/20

legionus:
> what happens if you miss the delete event? that's the purpose of the finalizer... it ensures the resource is not deleted until your operator has had a chance to handle it.

OK. I misunderstood what was meant by finalizer.

dmage:
@bparees if we miss a delete event, we'll find it out on a next CR reconciliation. It can be only when the operator is down, but we perform full reconciliation at startup.

bparees:
> @bparees if we miss a delete event, we'll find it out on a next CR reconciliation. It can be only when the operator is down, but we perform full reconciliation at startup.

yes, but we don't want to do CR reconciliation every 5s.  The finalizer approach also ensures that the resource doesn't go away until we've done all the appropriate cleanup.  Allowing the resource to go away first creates the potential for inconsistent state.

Finalizing also allows us to do cleanup that might be dependent on specific values in the resource being deleted, whereas reconciliation does not (we know the resource is missing, we don't necessarily know what state it was in before it got deleted).

legionus:
@bparees How we can miss delete event ? `watch` can miss something ? if so, then we can miss update event too and we must use reconciliation.

dmage:
@bparees ah, so you suggest to add finalizer to our CR? Yes, that might help us to cleanup global resources that wouldn't be deleted otherwise. However, we don't have any cleanup routines yet.

PR #31, opened 2018-10-09
Add finalizer

legionus:


pkg/operator/bootstrap.go
```diff
@@ -67,9 +75,12 @@ func (h *Handler) Bootstrap() (*regopapi.ImageRegistry, error) {
...
 					VolumeSource: corev1.VolumeSource{
-						PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
-							ClaimName: "image-registry-pvc",
-						},
+						EmptyDir: &corev1.EmptyDirVolumeSource{},
+						/*
+							PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+								ClaimName: "image-registry-pvc",
+							},
+						*/
```
bparees: why this change?

pkg/operator/finalizer.go
```diff
@@ -0,0 +1,167 @@
...
+				},
+			}
+
+			err := sdk.Get(cr)
+			if err != nil {
+				return fmt.Errorf("failed to get custom resource %s/%s: %s", o.ObjectMeta.Namespace, o.ObjectMeta.Name, err)
+			}
+		}
+
+		cr.ObjectMeta.Finalizers = finalizers
```
bparees: I don't think you can do this.  What if the update that caused the conflict was someone else adding a finalizer?  Now you'll have removed their finalizer.

pkg/operator/finalizer.go
```diff
@@ -0,0 +1,167 @@
...
+		kerrors.IsTimeout,
+		kerrors.IsServerTimeout,
+		kerrors.IsTooManyRequests,
+		utilnet.IsProbableEOF,
+		utilnet.IsConnectionReset,
+	}
+
+	retryTime := 3 * time.Second
+
+	err = wait.PollInfinite(retryTime, func() (stop bool, err error) {
```
bparees: why do we need to wait for the deletion to happen?

pkg/operator/bootstrap.go
legionus: Ops. my mistake.

pkg/operator/finalizer.go
legionus: Hm. How I should remove my finalizer ?

bparees: i think you need to iterate the list again and remove it again, since the list may have changed.

bparees: (that is, the same thing you're doing on line 59)

pkg/operator/finalizer.go
legionus: We do not allow two resources in the same space.

bparees: hm.  So you are waiting here until it goes away before you (re)create the default samplesresource?

bparees: err, image registry resource.

pkg/operator/finalizer.go
bparees: in practice? probably no one will.  But in theory, anyone w/ sufficient permission could and i'd rather not leave a potential hole open in case we do later introduce another finalizer for some reason and had forgotten about this.

legionus: Why then do we not follow the situation specifying our operator as an initializer (in theory this is also possible)?

pkg/operator/finalizer.go
legionus: > So you are waiting here until it goes away before you (re)create the default samplesresource?

I'm waiting for the removal of this particular resource so that you can create another resource and not finalize this one again (we can miss a delete event by definition).

pkg/operator/finalizer.go
bparees: What would that accomplish for us?

pkg/operator/finalizer.go
bparees: makes sense, thanks

legionus:
@bparees comments have been addressed.

pkg/operator/finalizer.go
dmage: @legionus they are mainly made for mutating/validating admission controllers.

pkg/operator/finalizer.go
```diff
@@ -0,0 +1,174 @@
...
+		if !kerrors.IsNotFound(err) {
+			for _, isRetryError := range errorFuncs {
+				if isRetryError(err) {
+					return
+				}
+			}
+
+			// If the error sends the Retry-After header, we respect it as an explicit confirmation we should retry.
+			if delaySeconds, shouldRetry := kerrors.SuggestsClientDelay(err); shouldRetry {
+				delayTime := time.Duration(delaySeconds)
```
dmage: time.Duration represents the time in nanoseconds. You should multiply the result by time.Second.

pkg/operator/finalizer.go
```diff
@@ -0,0 +1,174 @@
...
+	err = generate.RemoveConfigState(configState)
+	if err != nil {
+		return fmt.Errorf("unable to remove previous config state: %s", err)
+	}
+
+	return nil
+}
+
+func (h *Handler) finalizeResources(o *regopapi.ImageRegistry) error {
+	defer utilruntime.HandleCrash()
```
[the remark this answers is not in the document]
legionus: Removed

pkg/operator/finalizer.go
legionus: Ooops. Cut-and-Paste error. Sorry. Fixed.

pkg/operator/finalizer.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
+	"github.com/operator-framework/operator-sdk/pkg/sdk"
+)
+
+func (h *Handler) RemoveResources(o *regopapi.ImageRegistry) error {
+	modified := false
+
+	templetes, err := h.GenerateTemplates(o, &h.params)
+	if err != nil {
+		return fmt.Errorf("unable to genetate templates: %s", err)
```
bparees: typo: generate

legionus: Fixed

PR #33, opened 2018-10-12
Adopt existing deployment config if it exists

dmage:


pkg/operator/bootstrap.go
```diff
@@ -22,10 +27,115 @@ import (
...
+		if err := os.Setenv(key, value); err != nil {
+			return fail("unable to set environment variable %s: %v", key, err)
+		}
+	}
+	config, extraConfig, err := configuration.Parse(bytes.NewReader([]byte(`version: 0.1
+storage:
+  cache:
+    blobdescriptor: inmemory
+  filesystem:
+    rootdirectory: /registry
```
dmage: We need a configuration even if the config isn't mounted into the container from a secret. In this case we'll have to have a (hardcoded?) copy of the default configuration.

pkg/operator/bootstrap.go
```diff
@@ -22,10 +27,115 @@ import (
...
+
+func newSpecFromDeploymentConfig(dc *appsapi.DeploymentConfig) (imageregistryapi.ImageRegistrySpec, error) {
+	fail := func(format string, a ...interface{}) (imageregistryapi.ImageRegistrySpec, error) {
+		format = "unable to create an image registry spec from the DeploymentConfig %s/%s: " + format
+		a = append([]interface{}{dc.Namespace, dc.Name}, a...)
+		return imageregistryapi.ImageRegistrySpec{}, fmt.Errorf(format, a...)
+	}
+
+	containers := dc.Spec.Template.Spec.Containers
+	if len(containers) != 1 {
```
dmage: This may fail if the cluster has initialisation hooks that add sidecar containers for logging or something like this.

pkg/operator/bootstrap.go
dmage: In particular this logic will change the log level from debug to info on old installations, which might be good.

pkg/operator/bootstrap.go
```diff
@@ -22,10 +27,115 @@ import (
...
+
+	hardcodedEnvs := map[string]string{
+		"REGISTRY_HTTP_ADDR": ":5000",
+		"REGISTRY_HTTP_NET":  "tcp",
+	}
+	envs := map[string]string{}
+	for _, env := range registry.Env {
+		if env.ValueFrom != nil {
+			// FIXME(dmage): we should support it.
+			return fail("envrioment variable %s: valueFrom is not supported", env.Name)
```
bparees: typo in environment (search+replace, it's in a few places)

dmage:
Adoption for the filesystem storage driver is ready for reviewing.

But the more I work on it, the more I doubt that we will be able to test it.

bparees:
> But the more I work on it, the more I doubt that we will be able to test it.

what aspect of it do you think is not testable?  At a minimum we ought to be able to create our own "v3.11" type registry deployments w/ whatever configuration we want (even if the storage backing it doesn't actually work) and then ensure the operator migrates the config properly, or throws an error for configs it cannot migrate, right?

dmage:
@bparees my main concern that I may miss some important cases that is used in production. And that after migration we may have a slightly different configuration because we missed something (proxy, route, something else that I don't even know).

dmage:
Yep, today I've got a customer call, they put their config to /etc/registryconfig/config.yml, which I thought would never happen.

bparees:
> Yep, today I've got a customer call, they put their config to /etc/registryconfig/config.yml, which I thought would never happen.

well let's just try to do the best we can to enumerate all the sources of existing configuration inputs (env vars, config yamls, anything else?) and go from there.

pkg/operator/bootstrap.go
```diff
@@ -52,28 +64,74 @@ func (h *Handler) Bootstrap() (*regopapi.ImageRegistry, error) {
...
 		ObjectMeta: metav1.ObjectMeta{
-			Name:       "image-registry",
-			Namespace:  h.params.Deployment.Namespace,
-			Finalizers: []string{parameters.ImageRegistryOperatorResourceFinalizer},
+			Name:      resourceName(h.params.Deployment.Namespace),
+			Namespace: h.params.Deployment.Namespace,
 		},
-		Spec: regopapi.ImageRegistrySpec{
+	}
+	if err := sdk.Get(dc); err != nil {
```
bparees: this should distinguish between a not found error and other types of errors so we don't create a new registry in cases where we had an api failure checking for an existing one.

pkg/migration/migration.go
```diff
@@ -0,0 +1,438 @@
...
+		} else if env.Name == "REGISTRY_CONFIGURATION_PATH" {
+			configurationPath = env.Value
+		} else if strings.HasPrefix(env.Name, "REGISTRY_") {
+			envs[env.Name] = env.Value
+		} else {
+			return fail("unsupported environment variable %s", env.Name)
+		}
+	}
+
+	// TODO(dmage): registry.EnvFrom?
```
bparees: i'm ok not supporting that for now at least.

pkg/migration/migration.go
```diff
@@ -0,0 +1,438 @@
...
+
+	tls, tlsSecret, err := migrateTLS(config, podFileGetter)
+	if err != nil {
+		return fail("unable to process TLS configuration: %s", err)
+	}
+
+	return imageregistryapi.ImageRegistrySpec{
+		OperatorSpec: operatorapi.OperatorSpec{
+			ManagementState: operatorapi.Managed,
+			Version:         "migrated", // TODO(dmage): what do we need here?
```
bparees: i'm not sure what we want to use the version field for(or what we'd set it to normally), but it'd say for indicating this ImageRegistry value came from a migration, we should simply annotate it in some way.

pkg/operator/bootstrap.go
```diff
@@ -21,10 +24,51 @@ import (
...
+		return "docker-registry"
+	}
+	return "image-registry"
+}
+
+type namespaceResoruces struct {
+	namespace string
+}
+
+func (nr *namespaceResoruces) Secret(name string) (*corev1.Secret, error) {
```
bparees: s/Resoruces/Resources/ throughout this file

pkg/operator/bootstrap.go
```diff
@@ -21,10 +24,51 @@ import (
...
-func (h *Handler) Bootstrap() (*regopapi.ImageRegistry, error) {
-	crList := &regopapi.ImageRegistryList{
+func resourceName(namespace string) string {
+	if namespace == "default" {
+		return "docker-registry"
+	}
+	return "image-registry"
+}
+
+type namespaceResoruces struct {
```
bparees: this feels like it belongs in its own package/helper package.

pkg/migration/migration.go
```diff
@@ -0,0 +1,438 @@
...
+		}
+		if match {
+			bestVolumeMount = volumeMount
+
+			// TODO(dmage): write tests
+			key = filename[len(volumeMount.MountPath)-len(volumeMount.SubPath):]
+			if strings.HasPrefix(key, "/") {
+				key = key[1:]
+			}
+			key += volumeMount.SubPath
```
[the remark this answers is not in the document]
dmage: This is just wrong. I've added tests and updated the function.

dmage:
@bparees nm, the fix for #38 is much simpler that I thought.

bparees:
@dmage have your changes to the installer merged such that the installer is no longer creating a docker-registry in the default namespace?   We don't want to merge this until that is done, so we do not default down a path where we are trying to adopt that registry.

bparees:
/lgtm

thanks @dmage

PR #36, opened 2018-10-15
Create image-registry deployment by default

### LINKED ISSUE #34: need to retry failed deploymentconfig rollouts or switch to deployments
Upstream deployments constantly try to rollout the latest state until explicitly asked not to.  DeploymentConfigs give up unless triggered again.  Given an operator managed deploymentconfig that should be actively reconciled, you should be retrying failed deployments since they can later succeed.

The current state in openshift/installer is a good example where a blip in the first rollout will cause the image registry to be stuck indefinitely.  Switching to a deployment will resolve this problem.  Alternatively you can rebuild the oc rollout logic.

legionus:
Fixes #34

### DIFF
--- a/pkg/operator/handler.go
+++ b/pkg/operator/handler.go
@@ -42,7 +42,7 @@ func NewHandler(namespace string) (sdk.Handler, error) {
 
 	h := &Handler{
 		params:             p,
-		generateDeployment: generate.DeploymentConfig,
+		generateDeployment: generate.Deployment,
 	}
 
 	_, err := h.Bootstrap()


legionus:
God bless flakes.
/retest

legionus:
Life will be boring if CI will always work.
/retest

legionus:
```
could not wait for build: the build src failed with reason PushImageToRegistryFailed: Failed to push the image to the registry.

Registry server Address: 
Registry server User Name: serviceaccount
Registry server Email: serviceaccount@example.org
Registry server Password: <<non-empty>>
error: build error: Failed to push image: unauthorized: unable to validate token
```
/retest

bparees:
@legionus no other changes here are required to apply config changes to the deployment (instead of deploymentconfig)?

bparees:
> @legionus no other changes here are required to apply config changes to the deployment (instead of deploymentconfig)?

@legionus presumably because you're just modifying a podtemplatespec either way?

legionus:
> presumably because you're just modifying a podtemplatespec either way?

Yes. Also, all other parts are almost the same.

PR #37, opened 2018-10-15
deploy the registry operator and registry on the master

bparees:


bparees:
@deads2k ptal

@dmage @legionus fyi. Note that i have removed the NodeSelector field from the registry resource, I want to start with the minimal api we can get away with and then add fields as necessary because we're being asked to ensure our api changes are migratable from milestone 2 onward, so it'll be easier to ensure that if we're just adding new fields than if we have to remove one later.

PR #39, opened 2018-10-16
Report cluster status

legionus:
The resource is created in the same space in accordance with the [issue](https://jira.coreos.com/browse/NE-92): `The OperatorStatus should have the same name and namespace as the corresponding operator`.


pkg/operator/operatorstatus.go
```diff
@@ -0,0 +1,95 @@
...
+					Type:               osapi.OperatorProgressing,
+					Status:             osapi.ConditionUnknown,
+					LastTransitionTime: metaapi.Now(),
+				},
+				{
+					Type:               osapi.OperatorFailing,
+					Status:             osapi.ConditionUnknown,
+					LastTransitionTime: metaapi.Now(),
+				},
+			}
```
bparees: isn't this going to reset all the conditions that aren't being explicitly set?

we should only initialize conditions to unknown if 
1) they aren't set already and
2) we're not setting them now

pkg/operator/handler.go
```diff
@@ -295,8 +311,10 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 		}
 
 		if cr.Spec.Replicas == o.Status.ReadyReplicas {
+			h.operatorStatus(osapi.OperatorAvailable, osapi.ConditionTrue, "")
```
bparees: shouldn't we report available as long as there's at least 1 replica up?

pkg/operator/operatorstatus.go
bparees: oh nm, i see this is only if we didn't find the clusteroperator object.

pkg/operator/operatorstatus.go
```diff
@@ -0,0 +1,95 @@
...
+				Type:               condtype,
+				Status:             status,
+				Message:            msg,
+				LastTransitionTime: metaapi.Now(),
+			})
+
+			return sdk.Create(state)
+		}
+
+		updateOperatorCondition(state, &osapi.ClusterOperatorStatusCondition{
```
bparees: not using the return value of `modified` to avoid an unnecessary update?

deploy/02-rbac.yaml
```diff
@@ -77,6 +77,14 @@ rules:
   - routes/custom-host
   verbs:
   - "*"
+- apiGroups:
+  - operatorstatus.openshift.io
+  resources:
+  - operatorstatuses
```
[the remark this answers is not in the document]
legionus: also after this PR we will need  the `cluster-version-operator` to be installed. This operator should create a CRD for `ClusterOperator`.

pkg/operator/handler.go
legionus: I think we shouldn't, since our resource asked for more relicas.

legionus:
@bparees comments addressed.

pkg/operator/operatorstatus.go
legionus: When we have no information and we only create a resource, the lack of status is equivalent to `unknown`. It seemed to me useful to show that we will set all the statuses. Also in this way, I hardcode a position in the array (although this is not important).

deploy/02-rbac.yaml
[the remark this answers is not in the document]
bparees: No it should not.  We are not responsible for creating that CRD.  Our operator gets installed *by* the cluster version operator, so the CRD will be there.

If it is not, we should simply fail to update the status until the api is there.  (but this should really never be the case).

pkg/operator/operatorstatus.go
bparees: yeah, this is fine.  It originally looked to me like you were always resetting the statuses, but I see now you're just initializing them the first time.

deploy/02-rbac.yaml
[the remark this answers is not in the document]
legionus: @bparees I mean for testing.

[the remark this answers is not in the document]
bparees: got it, yes you'll have to manually install the CRD unless you use the openshift installer itself to setup the cluster.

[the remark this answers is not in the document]
legionus: Yep. The fact is that the `cluster-version-operator` does not have a CRD in the form of a static file.

legionus:
Good job, CI.
```
2018/10/17 15:22:04 Ran for 0s
error: could not resolve inputs: could not determine inputs for step [input:root]: could not resolve base image: Get https://api.ci.openshift.org/apis/image.openshift.io/v1/namespaces/openshift/imagestreamtags/release:golang-1.10: EOF
```
/retest

legionus:
Good job, CI (ci/prow/e2e-aws)
```
2018/10/17 15:30:24 Ran for 8m15s
error: could not run steps: could not copy stable imagestreamtag: Timeout: request did not complete within allowed duration
```

pkg/operator/handler.go
bparees: It's available.  This would be a case where setting "Progressing=true" might make sense.

Available=true  (the registry is available, you can access it).  Available does not mean "matches your desired state", only that it's accessible.

Progressing=true (we are still in the process of applying your desired configuration to the registry).

Progressing should also be set to true when there is a deployment in progress for any reason.  When the deployment is complete, Progressing would be set to false.

pkg/operator/handler.go
```diff
@@ -315,6 +315,10 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 			if errOp != nil {
 				logrus.Errorf("unable to update cluster status to %s=%s: %s", osapi.OperatorAvailable, osapi.ConditionTrue, errOp)
 			}
+			errOp = h.operatorStatusVersion(cr.Spec.Version)
+			if errOp != nil {
+				logrus.Errorf("unable to update cluster status version to %s: %s", cr.Spec.Version, errOp)
+			}
```
bparees: this seems like a strange place to be setting this value from.

Fundamentally we should just be asserting ClusterOperator's version any time we process a cr event.  (obviously if it hasn't changed, no update is needed).

pkg/operator/handler.go
bparees: @legionus tagging you in this thread just to make sure it's not missed since it's a comment on an obsolete commit.

pkg/operator/handler.go
legionus: @bparees I added this because of this paragraph (from the https://jira.coreos.com/browse/NE-92):
```
The operator must set Version to its version string (e.g., "v1.2.3"). The update payload will specify a manifest for the operator. CVO sees the version specified by the update payload and will wait until the operator reports the same version.
```
My understanding of this is that we must expose the version of the registry resource that we were asked to do. Operator version does not change. The version of the image-registry operator does not change as we do not update ourselves. Therefore, this cannot be a version of the operator itself.

I indicate the custom resource version when I succeeded. Up to this point, the previous version is working.

dmage: My understanding that it should be the version of our operator so that CVO can know that the operator's upgrade was successful. But the thing that I don't understand is how CVO will know which version it should expect to see.

bparees: not sure either. @jwforres any idea where the CVO gets our operators version from?  Or where our operator is supposed to get its version from?

I don't know if that level of status checking is wired up yet.

@legionus for now just hard code it something (`v0.0.1`) and leave a TODO for us to come back to it.  My primary concern for demos on wednesday is that we are creating the ClusterOperator object and setting some basic values in it.  As I said in my email and this morning, getting to the final implementation specifics will be an iterative process.

jwforres: I believe you want to put your version onto your ClusterOperator's status.version, on my 4.0 cluster from yesterday there are several ClusterOperator instances being created now and they all report this there

cc @abhinavdahiya

bparees: yeah this came up at lunch and basically:

1) the version value is entirely up to us.  at some point i'd like it to be based in part on the git commit, but for now, just hardcode it and let's move on.

2) the CVO (or whatever is consuming the ClusterOperator values) has no way to know what the value is supposed to be.  Whether it ever will is unclear, but again, for now, just set it to something and move on.

legionus:
@bparees @dmage comments addressed. please review again.

pkg/operator/handler.go
```diff
@@ -348,10 +391,24 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 				conditionResourceApply(o, operatorapi.ConditionFalse, fmt.Sprintf("unable to remove objects: %s", err), &statusChanged)
 			}
 		case operatorapi.Managed:
+			errOp := h.operatorStatus(osapi.OperatorProgressing, osapi.ConditionTrue, "registry server is in the process of deployment")
+			if errOp != nil {
+				logrus.Errorf("unable to update cluster status to %s=%s: %s", osapi.OperatorProgressing, osapi.ConditionFalse, errOp)
+			}
```
bparees: @legionus I think this will need a follow up.  consider the case where we're doing a relist:  

we see the registryresource
we set the condition to progressing=true
we "apply changes" but there are no changes, so no deployment is triggered
we will not get a deployment update event to trigger us to set progressing=false (until the relist for deployments occurs).

So really we should not be setting "progressing=true" if we are not actually triggering a deployment.

again i'm going to mark this lgtm so we can move forward, but please plan to follow up on this (unless i've missed something).

bparees:
@gabemontero fyi this is the registry operator status you'll need for your demo.

PR #41, opened 2018-10-17
get the image-registry image name from the cluster env

bparees:


bparees:
/assign @legionus 

in theory this will cause us to inherit the cluster-defined source for the image when the operator is installed by the CVO.

deploy/04-operator.yaml
```diff
@@ -36,3 +36,5 @@ spec:
                   fieldPath: metadata.namespace
             - name: OPERATOR_NAME
               value: "cluster-image-registry-operator"
+            - name: IMAGE
+              value: docker.io/openshift/origin-docker-registry
```
abhinavdahiya: This needs to be a exact match to the value in `name` from here https://github.com/openshift/cluster-image-registry-operator/pull/41/files#diff-2a129732795c02bf1b4c8ea7f7645ef1R12

bparees: thanks, will fix.

pkg/generate/podtemplatespec.go
```diff
@@ -225,7 +228,7 @@ func PodTemplateSpec(cr *v1alpha1.ImageRegistry, p *parameters.Globals) (corev1.
 			Containers: []corev1.Container{
 				{
 					Name:  "registry",
-					Image: cr.Spec.ImagePullSpec,
```
legionus: @bparees What is `ImagePullSpec` ? Should we ignore this field? If so, why is it needed in `OperatorSpec` ?

legionus:
Good job, CI.
```
aws_eip.nat_eip.5: Error creating EIP: AddressLimitExceeded: The maximum number of addresses has been reached.
	status code: 400, request id: 5154a95f-1ecc-4281-9d14-76865b4feeb5
```
/retest

dmage:
@bparees we don't have anything to migrate yet. But just out of curiosity how will we know that the provided image and the installed image have different major versions and additional actions are required? I don't see how we can get the version.

pkg/generate/podtemplatespec.go
legionus: @bparees The operator api has `ImagePullPolicy`. Since we don't use `ImagePullSpec`, should we use `ImagePullPolicy` from the custom resource or we will get it from another magic env variable ? 

https://github.com/openshift/api/blob/master/operator/v1alpha1/types.go#L31

bparees:
> @bparees we don't have anything to migrate yet. But just out of curiosity how will we know that the provided image and the installed image have different major versions and additional actions are required? I don't see how we can get the version.

Great question.  @jwforres any details on how that's supposed to work?  Essentially the operator can be told "install this registry image" w/o knowing anything about the registry image its supposed to install.  What happens if migration is needed prior/after updating the registry image?  How does the operator know it needs to do that?

bparees:
@legionus any concerns that can't be discussed in a followup?  This gets our operator to deploy the "correct" image-registry image.

legionus:
> any concerns that can't be discussed in a followup?

As you wish.
/lgtm

Issue #43, opened 2018-10-22
progressing status is wrong


bparees:
both the image-registry resource and the clusteroperator resource are reporting "progressing=true" when the deployment is complete.

The deployment itself reports:
```
  - lastTransitionTime: 2018-10-22T15:18:03Z
    lastUpdateTime: 2018-10-22T15:18:31Z
    message: ReplicaSet "image-registry-7cbbc9dc8c" has successfully progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
```

So either this is a misunderstanding on our part of what that means, or deployments have a bug in which they should be setting progressing to false/completed=true.

Certainly the "message: ReplicaSet "image-registry-7cbbc9dc8c" has successfully progressed." would imply this might be a deployments bug.

/assign @legionus

bparees:
correction, only the image-registry resource is reporting progressing.  The clusteroperator shows progressing=false, which makes sense since its logic is driven by the # of replicas (though i think that logic is wrong and we need to discuss it)

gabemontero:
If a fix gets in before the demo on Wednesday I'll try to include it.

legionus:
@bparees Do you have data on the number of replicas in the deployment status?

bparees:
It was 1.

Ben Parees | OpenShift

legionus:
But should be more than one or is it the number that you specified in the image-registry custom resource ?

legionus:
> But should be more than one or is it the number that you specified in the image-registry custom resource ?

It does not matter.

bparees:
> But should be more than one or is it the number that you specified in the image-registry custom resource ?

it was 1, and it should be 1 (1 was the requested number)

gabemontero:
FYI I was able to cut a new demo video using the fix from https://github.com/openshift/cluster-image-registry-operator/pull/44 and the `Progressing` condition showed what we expected.

bparees:
excellent!

PR #44, opened 2018-10-23
Update image-registry status based on deployment status

legionus:
Fixes: https://github.com/openshift/cluster-image-registry-operator/issues/43
Disclaimer: https://gist.github.com/legionus/0dc3335208a7567bbb9f01d2ed94e673

### DIFF
--- a/pkg/operator/handler.go
+++ b/pkg/operator/handler.go
@@ -11,10 +11,11 @@ import (
 	"github.com/operator-framework/operator-sdk/pkg/util/k8sutil"
 
 	kappsapi "k8s.io/api/apps/v1"
-	corev1 "k8s.io/api/core/v1"
+	coreapi "k8s.io/api/core/v1"
 	rbacapi "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 
 	appsapi "github.com/openshift/api/apps/v1"
 	operatorapi "github.com/openshift/api/operator/v1alpha1"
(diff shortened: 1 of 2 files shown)


dmage:
/cc @bparees 

I'm puzzled with our Progressing and Failing conditions.

If we find during reconciliation that our deployment is changed and we need to update it, should we switch on the progressing status before updating the deployment? What is our state when the deployment is updated, but its controller hasn't started to process it? What if its controller decide that it can't process it?

The proposed change looks good to me, but I don't know how to use such conditions.

pkg/operator/handler.go
```diff
@@ -79,6 +80,82 @@ type Handler struct {
...
+			if condition.Type != kappsapi.DeploymentProgressing {
+				continue
+			}
+			return condition.Status == coreapi.ConditionTrue
+		}
+	}
+	return false
+}
+
+func (h *Handler) syncDeploymentStatus(cr *regopapi.ImageRegistry, o runtime.Object, statusChanged *bool) {
```
gabemontero: There wasn't an issue with the Failing condition when @bparees and I looked at it yesterday, but I'm mildly curious why you left the updating of `osapi.OperatorFailing` in the main Handle method, but handle the other 2 here.

You could centralize things by adding err error as a parameter here and use this method where you are manipulating `osapi.OperatorFailing`.

pkg/operator/handler.go
```diff
@@ -123,16 +200,19 @@ func conditionResourceApply(cr *regopapi.ImageRegistry, status operatorapi.Condi
...
 }
 
-func conditionDeployment(cr *regopapi.ImageRegistry, status operatorapi.ConditionStatus, m string, modified *bool) {
-	if status == operatorapi.ConditionFalse {
+func conditionSyncDeployment(cr *regopapi.ImageRegistry, syncSuccessful operatorapi.ConditionStatus, m string, modified *bool) {
+	reason := "DeploymentProgressed"
+
+	if syncSuccessful == operatorapi.ConditionFalse {
+		reason = "DeploymentInProgress"
 		logrus.Errorf("condition failed on %s %s/%s: %s", cr.GetObjectKind().GroupVersionKind().Kind, cr.Namespace, cr.Name, m)
```
bparees: this should log what condition is being set to failed.  Also is this really an error we should always be logging?  We would expect this in cases where the deployment is still progressing.

bparees:
> Disclaimer: https://gist.github.com/legionus/0dc3335208a7567bbb9f01d2ed94e673

if your point here is that the deployment status is wrong (it says it is progressing when the deployment is actually complete), did you talk to the master team or open a bug?

pkg/operator/handler.go
legionus: After discussing on the scram last week, we agreed that `ClusterOperator.Status.Available` and `ClusterOperator.Status.Processing` should reflect deployment status. `ClusterOperator.Status.Available=True` means that the `image-registry` is available for requests.

`OperatorFailing` is operator state. It is set depending on whether we update the resources.

pkg/operator/handler.go
legionus: @bparees Removed.

legionus:
> if your point here is that the deployment status is wrong (it says it is progressing when the deployment is actually complete), did you talk to the master team or open a bug?

I want to note that the deployment status does not always reflect an error. Sometimes he can just stop and personally I don’t know how to show the state shown by me through our statuses. In my example, the update was not fully completed, and there are two versions of the pods in the cluster.

https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L413-L425

dmage:
>> Disclaimer: https://gist.github.com/legionus/0dc3335208a7567bbb9f01d2ed94e673

> if your point here is that the deployment status is wrong (it says it is progressing when the deployment is actually complete), did you talk to the master team or open a bug?

@bparees it failed, and it behaves exactly how it is described in the documentation.

bparees:
> If we find during reconciliation that our deployment is changed and we need to update it, should we switch on the progressing status before updating the deployment? 

Ideally, yes.

> What is our state when the deployment is updated, but its controller hasn't started to process it? 

progressing=true


> What if its controller decide that it can't process it?

progressing=true still.  

available - "the registry service can be accessed. It may not be running the number of instances you desire, the version you desire, or the configuration you desire"

progressing - "there is a difference between the running registry configuration and the requested/desired registry configuration.  We may or may not be actively trying to reconcile this difference.  You may need to take some action to ensure the desired configuration gets applied."

failed - "you have requested a configuration we cannot apply, because it is invalid or incompatible with respect to the existing registry configuration"

dmage:
@bparees in this case this PR is wrong, Deployments have different meaning for the Progressing condition.

> Kubernetes marks a Deployment as progressing when one of the following tasks is performed:
>
>   * The Deployment creates a new ReplicaSet.
>   * The Deployment is scaling up its newest ReplicaSet.
>   * The Deployment is scaling down its older ReplicaSet(s).
>   * New Pods become ready or available (ready for at least MinReadySeconds).
>
> -from https://kubernetes.io/docs/concepts/workloads/controllers/deployment/

It will be Progressing=false only if the last task was 10 minutes ago (by default) and it is not NewReplicaSetAvailable (i.e. it's deployed).

legionus:
@dmage Then we need to return to the analysis of the values of `replicas`/`updatedReplicas`/`readyReplicas`.

/hold

legionus:
@dmage PTAL

/hold cancel

bparees:
"Kubernetes marks a Deployment as complete when it has the following characteristics:"

so how is a deployment marked complete?

pkg/operator/handler.go
```diff
@@ -79,6 +80,69 @@ type Handler struct {
...
+	case *appsapi.DeploymentConfig:
+		return deploy.Status.UpdatedReplicas == deploy.Spec.Replicas &&
+			deploy.Status.Replicas == deploy.Spec.Replicas &&
+			deploy.Status.AvailableReplicas == deploy.Spec.Replicas &&
+			deploy.Status.ObservedGeneration >= deploy.Generation
+	case *kappsapi.Deployment:
+		return deploy.Status.UpdatedReplicas == *(deploy.Spec.Replicas) &&
+			deploy.Status.Replicas == *(deploy.Spec.Replicas) &&
+			deploy.Status.AvailableReplicas == *(deploy.Spec.Replicas) &&
+			deploy.Status.ObservedGeneration >= deploy.Generation
```
bparees: ah hah.  this looks right.

PR #47, opened 2018-10-24
update imageconfig on service creation/update

bparees:
this PR extracts the internal registry hostname from the observed service object and reflects it into the global Image configuration object for the cluster.


pkg/operator/handler.go
```diff
@@ -305,6 +305,14 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
 		if err != nil {
 			return err
 		}
+		svc := event.Object.(*coreapi.Service)
+		if cr != nil {
+			svcHostname := fmt.Sprintf("%s.%s.cluster.svc:%d", svc.Name, svc.Namespace, svc.Spec.Ports[0].Port)
```
dmage: `.cluster.svc`? Is it something new in 4.0?

deploy/00-crd.yaml
```diff
@@ -12,3 +12,17 @@ spec:
...
+  name: images.config.openshift.io
+spec:
+  group: config.openshift.io
+  scope: Cluster
+  version: v1
+  names:
+    kind: Image
+    listKind: ImageList
+    plural: images
+    singular: image
```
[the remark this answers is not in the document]
legionus: @bparees Why is it called a not `image-registry` ?

pkg/operator/handler.go
legionus: @bparees you mean `my-svc.my-namespace.svc.cluster.local` ?

https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#services

deploy/00-crd.yaml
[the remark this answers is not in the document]
bparees: I considered the same, but this is why we have api groups.  We need to accept there will be duplicate resource types across different api groups.  The name was already agreed on here:  https://github.com/openshift/api/pull/111

pkg/operator/handler.go
bparees: sorry yes i got it backwards.

bparees: pushed fix.

pkg/operator/handler.go
```diff
@@ -52,6 +52,7 @@ func NewHandler(namespace string) (sdk.Handler, error) {
 	p.Healthz.TimeoutSeconds = 5
 
 	p.Service.Name = "image-registry"
+	p.ImageConfig.Name = "cluster"
```
bparees: @deads2k @juanvallejo  relevant to your "observe the internal registry hostname" PR.  

this PR adds the logic to the image-registry operator to publish the config object, named "cluster" (per @deads2k's "canonical" comment on the api type)

gabemontero:
@bparees - is the change from this merged PR sufficient for the sample operator to determine if the registry is up before creating/updating imagestreams?

Seems like sample operator could create a watch on the new ImageConfig object now being generated, and look for the internal registry hostname to be set, no?

pkg/operator/handler.go
juanvallejo: @bparees thanks, I'll go ahead and update https://github.com/openshift/cluster-openshift-apiserver-operator/pull/41 accordingly

Issue #38, opened 2018-10-15
move registry custom resource to be cluster scoped


bparees:
Since we only want one registry per cluster anyway (can't actually tolerate more than one) and because finalizers deadlock themselves when the finalizer and the resource being finalized run in the same namespace, we need to switch the registry resource CRD to be cluster scoped instead of namespace scoped.

/cc @deads2k 

/assign @legionus

PR #48, opened 2018-10-31
Implement basic e2e test

dmage:


test/e2e/cluster_image_registry_operator_test.go
```diff
@@ -2,25 +2,83 @@ package e2e_test
...
+				},
+			},
+			Replicas: 1,
+		},
 	}
+	if err := sdk.Create(cr); err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		// TODO(dmage): can we do it better?
```
bparees: any reason not to just delete the whole namespace?

bparees:
@smarterclayton here's a prototype test structure that @dmage has come up with for testing the image-registry operator.  Probably worth your time to take a look and see if it aligns with where you see us going.  It looks like a reasonable starting point to me.

@dmage one challenge I do see is that since this leaves the "regular" (CVO installed) cluster registry operator running, it will be problematic to exercise logic that modifies cluster scoped resources (such as the logic that published the internalRegistryHostname to a cluster scoped config object).  For those tests we may need to do destructive testing against the "regular" (CVO installed) registry operator rather than installing a 2nd instance of the operator in a separate namespace.

test/e2e/cluster_image_registry_operator_test.go
dmage: The whole namespace will be deleted by the deferred Cleanup call. But first we need to remove the finalizer from our CR to avoid the deadlock that is described at #38.

bparees: I forgot that wasn't fixed yet, but let's go ahead and fix it rather than working around it in the tests, it should be an easy change.

bparees:
> @dmage one challenge I do see is that since this leaves the "regular" (CVO installed) cluster registry operator running, it will be problematic to exercise logic that modifies cluster scoped resources (such as the logic that published the internalRegistryHostname to a cluster scoped config object). For those tests we may need to do destructive testing against the "regular" (CVO installed) registry operator rather than installing a 2nd instance of the operator in a separate namespace.

this will be further compounded when https://github.com/openshift/cluster-image-registry-operator/issues/38 is fixed and both operators are watching the same registry resource.  I think we need to do the following:

1) fix https://github.com/openshift/cluster-image-registry-operator/issues/38
2) update this test logic to, as a pre-step, disable the registry operator in the CVO(take it out of management) and then delete the operator
3) exercise test logic using the operator instance installed by the test

Again this means these tests are destructive and cannot run either in parallel or in conjunction with other component tests on the cluster, but I am fine with that and consider it expected.

dmage:
aws_eip.nat_eip.4: Error creating EIP: AddressLimitExceeded: The maximum number of addresses has been reached.
/retest

pkg/testframework/imageregistry.go
```diff
@@ -0,0 +1,220 @@
...
+		TypeMeta: metaapi.TypeMeta{
+			APIVersion: imageregistryapi.SchemeGroupVersion.String(),
+			Kind:       "ImageRegistry",
+		},
+		ObjectMeta: metaapi.ObjectMeta{
+			Name: ImageRegistryName,
+		},
+	}
+
+	// TODO(dmage): the finalizer should be removed by the operator
```
bparees: is this broken today for some reason?

dmage: I don't think there's any reason. Maybe @legionus knows.

dmage:
@bparees I'd like to merge #70 first, because tests rely on the statuses.

bparees:
> @bparees I'd like to merge #70 first, because tests rely on the statuses.

sure

dmage:
@bparees a basic health check for the registry is to create a test user, a namespace and try to push/pull few blobs? Or is it enough just to get 403 from the /v2/ endpoint?

bparees:
> @bparees a basic health check for the registry is to create a test user, a namespace and try to push/pull few blobs? Or is it enough just to get 403 from the /v2/ endpoint?

Eventually we're going to be running the origin conformance suite against the cluster, which includes pushing/pulling images.  So once that suite is being run on all the cloud platforms we support, we'll have coverage that at least ensures we are setting up storage by default correctly.

So i don't think we need operator tests that push/pull content, unless you have specific config scenarios you want to try (ie you want to ensure that when we update the registry resource config in a specific way, the registry still behaves properly).

For most of the testing I would expect it is sufficient to modify the registry resource config, and then inspect the registry deployment to ensure the config settings got applied to the registry instance.  And then we assume the registry itself does the right things w/ those config values.

dmage:
Oh, c'mon!
/retest

bparees:
@dmage is this ready for a final review/lgtm ?

dmage:
@bparees yes, it is.

legionus:
Good job, CI!
/retest

PR #49, opened 2018-10-31
remove master node selector

bparees:


bparees:
This selector is causing frustration/confusion when people try to deploy the operator on clusters that do not have a master node labeled, or that have other default nodeselector's being applied to their resources.  Until we hear otherwise, I don't see a good reason to restrict ourselves to specifically the master node.

/assign @dmage

dmage:
https://github.com/openshift/cluster-image-registry-operator/blob/7c61167496ab038f3294f253dbbb54b2c06b9b18/pkg/resource/podtemplatespec.go#L220

@bparees should this node selector be removed as well?

bparees:
hm.  yes, i guess so, at least for now. will update.

bparees:
updated, thanks @dmage

PR #50, opened 2018-11-02
Use global config to bootstrap gcs driver

legionus:


bparees:
/lgtm

thanks for pushing this forward @legionus 

/cc @coreydaley

PR #51, opened 2018-11-03
fix storage configuration check

bparees:


pkg/clusterconfig/clusterconfig.go
```diff
@@ -93,9 +92,6 @@ func Get() (*Config, error) {
 
 	cm, err := client.ConfigMaps(configNamespace).Get(configName, metaapi.GetOptions{})
 	if err != nil {
-		if errors.IsNotFound(err) {
-			return cfg, nil
-		}
```
bparees: actually this was the key culprit, since you were always returning an empty configmap, the driver logic always fell through to the "unknown storage type" error.  But same result:  an error came back to the bootstrap logic and it doesn't handle it cleanly so it crashloops instead of creating the default resource.

legionus:
@bparees, you have discovered a new way to create PR. Instead of describing the problem in the description, you use a comments to your own PR.

Can you describe the problem?

The `clusterconfig.Get` returns an empty struct even if the configmap is not found because all fields are optional and caller should handle it on its own. Some parameters can be generated like a bucket name. Some of them not and in this case caller must throw an error.

In real world the global config (configmap, secret, it does not matter) should be already created before registry bootstrap.

bparees:
> @bparees, you have discovered a new way to create PR. Instead of describing the problem in the description, you use a comments to your own PR.

I felt that commenting specifically on the relevant change would be more helpful for you.

> Can you describe the problem?

the problem is your changes result in a crashlooping operator because an unexpected error is being returned to the bootstrap function.

> The clusterconfig.Get returns an empty struct even if the configmap is not found because all fields are optional and caller should handle it on its own. Some parameters can be generated like a bucket name. Some of them not and in this case caller must throw an error.

In reality it returns an empty struct and further processing examines that empty struct and throws an error that the driver type is unknown.   That error is then received by the bootstrap logic which promptly crashes because it's an unexpected error.  

Did you run your version of the operator against a cluster that does not have the configmap? (which no cluster has today)  If you did you would see the behavior.  Run the installer and you will see it.

bparees:
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/bootstrap.go#L87-L91

That i the logic which, when it sees the "unknown storage type" error (*not* a storage not configured error), returns the error up the stack and ultimately crashes the operator.

legionus:
> In reality it returns an empty struct and further processing examines that empty struct and throws an error that the driver type is unknown. That error is then received by the bootstrap logic which promptly crashes because it's an unexpected error.

This is the behavior I expected (more or less).

> In reality it returns an empty struct and further processing examines that empty struct and throws an error that the driver type is unknown. That error is then received by the bootstrap logic which promptly crashes because it's an unexpected error.

Ok, but to change this, it is more correct to return `ErrStorageNotConfigured` in line https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/storage.go#L88.

PR #52, opened 2018-11-03
Updating S3 configuration completion

coreydaley:


coreydaley:
@legionus @dmage @bparees ptal
I still need to write test(s) but please take a look at what I have done thus far.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -92,13 +102,59 @@ func Get() (*Config, error) {
...
+	if err != nil && errors.IsNotFound(err) {
+		logrus.Warnf("secret %s/%s not found, skipping aws configuration", kubeSystemNamespace, awsSecretName)
+	} else {
+		cm.Data["storage_type"] = string(regopapi.StorageTypeS3)
+		if v, ok := awsCreds.Data["aws_access_key_id"]; ok {
+			cm.Data["storage_aws_accesskey"] = string(v)
+		}
+		if v, ok := awsCreds.Data["aws_secret_access_key"]; ok {
+			cm.Data["storage_aws_secretkey"] = string(v)
+		}
```
bparees: this should all be refactored into functions that are specific for dealing with s3 storage configuration.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -92,13 +102,59 @@ func Get() (*Config, error) {
...
+		clusterConfig, err := client.ConfigMaps(kubeSystemNamespace).Get(clusterConfigName, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
```
bparees: should log something here that you couldn't find the cluster install config.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -92,13 +102,59 @@ func Get() (*Config, error) {
...
+				APIVersion: corev1.SchemeGroupVersion.String(),
+				Kind:       "ConfigMap",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      configName,
+				Namespace: configNamespace,
+			},
+			Data: make(map[string]string),
+		}
+	}
```
bparees: refactor this and the logic below that attempts to create/update it, into a single helper that is CreateOrUpdateConfigMap (also why aren't these configmap keys simply part of the S3 storage struct on the registry resource?)  It's just the region value right?

pkg/clusterconfig/clusterconfig.go
```diff
@@ -92,13 +102,59 @@ func Get() (*Config, error) {
...
+			if err := yaml.Unmarshal([]byte(clusterConfig.Data["install-config"]), &installConfig); err != nil {
+				return nil, err
+			}
+			cm.Data["storage_aws_region"] = installConfig.Platform.AWS.Region
+		}
+	}
+
+	// TODO: Insert other cloud configurations here
+	if errors.IsNotFound(cmerr) {
+		cm, err = client.ConfigMaps(configNamespace).Create(cm)
```
bparees: I don't think this configmap serves any purpose.  @legionus introduced it as a mock version of the install-config. Now that we have the install-config, we should just be reading the install config and setting config fields on the registry resource+secret.  There is no reason to copy content from the install-config into a configmap that is consumed by the registry operator.

pkg/storage/gcs/gcs.go
```diff
@@ -82,39 +78,7 @@ func (d *driver) Volumes() ([]coreapi.Volume, []coreapi.VolumeMount, error) {
...
-			cur.StringData = make(map[string]string)
-		}
-		cur.StringData["STORAGE_GCS_KEYFILE"] = gcfg.Storage.GCS.KeyfileData
-
-		if errors.IsNotFound(err) {
-			return sdk.Create(cur)
-		}
-		return sdk.Update(cur)
-	})
+	gcfg, err := util.GetConfig(d.Name, d.Namespace)
```
bparees: should be getting this directly from the install-config object

pkg/storage/s3/s3.go
```diff
@@ -106,14 +97,26 @@ func (d *driver) createBucket(svc *s3.S3) error {
 }
 
 func (d *driver) CompleteConfiguration() error {
+	s3cfg, err := util.GetConfig(d.Name, d.Namespace)
```
bparees: get this directly from the install-config object.

pkg/storage/s3/s3.go
```diff
@@ -106,14 +97,26 @@ func (d *driver) createBucket(svc *s3.S3) error {
...
 	}
 	svc := s3.New(sess)
 
+	if len(d.Config.Bucket) == 0 {
+		d.Config.Bucket = s3cfg.Storage.S3.Bucket
+	}
+	if len(d.Config.Region) == 0 {
+		d.Config.Region = s3cfg.Storage.S3.Region
+	}
+
```
bparees: comment on the bucket creation logic below:

You need to ensure that on re-sync, we are making sure the bucket exists.  That means a few things:
1) Adding a condition on the registry resource that indicates when we have successfully created the bucket (we should not be deploying the registry until we are sure we created the bucket)
2) a mechanism for ensuring that if the S3 bucket name changed, the condition is cleared until we create the new bucket.

this can be a follow up but i don't consider the s3 feature complete until we have these things.

pkg/storage/util/util.go
```diff
@@ -0,0 +1,65 @@
...
+	"k8s.io/apimachinery/pkg/api/errors"
+	metaapi "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+
+	regopapi "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1alpha1"
+	"github.com/openshift/cluster-image-registry-operator/pkg/clusterconfig"
+)
+
+func GetConfig(name string, namespace string) (*clusterconfig.Config, error) {
+	cfg, err := clusterconfig.Get()
```
bparees: this needs to be getting this from the install-config and cluster-wide credentials.

And this function is poorly named for something that has side-effects of setting up a secret.  Refactor it into a helper to setup the credential secret, and (if still appropriate) a helper to get the config.

bparees: and clusterconfig.Get() of course needs to be updated to just get the config directly from the install-config and get rid of the intermediary configmap.

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/clusterconfig/clusterconfig.go
coreydaley: 10-4, I thought that looked kinda fishy when i was doing it :(

pkg/storage/gcs/gcs.go
coreydaley: updated in next push

pkg/storage/s3/s3.go

pkg/storage/util/util.go
coreydaley: updated in next push

pkg/clusterconfig/clusterconfig.go
coreydaley: global config no longer needed, most of this is removed in the next push

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

test/e2e/aws_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+
+	regopapi "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1alpha1"
+	"github.com/openshift/cluster-image-registry-operator/pkg/clusterconfig"
+
+	clientcore "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+func TestAWS(t *testing.T) {
```
dmage: What are the preconditions for this tests? The cluster is deployed in AWS and the operator has finished its bootstrapping?

coreydaley: Correct.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -6,18 +6,24 @@ import (
 	"os/user"
 	"path/filepath"
 
+	"github.com/ghodss/yaml"
```
dmage: Use yaml from k8s.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -79,6 +85,39 @@ func GetConfig() (*restclient.Config, error) {
 	return nil, fmt.Errorf("could not locate a kubeconfig")
 }
 
+func getInstallConfig(client *clientcore.CoreV1Client) (*installer.InstallConfig, error) {
+	clusterConfig, err := client.ConfigMaps(kubeSystemNamespace).Get(clusterConfigName, metav1.GetOptions{})
+	if err != nil {
+		logrus.Errorf("%s/%s was not found or is otherwise inaccessible", kubeSystemNamespace, clusterConfigName)
```
dmage: It almost always a bad practice to log and return the error. It's up to the caller to decide if this worth to be logged.

pkg/storage/gcs/gcs.go
```diff
@@ -80,47 +77,23 @@ func (d *driver) Volumes() ([]coreapi.Volume, []coreapi.VolumeMount, error) {
...
-				return fmt.Errorf("failed to get secret %s: %s", cur.Name, err)
-			}
-		}
+	data["STORAGE_GCS_KEYFILE"] = keyfileData
 
-		if cur.StringData == nil {
-			cur.StringData = make(map[string]string)
-		}
-		cur.StringData["STORAGE_GCS_KEYFILE"] = gcfg.Storage.GCS.KeyfileData
+	return util.CreateOrUpdateSecret("image-registry", "openshift-image-registry", data)
```
dmage: Why can't we store this in private-configuration?

test/e2e/aws_test.go
```diff
@@ -0,0 +1,102 @@
...
+		t.Errorf("unable to get client: %#v", err)
+	}
+
+	cfg, err := clusterconfig.Get()
+	if err != nil {
+		t.Errorf("unable to get cluster configuration: %#v", err)
+	}
+
+	// If the storage type is not S3, skip this test.
+	if regopapi.ImageRegistryStorageType(cfg.Storage.Type) != regopapi.StorageTypeS3 {
```
dmage: It's better to change the type of the Type variable.

pkg/apis/imageregistry/v1alpha1/types.go
```diff
@@ -59,6 +59,8 @@ type ImageRegistryConfigStorageFilesystem struct {
 	VolumeSource corev1.VolumeSource `json:"volumeSource,omitempty"`
 }
 
+type ImageRegistryStorageType string
```
dmage: Why is it a part of our public API?

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/apis/imageregistry/v1alpha1/types.go
bparees: +1, @coreydaley please move this out of the api

pkg/storage/storage.go
```diff
@@ -111,42 +107,22 @@ func NewDriver(crname string, crnamespace string, cfg *opapi.ImageRegistryConfig
...
-			default:
-			  	storage="filesystem"
-		*/
+		return opapi.StorageTypeEmptyDir, nil
+	case installConfig.Platform.AWS != nil:
+		return opapi.StorageTypeS3, nil
+		// case installConfig.Platform.OpenStack != nil:
+		// 	return opapi.StorageTypeSwift, nil
+	default:
+		return opapi.StorageTypeFileSystem, nil
```
bparees: we can't start returning type FileSystem until the rest of the wiring actually does the right thing with it. comment this out.

and put the commented out block for Swift back in.

bparees: (make `default:` return emptystring for now)

pkg/storage/gcs/gcs.go
bparees: You are blindly blowing away any other keys that might be in the secret.  don't do that.

Either CreateOrUpdateSecret needs to merge the content of "data" with what's already there, or callers need to merge what is already there into "data".  The former is probably better for handling update conflicts.

pkg/util/util.go
```diff
@@ -0,0 +1,110 @@
...
+
+type S3 struct {
+	AccessKey string
+	SecretKey string
+	Bucket    string
+	Region    string
+}
+
+type Storage struct {
+	Type  regopapi.ImageRegistryStorageType
```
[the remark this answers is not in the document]
coreydaley: updated in next push

pkg/storage/storage.go
coreydaley: Updated in next push

test/e2e/main_test.go
```diff
@@ -54,7 +54,6 @@ func waitForOperator() error {
 			fmt.Printf("error waiting for operator deployment to exist: %v\n", err)
 			return false, nil
 		}
-		fmt.Println("found operator deployment")
```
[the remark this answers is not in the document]
coreydaley: yes, it was

pkg/apis/imageregistry/v1alpha1/types.go
coreydaley: updated in current push

test/e2e/aws_test.go
coreydaley: updated

pkg/storage/gcs/gcs.go
coreydaley: updated in next push

coreydaley:
@bparees ptal

bparees:
it looks good to me, but i'd like @dmage or @legionus to give it a once over as well.

@legionus i'd prefer to merge this before your refactor that does away w/ the SDK.  I know that makes for more work for your refactor, but getting AWS storage working is critical right now.

legionus:
@bparees I sad to hear that. ok, I will spend more time with rebase. as you wish.

legionus:
> yes i feel badly about it, thank you for understanding.

No. I do not understand, but it does not matter.

PR #53, opened 2018-11-05
Cluster-scoped image resource

### LINKED ISSUE #38: move registry custom resource to be cluster scoped

dmage:
Fixes #38.
/cc @legionus

### DIFF
--- a/cmd/cluster-image-registry-operator/main.go
+++ b/cmd/cluster-image-registry-operator/main.go
@@ -75,6 +75,7 @@ func main() {
 
 	watch(rbacapi.SchemeGroupVersion.String(), "ClusterRole", "", 0)
 	watch(rbacapi.SchemeGroupVersion.String(), "ClusterRoleBinding", "", 0)
+	watch(regopapi.SchemeGroupVersion.String(), "ImageRegistry", "", 10*time.Minute)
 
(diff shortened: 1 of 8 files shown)


legionus:
@dmage How about 

https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/handler.go#L177
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/handler.go#L462
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/handler.go#L468
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/finalizer.go#L83
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/podtemplatespec.go#L120
(and more)

?

dmage:
@legionus updated

legionus:
@dmage where is "github.com/openshift/cluster-image-registry-operator/pkg/metautil" ? ))

PR #54, opened 2018-11-05
Return ErrStorageNotConfigured if we cannot configure storage driver

legionus:
Fix crashloop described in #51

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


legionus:
Good job CI !

> 2018/11/05 16:39:32 Ran for 45m19s
> error: could not run steps: could not wait for pod to complete: could not wait for pod completion: the pod ci-op-sq8d9b4g/e2e-aws failed after 43m53s (failed containers: setup, test):  unknown

/retest

PR #55, opened 2018-11-06
dockerfile for running e2e tests

bparees:


bparees:
@legionus @dmage feel free to ignore this for now, just trying to get some baseline stuff working, we can revisit if/when i get something working.

PR #56, opened 2018-11-06
introduce a minimal e2e test that confirms the registry operator came up

bparees:


bparees:
@dmage i know this is going to conflict a bit w/ your (much better) long term framework but again i want to get a stake in the ground so I can start setting up CI jobs that invoke our e2e tests.

Issue #58, opened 2018-11-07
Fatal "unknown storage backend" when the backend is empty


wking:
@RobertKrawitz reports this operator crashing on both libvirt and AWS with:

```
time=""2018-11-05T19:42:08Z"" level=info msg=""Cluster Image Registry Operator Version: 601f5c3-dirty""
time=""2018-11-05T19:42:08Z"" level=info msg=""Go Version: go1.10.3""
time=""2018-11-05T19:42:08Z"" level=info msg=""Go OS/Arch: linux/amd64""
time=""2018-11-05T19:42:08Z"" level=info msg=""operator-sdk Version: 0.0.6+git""
time=""2018-11-05T19:42:08Z"" level=info msg=""Metrics service cluster-image-registry-operator created""
time=""2018-11-05T19:42:08Z"" level=info msg=""generating registry custom resource""
time=""2018-11-05T19:42:08Z"" level=fatal msg=""unknown storage backend: """
```

From the first line of the logs, you can see that's with 601f5c32fdd327cb2e2309a8aee0b5362f0b892d from this repository.  I'll file a pull request at least expanding the logs for the missing config-map case.

bparees:
this was fixed in https://github.com/openshift/cluster-image-registry-operator/pull/54

wking:
> this was fixed in #54 

Ah, sorry.  I'd searched for "unknown storage backend", saw no hits, and didn't dig deeper :p.

PR #60, opened 2018-11-07
pkg/clusterconfig: Log missing config-map

wking:
This may be why some operator runs are [dying with][1] (although I haven't read through #54 yet):

```
time=""2018-11-05T19:42:08Z"" level=info msg=""Cluster Image Registry Operator Version: 601f5c3-dirty""
time=""2018-11-05T19:42:08Z"" level=info msg=""Go Version: go1.10.3""
time=""2018-11-05T19:42:08Z"" level=info msg=""Go OS/Arch: linux/amd64""
time=""2018-11-05T19:42:08Z"" level=info msg=""operator-sdk Version: 0.0.6+git""
time=""2018-11-05T19:42:08Z"" level=info msg=""Metrics service cluster-image-registry-operator created""
time=""2018-11-05T19:42:08Z"" level=info msg=""generating registry custom resource""
time=""2018-11-05T19:42:08Z"" level=fatal msg=""unknown storage backend: """
```

When returning a potentially-troublesome, empty configuration with a `nil` error, we should at least log the fact that the expected config-map wasn't there.

[1]: https://github.com/openshift/cluster-image-registry-operator/issues/58#issue-378091526

### DIFF
--- a/pkg/clusterconfig/clusterconfig.go
+++ b/pkg/clusterconfig/clusterconfig.go
@@ -6,6 +6,7 @@ import (
 	"os/user"
 	"path/filepath"
 
+	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metaapi "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientcore "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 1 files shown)


bparees:
ultimately that configmap is stub that's going away, there's never going to be a reason to expect it actually exists, but sure we can merge this for now.

/lgtm

wking:
> ... there's never going to be a reason to expect it actually exists...

Where will the configuration come from in the absence of a similar config-map?

bparees:
> Where will the configuration come from in the absence of a similar config-map?

Similar perhaps, not this.  I expect the installer to provide sufficient information(in some form, but not this specific configmap) about the cluster for our operator to gather the information and make the right decision about what storage type to use.  If you want to make the installer produce this configmap for us, i'm certainly not going to complain, but i've heard nothing implying that is on the table, and the installer is the only thing that has the information needed.

PR #61, opened 2018-11-07
replace test-e2e makefile logic

bparees:


bparees:
@dmage @legionus fyi, i'm going to update our ci job to invoke this instead so we don't have to touch the release repo when we want to change our e2e invocation logic.

bparees:
replacing with https://github.com/openshift/cluster-image-registry-operator/pull/62

PR #64, opened 2018-11-07
Adapt DCs with deprecated environment variables

dmage:


pkg/migration/testdata/openshift-default-registry/deploymentconfig.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - name: REGISTRY_HTTP_ADDR
+          value: :5000
+        - name: REGISTRY_HTTP_NET
+          value: tcp
+        - name: REGISTRY_HTTP_SECRET
+          value: this-is-an-http-/secret/-for-the-registry
+        - name: REGISTRY_MIDDLEWARE_REPOSITORY_OPENSHIFT_ENFORCEQUOTA
+          value: "false"
+        - name: OPENSHIFT_DEFAULT_REGISTRY
+          value: openshift-registry.default.svc:5000
```
bparees: how is the test confirming this shows up in the migrated registry config?

dmage: this value isn't customizable anymore, so this value should be just ignored.

bparees: When we get to running migration, we need to think about the implications if someone has it set to a different value..  that is probably a migration blocker.  but ok for now.

PR #67, opened 2018-11-11
Add a RHEL7 dockerfile and standardize format

smarterclayton:


Dockerfile.rhel7
```diff
@@ -0,0 +1,11 @@
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.10 AS builder
+WORKDIR /go/src/github.com/openshift/cluster-image-registry-operator
+COPY . .
+RUN make build
+
+FROM registry.svc.ci.openshift.org/ocp/4.0:base
```
bparees: any particular reason you didn't go with `openshift/ocp-4.0:base` to stay aligned w/ the origin image naming pattern?

Dockerfile
```diff
@@ -1,12 +1,11 @@
...
 WORKDIR /go/src/github.com/openshift/cluster-image-registry-operator
 COPY . .
 RUN make build
 
-FROM centos:7
+FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+COPY --from=builder /go/src/github.com/openshift/cluster-image-registry-operator/tmp/_output/bin/cluster-image-registry-operator /usr/bin/
 RUN useradd cluster-image-registry-operator
 USER cluster-image-registry-operator
-COPY --from=builder /go/src/github.com/openshift/cluster-image-registry-operator/tmp/_output/bin/cluster-image-registry-operator /usr/bin
```
bparees: why did you reorder this?  the previous ordering was optimized for layer cache reuse.

(actually probably the best order would be for the binary copy to be the last statement)

```
LABEL blah
RUN useradd
USER foo
COPY manifest content (slowly changing)
COPY binary (fast changing)
```

is probably the ideal ordering.

bparees:
/lgtm

will deal w/ reordering these dockerfile lines in a followup.

Dockerfile.rhel7
smarterclayton: Because the repo has to be private.  Origin instead will be moved.

Dockerfile
smarterclayton: Three is no cache reuse in any of our build systems anymore.  Also, we are making this consistent across everyone, and people having special logic with ordering is confusing.

For 99% of binaries, we should be able to generate and validate these docker files from a central spot.  That's our end goal, because the variance here doesn't have much benefit.

PR #69, opened 2018-11-13
Refactoring

### LINKED ISSUE #66: Uninstall procedure
We need a mechanism that will allow us to remove the registry. We have the ManagementState removed, but there is no indication of the removal progress. I can't delete the custom resource because the operator will re-bootstrap it.

I need this because I want to uninstall the operator and removed everything that was generated by the operator.

legionus:
- [x] Remove image-registry-config-state.
- [x] Fix detection of changes in the ConfigMaps/Secrets.
- [x] Add workqueue with rate limit.
- [x] Ignore events about changing the status of the ImageRegistry resource.

Fixes #66


pkg/operator/handler.go
```diff
@@ -125,12 +126,19 @@ func (h *Handler) syncDeploymentStatus(cr *regopapi.ImageRegistry, o runtime.Obj
 		logrus.Errorf("unable to update cluster status to %s=%s: %s", osapi.OperatorAvailable, osapi.ConditionTrue, errOp)
 	}
 
+	conditionRemoved(cr, resourceRemoved, operatorAvailableMsg, statusChanged)
```
dmage: When we know that all resources have been removed, can we drop the finalizer?

legionus: @dmage This code is wrong. Changed.

PR #70, opened 2018-11-19
Update status logic

dmage:


pkg/operator/status.go
```diff
@@ -124,45 +116,77 @@ func isDeploymentStatusComplete(o runtime.Object) bool {
...
+		operatorProgressingMsg = "all resources are successfully applied, but the deployment does not exist"
+	} else if metaObject.GetDeletionTimestamp() != nil {
+		operatorProgressingMsg = "the deployment is being deleted"
+	} else if !isDeploymentStatusComplete(o) {
+		operatorProgressingMsg = "the deployment has not completed"
 	} else {
 		operatorProgressing = osapi.ConditionFalse
-		operatorProgressingMsg = "deployment removed"
+		operatorProgressingMsg = "everything is ready"
 	}
```
bparees: Progressing needs to be false if the current active deployment does not match the operator configuration.

So for example if the AWS credential secret has been updated, and the current deployment has not mounted that secret, progressing should be true.  (We would detect this by determining that the credential hash annotated on the deployment does not match the hash of the current credentials in the credential secret).

Similarly if the registry resource itself has been updated, but the current deployment doesn't reflect those changes...so again, hash of the registry resource, annotated onto the registry deployment, and compare them.

pkg/operator/controller.go
```diff
@@ -260,20 +256,16 @@ func (c *Controller) sync() error {
...
 
-	if cr == nil {
-		logrus.Debugf("ImageRegistry Name=%s not found. ignore.", resourceName(c.params.Deployment.Namespace))
-		return nil
-	}
-
-	if cr.ObjectMeta.DeletionTimestamp != nil {
-		cr, err = c.Bootstrap()
+	if cr == nil || cr.ObjectMeta.DeletionTimestamp != nil {
+		_, err = c.Bootstrap()
```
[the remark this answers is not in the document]
dmage: I don't know why, but the finalizer is executed by Bootstrap.

pkg/operator/status.go
dmage: Did you want to say "Progressing needs to be *true* if the current active deployment does not match the operator configuration"? I think progressing should be true if we haven't reached the desired state. And that's what was implemented here.

pkg/operator/controller.go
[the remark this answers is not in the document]
bparees: yeah that sounds like something we should refactor :)

pkg/operator/status.go
bparees: Sorry, yes that is what i meant to say.

I think all the existing logic here is fine, but i think it's missing the case i described where the current deployment is complete(all replicas active/rolled out), but the current deployment doesn't reflect the desired configuration.  Unless "isDeploymentStatusComplete" is actually checking that?

dmage: This function is called after all resources are updated according to the custom resource, so if applyError is nil, the deployment should reflect the desired configuration. There might be a race if someone touched the deployment just after we had updated it. But in this case we would receive an event for the deployment and run this function again.

pkg/operator/controller.go
```diff
@@ -116,24 +125,22 @@ func (c *Controller) syncRoutes(cr *regopapi.ImageRegistry, modified *bool) erro
...
 
+	var applyError error
 	for name, gen := range routes {
 		tmpl, err := gen(cr, &c.params)
 		if err != nil {
-			conditionResourceApply(cr, operatorapi.ConditionFalse,
-				fmt.Sprintf("unable to make template for route %s: %s", name, err),
-				modified,
-			)
+			applyError = fmt.Errorf("unable to make template for route %s: %s", name, err)
```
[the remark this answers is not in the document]
dmage: updated

pkg/operator/status.go
bparees: ok

pkg/operator/controller.go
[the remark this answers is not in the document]
dmage: updated

bparees:
needs rebase

PR #72, opened 2018-11-20
add temporary hack to copy service ca to cluster nodes

bparees:
This is a temporary hack to ensure the service-ca (which signs the registry service) gets copied to the host nodes.

In the future this needs to be replaced/enhanced in such a way that 

1) it needs less privileges(probably)
2) ~~should use the base image from the release, not centos~~
3) the image-registry operator reports on the state of the daemonset as part of the registry operator status
4) we also copy down additional CAs that are configured in the cluster that represent other registries the nodes may need to pull from.  (ie image.config.openshift.io's additionalCA configmap)

### DIFF
--- a/deploy/08-ca-daemonset.yaml
+++ b/deploy/08-ca-daemonset.yaml
@@ -1,0 +1,55 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: registry-ca-hostmapper
+  namespace: openshift-image-registry
+spec:
+  selector:
+    matchLabels:
+      name: registry-ca-hostmapper
+  template:
+    metadata:
+      labels:
+        name: registry-ca-hostmapper
+    spec:      
+      tolerations:
+      - effect: NoSchedule
+        key: node-role.kubernetes.io/master
+        operator: Exists
+      serviceAccountName: registry-ca-hostmapper
+      containers:
+      - name: registry-ca-hostmapper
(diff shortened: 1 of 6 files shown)


bparees:
@derekwaynecarr @mrunalp ptal

bparees:
@dmage @legionus fyi

deploy/08-ca-daemonset.yaml
```diff
@@ -0,0 +1,44 @@
...
+      containers:
+      - name: registry-ca-hostmapper
+        securityContext:
+          privileged: true
+        image: docker.io/centos:7
+        command: 
+        - "/bin/sh"
+        args:
+        - "-c"
+        - "mkdir /etc/docker/certs.d/image-registry.openshift-image-registry.svc.cluster.local:5000; cp /tmp/serviceca/service-ca.crt /etc/docker/certs.d/image-registry.openshift-image-registry.svc.cluster.local:5000; sleep 60"
```
mrunalp: what is the sleep for?

bparees: so the pod dies every minute and restarts, which will cause it to pick up any change to the CA in the configmap and re-copy it.

bparees:
@ironcladlou fyi this (temporarily) fixes the CA trust issue for the registry.

@smarterclayton not sure who's around who can review this, maybe you can?

PR #74, opened 2018-11-20
autoconfigure emptydir storage on libvirt

bparees:


pkg/operator/bootstrap.go
```diff
@@ -172,7 +172,5 @@ func (c *Controller) Bootstrap() (*imageregistryapi.ImageRegistry, error) {
 		}
 	}
 
-	addImageRegistryChecksum(cr)
```
bparees: @legionus this was causing us to consider the initial bootstrap CR to be already processed and ignore it.

bparees:
@coreydaley @dmage @legionus ptal.

@coreydaley this will likely conflict a little w/ your S3 PR, but ideally the two PRs should look very similar in how/when they auto-configure storage (except of course that configuring emptydir involves a lot fewer steps).

pkg/storage/emptydir/emptydir.go
```diff
@@ -0,0 +1,81 @@
+package emptydir
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+
+	opapi "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1alpha1"
```
coreydaley: We should have a follow up PR at some point to unify the nickname for this package across the repository, I think we are currently using at least three different nicknames.  Maybe after @legionus refactoring away from the sdk lands.

pkg/operator/bootstrap.go
dmage: My PR removes this as well. One more rebase is coming...

pkg/storage/emptydir/emptydir.go
bparees: i copied this one from filesystem.go, so it's not introducing a new nickname anyway.

dmage:
/lgtm cancel
The bootstrap needs to be fixed.

pkg/storage/storage.go
```diff
@@ -37,10 +40,14 @@ func newDriver(crname string, crnamespace string, cfg *opapi.ImageRegistryConfig
 		drivers = append(drivers, azure.NewDriver(crname, crnamespace, cfg.Azure))
 	}
 
-	if cfg.Filesystem != nil {
+	if cfg.Filesystem != nil && cfg.Filesystem.VolumeSource.EmptyDir == nil {
```
dmage: What's the difference between the `filesystem` and `emptydir` drivers?

dmage:
Perhaps we need to disable CompleteConfiguration for the S3 storage to get this merged.

pkg/storage/storage.go
bparees: filesystem driver will (ultimately) create a PVC.  Perhaps it should be renamed to the persistentvolume driver.

emptydir sets up an emptydir volumesource.

pkg/storage/storage.go
```diff
@@ -71,26 +78,75 @@ func newDriver(crname string, crnamespace string, cfg *opapi.ImageRegistryConfig
...
+	// if we can't determine what platform we're on, fallback to creating
+	// a PVC for the registry.
+	storage := ""
+	switch {
+	case installConfig.Platform.Libvirt != nil:
+		storage = "emptydir"
+		// TODO - Corey enable s3, someone enable swift in the future.
+		/*
+			case installConfig.Platform.AWS != nil:
+				storage = "s3"
```
bparees: @coreydaley fyi ^^

bparees:
thanks @dmage, i changed "getPlatformStorage" to only return a valid storage type for libvirt for now.  As we add support for other platforms+storage we can re-enable returning the correct storagetypes for other platforms.

PR #80, opened 2018-11-26
tmp directory should not contain non-temporary information

coreydaley:


legionus:
These files were created by `operator-sdk`.
/lgtm

PR #84, opened 2018-11-26
Set user tags on s3 bucket

coreydaley:
- Set installConfig.Platform.AWS.UserTags as tags on the created S3 bucket

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -13,6 +13,8 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 
+	installer "github.com/openshift/installer/pkg/types"
+
 	opapi "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1alpha1"
 	"github.com/openshift/cluster-image-registry-operator/pkg/clusterconfig"
 	"github.com/openshift/cluster-image-registry-operator/pkg/storage/util"
@@ -82,7 +84,7 @@ func (d *driver) checkBucketExists(svc *s3.S3) error {
 }
 
 // createBucket attempts to create an s3 bucket with the given name
-func (d *driver) createAndTagBucket(svc *s3.S3, clusterID string) error {
+func (d *driver) createAndTagBucket(svc *s3.S3, installConfig *installer.InstallConfig) error {
 	createBucketInput := &s3.CreateBucketInput{
 		Bucket: aws.String(d.Config.Bucket),
 	}
@@ -96,21 +98,23 @@ func (d *driver) createAndTagBucket(svc *s3.S3, clusterID string) error {
 		return err
 	}
 
-	tagBucketInput := &s3.PutBucketTaggingInput{
-		Bucket: aws.String(d.Config.Bucket),
-		Tagging: &s3.Tagging{
-			TagSet: []*s3.Tag{
-				{
-					Key:   aws.String("tectonicClusterID"),
-					Value: aws.String(clusterID),
-				},
-			},
-		},
-	}
-	if _, err := svc.PutBucketTagging(tagBucketInput); err != nil {
-		return err
-	}
+	if installConfig.Platform.AWS != nil {
+		var tagSet []*s3.Tag
+		tagSet = append(tagSet, &s3.Tag{Key: aws.String("tectonicClusterID"), Value: aws.String(installConfig.ClusterID)})
+		for k, v := range installConfig.Platform.AWS.UserTags {
+			tagSet = append(tagSet, &s3.Tag{Key: aws.String(k), Value: aws.String(v)})
+		}
 
+		tagBucketInput := &s3.PutBucketTaggingInput{
+			Bucket: aws.String(d.Config.Bucket),
+			Tagging: &s3.Tagging{
+				TagSet: tagSet,
+			},
+		}
+		if _, err := svc.PutBucketTagging(tagBucketInput); err != nil {
+			return err
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/apis/imageregistry/v1alpha1/types.go
```diff
@@ -42,14 +42,17 @@ type ImageRegistryConfigStorageS3 struct {
 	Region         string `json:"region,omitempty"`
 	RegionEndpoint string `json:"regionEndpoint,omitempty"`
 	Encrypt        bool   `json:"encrypt,omitempty"`
+	ManagedStorage bool   `json:"managedStorage,omitempty"`
```
bparees: this needs to be part of status, not part of spec.  spec is things the user controls.

bparees: I was expecting a single field, like:  status.ManagedStorage   or possibly status.Storage.ManagedStorage

we don't need a unique field for each storagetype struct.

pkg/storage/azure/azure.go
```diff
@@ -76,5 +77,14 @@ func (d *driver) ValidateConfiguration(cr *opapi.ImageRegistry, modified *bool)
 		*modified = true
 	}
 
+	if v, ok := util.GetStateValue(&cr.Status, "managedStorage"); ok {
+		if v != strconv.FormatBool(d.Config.ManagedStorage) {
+			return fmt.Errorf("S3 storage management change is not supported: expected %s, but got %s", v, strconv.FormatBool(d.Config.ManagedStorage))
```
bparees: s/S3/Azure/

pkg/storage/gcs/gcs.go
```diff
@@ -147,5 +148,14 @@ func (d *driver) ValidateConfiguration(cr *opapi.ImageRegistry, modified *bool)
 		*modified = true
 	}
 
+	if v, ok := util.GetStateValue(&cr.Status, "managedStorage"); ok {
+		if v != strconv.FormatBool(d.Config.ManagedStorage) {
+			return fmt.Errorf("S3 storage management change is not supported: expected %s, but got %s", v, strconv.FormatBool(d.Config.ManagedStorage))
```
bparees: s/S3/GCS/

pkg/storage/s3/s3.go
```diff
@@ -96,15 +99,18 @@ func (d *driver) createAndTagBucket(svc *s3.S3, clusterID string) error {
 		return err
 	}
 
+	d.Config.ManagedStorage = true
+
+	var tagSet []*s3.Tag
+	tagSet = append(tagSet, &s3.Tag{Key: aws.String("tectonicClusterID"), Value: aws.String(installConfig.ClusterID)})
+	for k, v := range installConfig.Platform.AWS.UserTags {
+		tagSet = append(tagSet, &s3.Tag{Key: aws.String(k), Value: aws.String(v)})
+	}
```
bparees: add a test that confirms this is being set?  or do we not have usertags for the clusters being stood up by CI?

bparees:
@coreydaley please split this into two PRs, one for the usertags and the other for the managedstorage status field.

pkg/storage/s3/s3.go
coreydaley: Currently the UserTags are empty, so there is nothing to check.

bparees: ok

wking: >  or do we not have usertags for the clusters being stood up by CI?

CI sets `expirationDate`, openshift/release#1761.

pkg/apis/imageregistry/v1alpha1/types.go
coreydaley: I started doing that initially, but the Storage.ManagedStorage field would not be easily accessible during the CompleteConfiguration phase of configuration the storage as d.Config only holds the specific information for that storage driver.  This seemed like a cleaner way to do it.

coreydaley:
Will split it up in the next push

pkg/storage/s3/s3.go
```diff
@@ -96,15 +99,18 @@ func (d *driver) createAndTagBucket(svc *s3.S3, clusterID string) error {
 		return err
 	}
 
+	d.Config.ManagedStorage = true
+
+	var tagSet []*s3.Tag
+	tagSet = append(tagSet, &s3.Tag{Key: aws.String("tectonicClusterID"), Value: aws.String(installConfig.ClusterID)})
+	for k, v := range installConfig.Platform.AWS.UserTags {
```
wking: We're unlikely to hit tgis code when `installConfig.Platform.AWS` is `nil`.  But it's probably tood to add a guard for it anyway, to avoid a panic if it does happen.

pkg/apis/imageregistry/v1alpha1/types.go
bparees: it has to be part of status regardless.  I can buy an argument for putting the field on each storage-type specific struct, but it has to be a status struct.

pkg/storage/azure/azure.go
coreydaley: updated in next push

pkg/storage/gcs/gcs.go
coreydaley: updated in next push

pkg/storage/s3/s3.go
coreydaley: I added testing that checks the UserTags against the tags we set on the s3 bucket

pkg/apis/imageregistry/v1alpha1/types.go
coreydaley: updated per your suggestion

pkg/storage/s3/s3.go
bparees: +1

bparees:
>> @coreydaley please split this into two PRs, one for the usertags and the other for the managedstorage status field.
> Will split it up in the next push

separate commits!=separate PRs.  I was trying to help you expedite the merge since i expect the usertags PR to be less controversial than whatever discussion needs to happen around how we want to represent the ManagedStorage status.

anyway add the nil checking that @wking suggested and i think this is lgtm, but i guess we'll see if @legionus or @dmage have any concerns on the ManagedStorage status field.

pkg/storage/s3/s3.go

coreydaley:
@bparees I split this into two PRs now.  ptal

bparees:
/lgtm

thanks @coreydaley

PR #86, opened 2018-11-27
Set Status.ManagedStorage on the registry operator custom resource

coreydaley:
if we create the s3 bucket

### DIFF
--- a/pkg/apis/imageregistry/v1alpha1/types.go
+++ b/pkg/apis/imageregistry/v1alpha1/types.go
@@ -109,5 +109,10 @@ type ImageRegistryStatus struct {
 
 	InternalRegistryHostname string `json:"internalRegistryHostname"`
 
+	// ManagedStorage is a boolean which denotes whether or not
+	// we created the registry storage medium (such as an
+	// S3 Bucket)
+	ManagedStorage bool `json:"managedStorage"`
+
 	State []ImageRegistryStatusState `json:"state"`
 }
--- a/pkg/clusterconfig/clusterconfig.go
+++ b/pkg/clusterconfig/clusterconfig.go
@@ -58,13 +58,22 @@ type Config struct {
 	Storage Storage
 }
 
-func GetInstallConfig() (*installer.InstallConfig, error) {
+func GetCoreClient() (*coreset.CoreV1Client, error) {
 	kubeconfig, err := client.GetConfig()
 	if err != nil {
 		return nil, err
 	}
 
 	client, err := coreset.NewForConfig(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+
+	return client, nil
+}
+
+func GetInstallConfig() (*installer.InstallConfig, error) {
+	client, err := GetCoreClient()
 	if err != nil {
 		return nil, err
 	}
@@ -83,16 +92,10 @@ func GetInstallConfig() (*installer.InstallConfig, error) {
 }
 
 func GetAWSConfig() (*Config, error) {
-	kubeconfig, err := client.GetConfig()
+	client, err := GetCoreClient()
 	if err != nil {
 		return nil, err
 	}
-
-	client, err := coreset.NewForConfig(kubeconfig)
-	if err != nil {
-		return nil, err
-	}
-
 	cfg := &Config{}
 
 	installConfig, err := GetInstallConfig()
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -140,7 +140,7 @@ func (c *Controller) Bootstrap() error {
 			return err
 		}
 	} else {
-		err = driver.CompleteConfiguration()
+		err = driver.CompleteConfiguration(&cr.Status)
 		if err != nil {
 			return err
 		}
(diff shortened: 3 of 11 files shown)


pkg/operator/bootstrap.go
```diff
@@ -140,7 +140,7 @@ func (c *Controller) Bootstrap() error {
 			return err
 		}
 	} else {
-		err = driver.CompleteConfiguration()
+		err = driver.CompleteConfiguration(&cr.Status)
```
bparees: how do we get here if initially a registry resource already exists, with no storage configuration, and then someone updates the registry resource to include storage configuration?  (as used to be the case before you added the s3 auto-configuration, and is still the case for other storage types today)

does that even work anymore?  (supplying the storage configuration after the fact)

coreydaley: afaik we would not reach the driver.CompleteConfiguration() function, it is only called from this one spot in bootstrap.go.

legionus: My idea was that the `CompleteConfiguration` function should be called only to complete the configuration from global config when creating a new custom resource. Since the storage type cannot be changed, it is not necessary to coplete the configuration in other cases. A configuration change will cause an error.

> "no storage configured" to "use this storage configuration" should send us through the driver configuration codepath.

@bparees `no storage configured`  can _only_ happen if someone creates custom resource (without storage section) before operator first start. If the operator has already been started, then it will add the finalizer and you will not be able to delete the resource or change the storage type.

bparees: Well those assumptions are no longer valid since "CompleteConfiguration" is now being used to actually create the buckets, not just fill out the storage configuration spec.

Either the bucket creation needs to happen elsewhere, or CompleteConfiguration needs to be invoked when the user fills in the storage type later.

dmage: @bparees in this case we need to refactor our clusterconfig. It will [panic if S3 will be used on a non-AWS platform](https://github.com/openshift/cluster-image-registry-operator/blob/add05f58a0dc895f76afae44147008b287e420f1/pkg/clusterconfig/clusterconfig.go#L103).

dmage:
Can we remove https://github.com/openshift/cluster-image-registry-operator/blob/add05f58a0dc895f76afae44147008b287e420f1/pkg/clusterconfig/clusterconfig.go#L46?

pkg/operator/bootstrap.go
bparees: yes we will have to revisit that at some point.

pkg/apis/imageregistry/v1alpha1/types.go
```diff
@@ -109,5 +109,7 @@ type ImageRegistryStatus struct {
 
 	InternalRegistryHostname string `json:"internalRegistryHostname"`
 
+	ManagedStorage bool `json:"managedStorage"`
```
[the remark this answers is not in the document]
coreydaley: updated in current push

pkg/clusterconfig/clusterconfig.go
```diff
@@ -58,7 +58,7 @@ type Config struct {
 	Storage Storage
 }
 
-func GetInstallConfig() (*installer.InstallConfig, error) {
+func GetClient() (*coreset.CoreV1Client, error) {
```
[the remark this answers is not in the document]
coreydaley: updated in current push

coreydaley:
@legionus @dmage @bparees comments/suggestions have been addressed and tests are passing.

legionus:
> we need to allow admins to provide their own bucket name. What would be the motivation to remove it?

Because it is useless. In my original code I filled this field from configmap:

https://github.com/openshift/cluster-image-registry-operator/blob/13f0a615cee21fca90ef1221e48d52659c71986c/pkg/clusterconfig/clusterconfig.go#L75

But now after the creation of `GetAWSConfig` it is not used:

https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/clusterconfig/clusterconfig.go#L85-L119

coreydaley:
@bparees @legionus Correct, S3.Bucket is not used right now, but once the installer includes an option to specific a bucket to use for image registry storage, we can populate that field in GetAWSConfig().

pkg/operator/bootstrap.go
bparees: @legionus or we revisit where the create bucket is being done, I don't care what approach is taken.  I am trying to raise your awareness to scenarios that you need to consider and handle properly.  I do not care how you address them (though of course i may have feedback when i see the proposed change).

bparees:
> @bparees @legionus Correct, S3.Bucket is not used right now, but once the installer includes an option to specific a bucket to use for image registry storage, we can populate that field in GetAWSConfig().

what if i create a CR myself with the s3 storage configuration populated, before bootstrapping happens?  Are you telling me S3.Bucket is not used even then?  That is broken if true.

dmage:
@bparees no, no, this is about the field in the cluster config structure. We don’t have a source for it. The field Bucket in CR should work as expected.

pkg/operator/bootstrap.go
legionus: > I do not care how you address them (though of course i may have feedback when i see the proposed change).

@bparees as you wish.

legionus:
> what if i create a CR myself with the s3 storage configuration populated, before bootstrapping happens? Are you telling me S3.Bucket is not used even then? That is broken if true.

@bparees You did not open the links that I posted.

PR #87, opened 2018-11-27
Rollback from k8s.io/klog to github.com/golang/glog

legionus:
After discussion with Master team, we came to the conclusion that for
now you need to roll back to glog and wait for kubernetes-1.13 where it
is likely to switch to klog.

But we need to be careful, since messages from modules that use klog
will be lost now and we have no protection against this.

Fixes github.com/openshift/cluster-image-registry-operator/issues/85

### DIFF
--- a/cmd/cluster-image-registry-operator/main.go
+++ b/cmd/cluster-image-registry-operator/main.go
@@ -5,9 +5,10 @@ import (
 	"os"
 	"runtime"
 
+	"github.com/golang/glog"
+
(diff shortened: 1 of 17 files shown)


bparees:
@legionus please add an e2e test that confirms logging is working so we don't regress this in the future.

legionus:
@bparees I want to it after #48 . ok?

coreydaley:
/retest
out of vpc error

PR #89, opened 2018-11-27
fix internalregistryhostname handling

bparees:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1653563

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -35,9 +35,9 @@ func resourceName(namespace string) string {
 }
 
 func addImageRegistryChecksum(cr *regopapi.ImageRegistry) {
-	dgst, err := resource.Checksum(cr.Spec)
+	dgst, err := resource.Checksum(resource.ImageRegistryChecksumInput{cr.Spec, cr.Status})
 	if err != nil {
-		glog.Errorf("unable to generate checksum from ImageRegistry spec: %s", err)
+		glog.Errorf("unable to generate checksum from ImageRegistry: %s", err)
 		return
 	}
 
--- a/pkg/resource/generator.go
+++ b/pkg/resource/generator.go
@@ -19,6 +19,11 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
 )
 
+type ImageRegistryChecksumInput struct {
+	Spec   regopapi.ImageRegistrySpec
+	Status regopapi.ImageRegistryStatus
+}
+
 func Checksum(o interface{}) (string, error) {
 	data, err := json.Marshal(o)
 	if err != nil {
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -164,15 +164,15 @@ func (c *Controller) Handle(action string, o interface{}) {
 	glog.V(1).Infof("Processing %s object %s", action, objectInfo)
 
 	if cr, ok := o.(*regopapi.ImageRegistry); ok {
-		dgst, err := resource.Checksum(cr.Spec)
+		dgst, err := resource.Checksum(resource.ImageRegistryChecksumInput{cr.Spec, cr.Status})
 		if err != nil {
-			glog.Errorf("unable to generate checksum for ImageRegistry spec: %s", err)
+			glog.Errorf("unable to generate checksum for ImageRegistry: %s", err)
 			dgst = ""
 		}
 
 		curdgst, ok := object.GetAnnotations()[parameters.ChecksumOperatorAnnotation]
 		if ok && dgst == curdgst {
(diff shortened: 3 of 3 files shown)


bparees:
/assign @legionus 

There were two issues here:

1) the not nil check was wrong.
2) we can't use just a checksum of the spec to determine if we need to process the registry resource, our handler needs to look at changes to status fields also because we drive changes by updating status and then expecting the handler to see those changes and act on them.

bparees:
I will add a test to prevent future regressions to this after @dmage's e2e framework lands.

pkg/operator/bootstrap.go
```diff
@@ -35,7 +35,7 @@ func resourceName(namespace string) string {
 }
 
 func addImageRegistryChecksum(cr *regopapi.ImageRegistry) {
-	dgst, err := resource.Checksum(cr.Spec)
+	dgst, err := resource.Checksum(cr)
 	if err != nil {
 		glog.Errorf("unable to generate checksum from ImageRegistry spec: %s", err)
```
coreydaley: error message needs updated, no longer just creating a checksum from the spec

dmage:
> because we drive changes by updating status and then expecting the handler to see those changes and act on them.

Can you give an example when the handler gets information from status?

bparees:
Propagation of the internal Registry host name is an example. We update the
status then when we see the status update we propagate the value to the
cluster config.

The reason for that is to avoid trying to do too many things at once and
handle what to do if you fail halfway through doing them.

Ben Parees | OpenShift

legionus:
> the not nil check was wrong.

agree

> we can't use just a checksum of the spec to determine if we need to process the registry resource, our handler needs to look at changes to status fields also because we drive changes by updating status and then expecting the handler to see those changes and act on them.

Please give an example. The status is updated only by our operator. There is no need to respond to your own updates.

legionus:
> Propagation of the internal Registry host name is an example. We update the
status then when we see the status update we propagate the value to the
cluster config.

Why we cannot update the cluster config in same time as the `InternalRegistryHostname` ?

bparees:
> Why we cannot update the cluster config in same time as the InternalRegistryHostname ?

I gave you the reason above:  "The reason for that is to avoid trying to do too many things at once and
handle what to do if you fail halfway through doing them."

This is the pattern being used by other operators as well.  

Can you give me a reason why including the status in the hash is a problem?

legionus:
> I gave you the reason above: "The reason for that is to avoid trying to do too many things at once and handle what to do if you fail halfway through doing them."
> Can you give me a reason why including the status in the hash is a problem?

@bparees Now we have just one handler for any related events. Creating an event from our own actions will lead us to an extra recheck of all the resources that we control. We will need to get all of them. This was the case before as we checked all the resources that we control if the custom resource was changed.
Checking checksums and owners is designed to reduce the number of events.

So, such an event from our actions will be expensive.

bparees:
> @bparees Now we have just one handler for any related events. Creating an event from our own actions will lead us to an extra recheck of all the resources that we control. We will need to get all of them.

we should be caching the resources we care about in a listwatcher, not getting them from the apiserver every time, so it should not be expensive.  If we are not, that should be fixed.

However I think there is a problem here that the checksum is now going to include the checksum annotation, which means it will keep changing/triggering a new change.  I will fix that.

bparees:
checksum logic updated to incude Spec+Status but not the annotations or other metadata fields.

pkg/resource/generator.go
```diff
@@ -19,6 +19,11 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
 )
 
+type ChecksumInput struct {
+	Spec   regopapi.ImageRegistrySpec
```
legionus: If this is a generic `struct` then the field type should be a `interface{}`. If this `struct` is only for `ImageRegistry` then please rename it to `ImageRegistryChecksumInput` or something.

legionus:
> we should be caching the resources we care about in a listwatcher, not getting them from the apiserver every time, so it should not be expensive. If we are not, that should be fixed.

Now it's not. To fix this we need another refactoring sometime in the future.

pkg/resource/generator.go
bparees: renamed.

bparees:
> Now it's not. To fix this we need another refactoring sometime in the future.

ok, I openend https://github.com/openshift/cluster-image-registry-operator/issues/94 to track it.  It's not the most urgent thing in my mind.

legionus:
@bparees please wait for https://github.com/openshift/cluster-image-registry-operator/pull/92

bparees:
replaced by https://github.com/openshift/cluster-image-registry-operator/pull/92

PR #90, opened 2018-11-27
create config namespaces as part of image registry operator bringup

bparees:


adambkaplan:
good practice to rename the prefixes so the ordering is clear?

bparees:
> good practice to rename the prefixes so the ordering is clear?

the ordering doesn't matter in this case, so renaming everything would be a waste.  and since this is a (hopefully) temporary home, i'm even less concerned.

bparees:
(In general my OCD would agree w/ you that i'd like everything to have an explicit ordering, but in practice we should probably get comfortable w/ having multiple things at the same priority if they do not have a dependency)

deploy/01-openshift-config-namespace.yaml
```diff
@@ -0,0 +1,10 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-config-managed
```
smarterclayton: What is this namespace?

smarterclayton: I understand the other one, what is this one?

PR #92, opened 2018-11-28
Recreate CR when it is deleted

dmage:
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1653563

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -140,7 +140,6 @@ func (c *Controller) CreateOrUpdateResources(cr *regopapi.ImageRegistry, modifie
 
 func (c *Controller) Handle(action string, o interface{}) {
 	object, ok := o.(metaapi.Object)
-
 	if !ok {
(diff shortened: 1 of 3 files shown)


pkg/operator/controller.go
```diff
@@ -163,21 +162,8 @@ func (c *Controller) Handle(action string, o interface{}) {
 
 	glog.V(1).Infof("Processing %s object %s", action, objectInfo)
 
-	if cr, ok := o.(*regopapi.ImageRegistry); ok {
-		dgst, err := resource.Checksum(cr.Spec)
```
bparees: if we're not going to look at this any more, should we remove the code that's setting it?

dmage:
Now it also fixes #93.

pkg/operator/controller.go
dmage: done

bparees:
e2e-aws-operator passed, manually merging.

Issue #91, opened 2018-11-28
Can't update empty volumeSource to use emptyDir


dmage:
The operator doesn't allow to change configuration from
```
  storage:
    filesystem: {}
```
to
```
  storage:
    filesystem:
      volumeSource:
        emptyDir: {}
```
because `storage type change is not supported: expected storage type filesystem, but got emptydir`.

/kind bug

bparees:
I think for now we need to remove the logic that prevents storage changes.

we can discuss it on scrum or elsewhere, but i'm coming to the conclusion that trying to block storage changes is going to be too painful in light of the need for users to potentially tweak settings like adding cloudfront to s3, or us generally getting storage wrong.

i'd rather start w/ it being allowed and in the future when we have more time we can figure out how to put sensible restrictions in place along with a proper way for users to get around those restrictions.

PR #96, opened 2018-11-29
Allow to change storage type

legionus:
Fixes https://github.com/openshift/cluster-image-registry-operator/issues/91


legionus:
@bparees Maybe we can get back to the idea of configuring the storage as `emptydir` nothing specified ?

bparees:
> @bparees Maybe we can get back to the idea of configuring the storage as emptydir if nothing specified ?

maybe, but for now i'd prefer not to because if someone sets up a "real" cluster and we accidentally/intentionally default to emptydir, we are setting that user up for a bad experience when they lose all their blobs, if they don't realize they're on emptydir storage.  Libvirt is currently the only environment where I feel ok if you lose your data.

/lgtm

bparees:
If we want to default it to something, I could see defaulting it to filesystem/PVC storage.

dmage:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-image-registry-operator/96/pull-ci-openshift-cluster-image-registry-operator-master-e2e-aws-operator/233
/retest

PR #98, opened 2018-11-29
Add tests for Failing status and for basic logs

dmage:


dmage:
/hold

Ok, recent changes has made impossible to get into the Failing state, all configurations are valid.

dmage:
Added a new check that Replicas >= 0.

/hold cancel

PR #105, opened 2018-12-01
Fix test

dmage:


bparees:
go for loop behavior strikes again.

/lgtm

Issue #75, opened 2018-11-20
switch from  clusteroperators.operatorstatus.openshift.io to clusteroperator.config.openshift.io


bparees:
We should be creating/maintaining the new `clusteroperator.config.openshift.io` resource, not the old `clusteroperators.operatorstatus.openshift.io` one.

Also clusteroperator.config.openshift.io is not namespaced.

bparees:
and add an e2e test that confirms this is getting set.

bparees:
Needs additional RBAC permissions to update clusteroperator/status

bparees:
it's gone from clusteroperator.operatorstatus.openshift.io, but it's not being created in clusteroperator.config.openshift.io:

```
$ oc get clusteroperator.config.openshift.io
NAME                                                      VERSION                                    AVAILABLE   PROGRESSING   FAILING   SINCE
machine-api-operator                                      v0.0.0-was-not-built-properly              True                                1m
machine-config-operator                                   3.11.0-352-g0cfc4183-dirty                 True        False         False     12m
openshift-cluster-kube-apiserver-operator                                                                                      False     
openshift-cluster-kube-controller-manager-operator                                                                             True      
openshift-cluster-kube-scheduler-operator                                                            True        False         False     58s
openshift-cluster-openshift-apiserver-operator                                                       True        False         False     1m
openshift-cluster-openshift-controller-manager-operator                                                                                  
openshift-cluster-samples-operator                        fbe415f08bd5afdf65b130a14aa410792982e1b1   True        False         False     4h
openshift-ingress-operator                                                                           True        False         False     4h
```

```
installer (master)$ oc get clusteroperator.operatorstatus.openshift.io --all-namespaces
NAMESPACE                        NAME                            VERSION   AVAILABLE   PROGRESSING   SINCE
openshift-cluster-dns-operator   openshift-dns                   0.0.1     True        False         4h
openshift-service-cert-signer    openshift-service-cert-signer               
```

This is why I wanted a test :)

bparees:
I see two issues, working on it.

PR #106, opened 2018-12-01
S3 config should not stomp local secret config

### LINKED ISSUE #79: s3 bucket management
Follow up to https://github.com/openshift/cluster-image-registry-operator/pull/52 to resolve the items described in:
https://github.com/openshift/cluster-image-registry-operator/pull/52#discussion_r234775884

You need to ensure that on re-sync, we are making sure the bucket exists. That means a few things:

* Adding a condition on the registry resource that indicates when we have successfully created the bucket (we should not be deploying the registry until we are sure we created the bucket)
* a mechanism for ensuring that if the S3 bucket name changed, the condition is cleared until we create the new bucket.

### LINKED ISSUE #99: cluster s3 config should not stomp local secret config
Today we copy s3 config from the cluster, into the secret in the image-registry namespace, and then drive that into the registry deployment.

Instead the registry operator should be:

1) checking if there is a local secret.  If so use it.
2) If not, check if there is cluster-wide s3 config, if so, use it directly (set it on the registry deployment, don't copy it into a local secret)

We also need to be watching (2) for changes/resyncing it when there is no local secret.

coreydaley:
Cluster S3 configuration should not stomp user configuration
    
  - Checks for a user supplied S3 configuration to use
  - If user config is not found, use the system configuration
  - Add a test to ensure that a user provided configuration will override the sytem provided one
  - Add image-registry-private-configuration-user to the README.md
  - Finalize S3 storage (remove bucket if managed by us)
  - Don't fail on S3 bucket tagging and encryption, just set the appropriate conditions to false

Fixes #79 
Fixes #99


pkg/clusterconfig/clusterconfig.go
```diff
@@ -106,12 +113,21 @@ func GetAWSConfig() (*Config, error) {
...
+		// If no user defined secret is found, use the system one
+		sec, err = client.Secrets(installerConfigNamespace).Get(installerAWSCredsName, metav1.GetOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("unable to get secret  \"%s/%s\": %v", installerConfigNamespace, installerAWSCredsName, err)
+		}
+	} else if err != nil {
+		return nil, err
 	}
 
-	cfg.Storage.Type = StorageTypeS3
```
coreydaley: Removing this since it is defined earlier in the function.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -25,6 +30,8 @@ const (
 	installerConfigNamespace = "kube-system"
 	installerConfigName      = "cluster-config-v1"
 	installerAWSCredsName    = "aws-creds"
+	imageRegistryNamespace   = "openshift-image-registry"
```
bparees: we don't have this somewhere else already?

pkg/clusterconfig/clusterconfig.go
```diff
@@ -25,6 +30,8 @@ const (
...
+	userPrivateConfigName    = "image-registry-private-configuration-user"
```
bparees: where's the `image-registry-private-configuration` constant defined?  can be put this with that or vice versa?

pkg/storage/s3/s3.go
```diff
@@ -122,12 +122,12 @@ func (d *driver) createAndTagBucket(svc *s3.S3, installConfig *installer.Install
...
-
-	data["REGISTRY_STORAGE_S3_ACCESSKEY"] = accessKey
-	data["REGISTRY_STORAGE_S3_SECRETKEY"] = secretKey
+	data := map[string]string{
+		"REGISTRY_STORAGE_S3_ACCESSKEY": accessKey,
+		"REGISTRY_STORAGE_S3_SECRETKEY": secretKey,
+	}
 
-	return util.CreateOrUpdateSecret("image-registry", "openshift-image-registry", data)
+	return util.CreateOrUpdateSecret("image-registry-private-configuration", "openshift-image-registry", data)
```
bparees: constant

test/e2e/aws_test.go
```diff
@@ -184,4 +185,22 @@ func TestAWS(t *testing.T) {
...
+		t.Errorf("unable to create secret \"openshift-image-registry/image-registry-configuration-user\": %#v", err)
+	}
+
+	cfgUser, err := clusterconfig.GetAWSConfig()
+	if err != nil {
+		t.Errorf("unable to get aws configuration: %#v", err)
+	}
+	if fakeAWSCredsData["REGISTRY_STORAGE_S3_ACCESSKEY"] != cfgUser.Storage.S3.AccessKey || fakeAWSCredsData["REGISTRY_STORAGE_S3_SECRETKEY"] != cfgUser.Storage.S3.SecretKey {
+		t.Errorf("expected system configuration to be overridden by the user configuration but it wasn't.")
+	}
```
bparees: since you've gotten this far, you might as well add a test that confirms we report a useful status condition when you can't create the s3 bucket because the credentials are invalid.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -102,21 +109,38 @@ func GetAWSConfig() (*Config, error) {
...
+		if v, ok := sec.Data["aws_access_key_id"]; ok {
+			cfg.Storage.S3.AccessKey = string(v)
+		}
+		if v, ok := sec.Data["aws_secret_access_key"]; ok {
+			cfg.Storage.S3.SecretKey = string(v)
+		}
+	} else if err != nil {
+		return nil, err
+	} else {
+		glog.Infof("Using user defined AWS credentials from secret \"%s/%s\"", imageRegistryNamespace, userPrivateConfigName)
```
bparees: need to handle watching this secret in case the value is changed (or the secret is introduced later.  or deleted later).

pkg/clusterconfig/clusterconfig.go
coreydaley: It wasn't, but i added it in the next push

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/storage/s3/s3.go

pkg/clusterconfig/clusterconfig.go
coreydaley: Should I introduce that code directly into https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/operator/controller.go#L162 which runs periodically , or should it be some kind of separate watcher?

dmage: @coreydaley it handles events from watchers: https://github.com/openshift/cluster-image-registry-operator/blob/83f0b8d3eddf2175496f915b7c095baa1d7f57a9/pkg/operator/controller.go#L282-L292

test/e2e/aws_test.go
```diff
@@ -184,4 +185,22 @@ func TestAWS(t *testing.T) {
...
 
+	// Check that the aws access key and secret key in the system configuration
+	// can be overridden by a user defined secret
+	fakeAWSCredsData := map[string]string{
+		"REGISTRY_STORAGE_S3_ACCESSKEY": "myAccessKey",
+		"REGISTRY_STORAGE_S3_SECRETKEY": "mySecretKey",
+	}
+
+	if err := util.CreateOrUpdateSecret("image-registry-private-configuration-user", "openshift-image-registry", fakeAWSCredsData); err != nil {
+		t.Errorf("unable to create secret \"openshift-image-registry/image-registry-configuration-user\": %#v", err)
```
dmage: Why isn't it a fatal error?

pkg/clusterconfig/clusterconfig.go
```diff
@@ -102,21 +109,38 @@ func GetAWSConfig() (*Config, error) {
...
+		if v, ok := sec.Data["REGISTRY_STORAGE_S3_ACCESSKEY"]; ok {
```
dmage: What if the user has made a typo and the key isn't present in the secret?

pkg/clusterconfig/clusterconfig.go
```diff
@@ -102,21 +109,38 @@ func GetAWSConfig() (*Config, error) {
...
-	if v, ok := sec.Data["aws_secret_access_key"]; ok {
-		cfg.Storage.S3.SecretKey = string(v)
+	// Look for a user defined secret to get the AWS credentials from first
+	sec, err := client.Secrets(imageRegistryNamespace).Get(userPrivateConfigName, metav1.GetOptions{})
+	if err != nil && errors.IsNotFound(err) {
+		glog.Infof("Optional user defined AWS credentials in secret \"%s/%s\" not found, ignoring.", imageRegistryNamespace, userPrivateConfigName)
+		// If no user defined secret is found, use the system one
+		sec, err = client.Secrets(installerConfigNamespace).Get(installerAWSCredsName, metav1.GetOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("unable to get secret  \"%s/%s\": %v", installerConfigNamespace, installerAWSCredsName, err)
```
dmage: two spaces, and it's better to use %q to get a quoted string

pkg/clusterconfig/clusterconfig.go
```diff
@@ -102,21 +109,38 @@ func GetAWSConfig() (*Config, error) {
...
+		}
+		glog.Infof("Using cluster defined AWS credentials from secret \"%s/%s\"", installerConfigNamespace, installerAWSCredsName)
```
dmage: I don't think we want to have two info messages on each event. If it's important to indicate which configuration is in use, it's better to use status.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -102,21 +109,38 @@ func GetAWSConfig() (*Config, error) {
...
+		if v, ok := sec.Data["aws_access_key_id"]; ok {
```
dmage: There should be a warning or an error if the key isn't in the secret.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -102,21 +109,38 @@ func GetAWSConfig() (*Config, error) {
...
+			cfg.Storage.S3.SecretKey = string(v)
+		}
+	} else if err != nil {
+		return nil, err
+	} else {
+		glog.Infof("Using user defined AWS credentials from secret \"%s/%s\"", imageRegistryNamespace, userPrivateConfigName)
+		if v, ok := sec.Data["REGISTRY_STORAGE_S3_ACCESSKEY"]; ok {
+			cfg.Storage.S3.AccessKey = string(v)
+		}
+		if v, ok := sec.Data["REGISTRY_STORAGE_S3_SECRETKEY"]; ok {
```
legionus: Please check that the `image-registry-private-configuration-user` does not contain any unknown keys to avoid typos.

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/apis/imageregistry/v1alpha1/types.go
```diff
@@ -10,7 +10,10 @@ import (
 )
 
 const (
-	OperatorStatusTypeRemoved = "Removed"
+	OperatorStatusTypeRemoved             = "Removed"
+	OperatorNamespace                     = "openshift-image-registry"
```
legionus: No. Use: 
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/client/client.go#L65-L74

pkg/clusterconfig/clusterconfig.go
```diff
@@ -116,30 +115,37 @@ func GetAWSConfig() (*Config, error) {
...
 		sec, err = client.Secrets(installerConfigNamespace).Get(installerAWSCredsName, metav1.GetOptions{})
 		if err != nil {
-			return nil, fmt.Errorf("unable to get secret  \"%s/%s\": %v", installerConfigNamespace, installerAWSCredsName, err)
+			return nil, fmt.Errorf("unable to get secret %q: %v", fmt.Sprintf("%s/%s", installerConfigNamespace, installerAWSCredsName), err)
 		}
-		glog.Infof("Using cluster defined AWS credentials from secret \"%s/%s\"", installerConfigNamespace, installerAWSCredsName)
 		if v, ok := sec.Data["aws_access_key_id"]; ok {
 			cfg.Storage.S3.AccessKey = string(v)
+		} else {
+			return nil, fmt.Errorf("Secret %q does not contain required key \"aws_access_key_id\"", fmt.Sprintf("%s/%s", regopapi.OperatorNamespace, regopapi.ImageRegistryPrivateConfiguration))
```
dmage: sec is a secret from the installer.

pkg/clusterconfig/clusterconfig.go
bparees: yeah it should be part of the existing controller which is already watching secrets anyway, you just need to add the logic to handle *this* secret.

pkg/clusterconfig/clusterconfig.go
bparees: ?

dmage: The error mentions `fmt.Sprintf("%s/%s", regopapi.OperatorNamespace, regopapi.ImageRegistryPrivateConfiguration)`, but the lookup was in the installer secret.

bparees: ah yes, good catch!

pkg/apis/imageregistry/v1alpha1/types.go
coreydaley: Did you mean https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/client/client.go#L56 to get the namespace?

pkg/clusterconfig/clusterconfig.go
coreydaley: updated in next push

pkg/apis/imageregistry/v1alpha1/types.go
coreydaley: Fixed in next push.

coreydaley:
Blocked by https://github.com/openshift/cluster-image-registry-operator/issues/75

coreydaley:
@bparees ptal

test/e2e/aws_test.go
coreydaley: updated in current push

pkg/resource/generator.go
```diff
@@ -86,8 +96,132 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
...
+		return err
+	}
+
+	cfg, err := clusterconfig.GetAWSConfig()
+	if err != nil {
+		return err
+	}
+
+	// Get the existing image-registry-private-configuration secret
+	sec, err := client.Secrets(g.params.Deployment.Namespace).Get(regopapi.ImageRegistryPrivateConfiguration, metav1.GetOptions{})
```
bparees: as a rule for content we are already watching, and thus have cached in a listwatcher, we should retrieve it from the cache, not make a new api call.  There are a few spots in this PR that this applies to.  I'm not going to make you fix it in this PR, but it's something to keep in mind and come back to.

pkg/resource/generator.go
```diff
@@ -86,8 +96,132 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
...
+			driver.Config.Bucket = cr.Spec.Storage.S3.Bucket
+			// Check to see if the bucket exists
+			// and if we can access it
+			if err := driver.CheckBucketExists(cr); err != nil {
+				installConfig, err := clusterconfig.GetInstallConfig()
+				if err != nil {
+					return err
+				}
+				// If the bucket doesn't exist, try to create it
+				if err := driver.CreateAndTagBucket(installConfig, cr); err != nil {
```
bparees: if we create the bucket (after a name change, or basically ever) that has implications for the ManagedStorage value, so it needs to be updated, right?

(Suppose we started out using a bucket they created, so ManagedStorage=false, we won't delete the bucket on teardown...  then they change the bucketname to something else and we *do* create the bucket...now we own it.  If they change the bucket name to something else that already exists, we don't own it.  So you need to be keeping that state updated)

pkg/resource/generator.go
```diff
@@ -86,8 +96,132 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
...
+					return err
+				}
+				// If we got here, image registry resource
+				// was modified by CreateAndTagBucket
+				*modified = true
+			}
+		} else if cr.Spec.Storage.S3.Bucket != "" {
+			// If the bucket name didn't change,
+			// check to see if the current bucket exists
+			if err := driver.CheckBucketExists(cr); err != nil {
```
bparees: why aren't we attempting to create it if it doesn't exist?

pkg/resource/generator.go
```diff
@@ -86,8 +96,132 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
 	return mutators, nil
 }
 
+// syncStorage checks:
+// 1.)  to make sure that an existing storage medium still exists and we can access it
+// 2.)  to see if the storage medium name changed and we need to:
+//      a.) check to make sure that we can access the storage or
+//      b.) see if we need to try to create the new storage
+func (g *Generator) syncStorage(cr *regopapi.ImageRegistry, modified *bool) error {
```
bparees: We should only be managing the storage in one place, so the fact that you're doing this in both CompleteConfiguration and syncStorage is wrong to me.

Eventually consistent behavior would say the CompleteConfiguration itself should not be responsible for creating/managing the storage.  It can setup whatever initial configuration is needed, but then the standard sync/watch loop should see the configuration (either as created by CompleteConfiguration, or edited/provided by the user) and make the necessary storage checks/updates.

As soon as you start doing it from two different code paths you're creating assumptions about system state and initialization order.

coreydaley: Updated in next push

pkg/resource/generator.go
coreydaley: updated in next push

pkg/resource/generator.go
coreydaley: ok, i will keep that in mind in the future

coreydaley:
@bparees ptal
I refactored a bunch of stuff and addressed your comments.
I also merged in my other PR since I needed a few things from it in this one.

I am still working on the test, but I believe everything else is ready for another look.

pkg/resource/generator.go
coreydaley: This should be updated now.

pkg/storage/s3/s3.go
```diff
@@ -76,38 +103,127 @@ func (d *driver) Volumes() ([]corev1.Volume, []corev1.VolumeMount, error) {
...
+		}
+
+		if _, err := svc.CreateBucket(createBucketInput); err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				case s3.ErrCodeBucketAlreadyExists:
+					if cr.Spec.Storage.S3.Bucket != "" {
+						break
+					}
+					continue
```
bparees: potential infinite loops scare me.  Put a cap of 5000 on this or something and update the status/condition to indicate we failed to generate a unique bucket name.  They can always restart the operator if they want another 5000 runs at it.

pkg/storage/s3/s3.go
```diff
@@ -129,86 +253,68 @@ func (d *driver) createAndTagBucket(svc *s3.S3, installConfig *installer.Install
 	encryptionConfig := &s3.ServerSideEncryptionConfiguration{Rules: encryptionRules}
 	bucketEncryptionInput := &s3.PutBucketEncryptionInput{Bucket: aws.String(d.Config.Bucket), ServerSideEncryptionConfiguration: encryptionConfig}
 
-	_, err := svc.PutBucketEncryption(bucketEncryptionInput)
+	_, err = svc.PutBucketEncryption(bucketEncryptionInput)
 	if err != nil {
-		return err
+		if aerr, ok := err.(awserr.Error); ok {
+			// If we can't set default encryption on the bucket don't return an error
+			// just set the StorageEncrypted condition to false
```
bparees: what can the user/admin do to get encryption added to the bucket later?  Do/can they just go encrypt the bucket manually?  Does anything else need to be updated if they do that (e.g. on the registry configuration?).

Is/should there be a way for them to force the operator to attempt to encrypt the bucket again once they've resolved the issue?  (my inclination is to say yes, eventually, but probably not for 4.0).

pkg/resource/generator.go
```diff
@@ -86,8 +94,140 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
...
+	if err != nil {
+		return err
+	}
+
+	// If the storage medium name is blank, we need to create it
+	name, err := driver.GetStorageName(cr)
+	if err != nil {
+		return err
+	}
+	if name == "" {
```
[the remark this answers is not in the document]
coreydaley: updated in next push

pkg/resource/generator.go
```diff
@@ -86,8 +94,140 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
...
+		exists, err := driver.StorageExists(cr)
+		if err != nil {
+			return err
+		}
+		if !exists {
+			runCreate = true
+		}
+	}
+
+	exists, err := driver.StorageExists(cr)
```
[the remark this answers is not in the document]
coreydaley: updated in next push

pkg/resource/generator.go
```diff
@@ -168,6 +308,16 @@ func (g *Generator) Apply(cr *regopapi.ImageRegistry, modified *bool) error {
 		return fmt.Errorf("unable to remove obsolete routes: %s", err)
 	}
 
+	err = g.syncSecrets(cr, modified)
```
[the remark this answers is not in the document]
coreydaley: updated in next push

pkg/storage/s3/s3.go
```diff
@@ -10,23 +10,50 @@ import (
...
+	s3Service *s3.S3
+)
+
 type driver struct {
 	Name      string
 	Namespace string
 	Config    *opapi.ImageRegistryConfigStorageS3
 }
 
+func (d *driver) getSVC() (*s3.S3, error) {
```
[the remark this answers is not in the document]

pkg/storage/s3/s3.go
```diff
@@ -76,38 +103,127 @@ func (d *driver) Volumes() ([]corev1.Volume, []corev1.VolumeMount, error) {
...
-		Bucket: aws.String(d.Config.Bucket),
+// StorageChanged checks to see if the name of the storage medium
+// has changed
+func (d *driver) StorageChanged(cr *opapi.ImageRegistry) bool {
+	var storageChanged bool
+
+	if cr.Status.Storage.State.S3 == nil {
+		storageChanged = true
+	} else {
+		if cr.Status.Storage.State.S3.Bucket != cr.Spec.Storage.S3.Bucket {
```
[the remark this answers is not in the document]

pkg/storage/s3/s3.go

pkg/storage/s3/s3.go
```diff
@@ -76,38 +103,127 @@ func (d *driver) Volumes() ([]corev1.Volume, []corev1.VolumeMount, error) {
...
+					}
+					continue
+				default:
+					util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionFalse, aerr.Code(), aerr.Error())
+					return err
+				}
+			}
+		} else {
+			break
+		}
```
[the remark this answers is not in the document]

pkg/operator/finalizer.go
```diff
@@ -49,7 +50,16 @@ func (c *Controller) finalizeResources(o *regopapi.ImageRegistry) error {
...
+		return err
+	}
+	err = driver.RemoveStorage(o)
+	if err != nil {
+		return err
+	}
+
+	err = c.RemoveResources(o)
 	if err != nil {
 		errOp := c.clusterStatus.Update(osapi.OperatorFailing, osapi.ConditionTrue, "unable to remove registry")
```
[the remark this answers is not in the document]
coreydaley: updated in next push

pkg/resource/generator.go
```diff
@@ -86,8 +94,140 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
...
+	}
+
+	// Get the existing image-registry-private-configuration secret
+	sec, err := client.Secrets(g.params.Deployment.Namespace).Get(regopapi.ImageRegistryPrivateConfiguration, metav1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("unable to get secret %q: %v", fmt.Sprintf("%s/%s", g.params.Deployment.Namespace, regopapi.ImageRegistryPrivateConfiguration), err)
+	}
+
+	// If the Image Registry is currently using S3 storage
+	if cr.Spec.Storage.S3 != nil {
```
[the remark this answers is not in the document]
coreydaley: updated in next push

pkg/storage/s3/s3.go
coreydaley: The user could log into the aws console and turn on default encryption if they want to, and nothing in the registry config needs to be changed.

We could add an option to the config that says the storage should be encrypted, and we check to see if it's encrypted on sync and if it's not, we attempt to encrypt it, if the condition is false and shouldEncrypt is true or something.

test/e2e/basic_empty_dir_test.go
```diff
@@ -1,73 +1,73 @@
...
+// 		t.Error("error: the log doesn't contain the operator's version")
+// 	}
+// 	if !logs.Contains(regexp.MustCompile(`status changed`)) {
+// 		badlogs = true
+// 		t.Error("error: the log doesn't contain changes")
+// 	}
+// 	if badlogs {
+// 		testframework.DumpPodLogs(t, logs)
+// 	}
+// }
```
bparees: why is this entire test being disabled?

test/e2e/unmanaged_test.go
```diff
@@ -1,80 +1,80 @@
...
+// 		conds := testframework.GetImageRegistryConditions(cr)
+// 		t.Logf("image registry: %s", conds)
+// 		return conds.Available.IsFalse() && conds.Progressing.IsFalse(), err
+// 	})
+// 	if err != nil {
+// 		testframework.DumpImageRegistryResource(t, client)
+// 		testframework.DumpOperatorLogs(t, client)
+// 		t.Fatal(err)
+// 	}
+// }
```
bparees: and this one?

pkg/resource/generator.go
```diff
@@ -86,8 +93,125 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
...
+
+		// Make sure that the annotations map isn't nil!
+		if deployment.Spec.Template.Annotations == nil {
+			deployment.Spec.Template.Annotations = make(map[string]string)
+		}
+
+		// Update the deployment template with an annotation of the checksum of the secret
+		// to trigger the Image Registry to be redeployed
+		deployment.Spec.Template.Annotations[fmt.Sprintf("%s-checksum", upSec.ObjectMeta.Name)] = upSecChecksum
+		if _, err := appsclient.Deployments(g.params.Deployment.Namespace).Update(deployment); err != nil {
```
bparees: why isn't this done as part of the deployment generator/template?  Having two writers updating the deployment resource is liable to be problematic.

test/e2e/basic_empty_dir_test.go
coreydaley: Doh, didin't mean to commit that, i was just trying to run my test by itself so it was faster locally. fixed in next push

test/e2e/unmanaged_test.go
coreydaley: same, didn't mean to commit

pkg/resource/generator.go
coreydaley: Updated in current push

coreydaley:
@bparees ptal
This should be ready for another look.  I separated the aws-sdk-go bump into it's own commit for ease of review.

coreydaley:
@bparees bump

pkg/resource/podtemplatespec.go
```diff
@@ -111,6 +111,7 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, params *parameters.
 	}
 
 	deps := newDependencies()
+	deps.AddSecret(v1alpha1.ImageRegistryPrivateConfigurationUser)
```
bparees: shouldn't this be `v1alpha1.ImageRegistryPrivateConfiguration` ?

coreydaley: No, it's already watching that ImageRegistryPrivateConfiguration secret

coreydaley: Fixed in current push

PR #108, opened 2018-12-03
Add test for Unmanaged state

dmage:


test/e2e/unmanaged_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	err = wait.Poll(1*time.Second, testframework.AsyncOperationTimeout, func() (stop bool, err error) {
+		cr, err = client.ImageRegistries().Get(testframework.ImageRegistryName, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+
+		conds := testframework.GetImageRegistryConditions(cr)
+		t.Logf("image registry: %s", conds)
+		return conds.Available.IsFalse() && conds.Progressing.IsFalse(), err
```
bparees: why does marking the CR unmanaged result in the Available status going to false?

PR #109, opened 2018-12-04
[WIP] Rework resource generators

legionus:


legionus:
This PR is not needed because of https://github.com/openshift/cluster-image-registry-operator/pull/112

PR #110, opened 2018-12-04
use short hostname for registry

bparees:


bparees:
@openshift/sig-developer-experience fyi.  apparently we're supposed to use shortnames, not FQDN

PR #112, opened 2018-12-05
Copy CAs from openshift-config namespace

dmage:


dmage:
/hold

It doesn't work. It has problems with detecting changes (it gets empty object from the lister).

Issue #42, opened 2018-10-18
Handle Proxy field in image registry spec.


dmage:
We don't have handlers for the fields Proxy.HTTP and Proxy.HTTPS.

bparees:
Just needs to set the proxy value using the appropriate registry env var on the registry deployment.

PR #116, opened 2018-12-11
Use cr.Spec.Proxy.HTTP and cr.Spec.Proxy.HTTPS

legionus:
Fixes https://github.com/openshift/cluster-image-registry-operator/issues/42

### DIFF
--- a/pkg/resource/podtemplatespec.go
+++ b/pkg/resource/podtemplatespec.go
@@ -124,6 +124,8 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, params *parameters.
 	}
 
 	env = append(env,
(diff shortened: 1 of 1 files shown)


pkg/resource/podtemplatespec.go
```diff
@@ -124,6 +124,8 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, params *parameters.
 	}
 
 	env = append(env,
+		corev1.EnvVar{Name: "HTTP_PROXY", Value: cr.Spec.Proxy.HTTP},
+		corev1.EnvVar{Name: "HTTPS_PROXY", Value: cr.Spec.Proxy.HTTPS},
```
bparees: don't we need to add NO_PROXY also?

PR #118, opened 2018-12-11
use a status subresource for the image config api

bparees:


bparees:
Note that testing this locally was a pain because the CVO keeps stomping on my resource changes, so i'm going to let the PR job do it... may result in failing tests if i got something wrong here.

bparees:
@dmage note that we probably want to do this for the registry resource's status field too at some point, but there are too many people touching the registry operator status right now for me to want to touch that.

pkg/resource/imageconfig.go
```diff
@@ -71,9 +70,7 @@ func (gic *generatorImageConfig) Create() error {
 func (gic *generatorImageConfig) Update(o runtime.Object) (bool, error) {
 	ic := o.(*configapi.Image)
 
-	newObjectMeta := gic.objectMeta()
-
-	modified := strategy.Metadata(&ic.ObjectMeta, &newObjectMeta)
```
dmage: Removing this line means that our operator won't restore metadata (OwnerReferences) if someone changes it.

dmage:
@bparees yes, I know, I've told you on scrums that we need to do it.

deploy/02-rbac.yaml
```diff
@@ -93,7 +94,7 @@ metadata:
   name: default-account-cluster-image-registry-operator
 subjects:
 - kind: ServiceAccount
-  name: default
+  name: cluster-image-registry-operator
```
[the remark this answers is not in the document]
dmage: 04-operator.yaml needs to be updated to use this SA.

pkg/resource/imageconfig.go
bparees: this was somewhat intentional.  the image config cluster object should not be owned by the image registry operator, nor tied to the image-registry resource lifecycle. It represents configuration of things beyond the registry itself and if if a user customizes it, we don't want it deleted out from under them.

i'm going to update this PR to *not* set the ownerref when we create the object.

deploy/02-rbac.yaml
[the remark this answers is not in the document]
bparees: that change is already in this PR, no?

bparees:
> @bparees yes, I know, I've told you on scrums that we need to do it.

yes you have, sorry.

bparees:
@dmage updated the ownerref bit. ptal.

pkg/resource/imageconfig.go
bparees: updated.  the idea here is that this generator is responsible for only two things:

1) creating the object if it does not exist already (it may exist because a user can create it themselves, or something else could have created it. it's a global config object).
2) updating one specific status field within the object (status.internalregistryhostname).  it is not responsible for maintaining the object metadata, it does not own those fields.

bparees:
/hold
i'm going to add a test that confirms internalRegistryHostname is being set on the cluster config object.

bparees:
/hold cancel
test added

bparees:
/retest

@dmage ok i think this is ready for re-review

bparees:
actually @dmage is on PTO after today, so I guess @coreydaley  is up.
/assign @coreydaley

PR #120, opened 2018-12-13
switch from clusteroperators.operatorstatus.openshift.io to clusteroperator.config.openshift.io

legionus:
Fixes https://github.com/openshift/cluster-image-registry-operator/issues/75


bparees:
@legionus please add an e2e test that confirms the ClusterOperator status object is being created and has the correct conditions set on it, and then lgtm.

@coreydaley fyi.

legionus:
@bparees can I do it as follow up ?

bparees:
> @bparees can I do it as follow up ?

ok.

/lgtm

PR #122, opened 2018-12-14
A few small cleanup items

coreydaley:
Fixing a few small issues that I have run across:

- Update deprecated ImageRegistry() with ImageRegistryV1alpha1()
- Use snake case instead of CAMEL_CASE
- fix using %q when we shouldn't be
- variable name masks imported package name (and package was imported twice"

Each in it's own commit for easy review.


coreydaley:
@bparees ptal, should be a quick review

PR #123, opened 2018-12-14
update clusteroperator status properly and add rbac permission to do so

bparees:
fixes https://github.com/openshift/cluster-image-registry-operator/issues/75

### DIFF
--- a/deploy/02-rbac.yaml
+++ b/deploy/02-rbac.yaml
@@ -80,9 +80,10 @@ rules:
   verbs:
   - "*"
 - apiGroups:
-  - operatorstatus.openshift.io
+  - config.openshift.io
   resources:
   - clusteroperators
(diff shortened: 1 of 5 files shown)


bparees:
@coreydaley @legionus ptal

bparees:
(@legionus there is more testing to be done here to ensure we are updating the clusteroperator status correctly in other conditions, but at least this is a meets min to ensure we are populating it in the first place)

bparees:
/assign @coreydaley 

@coreydaley note that the registry resource is *also* still getting populated with status conditions (the same ones that are being put on the clusteroperator resource).  That's ok for now, though we probably want to rethink it at some point so that the registry resource has the more detailed status information and the clusteroperator has the more general/common-across-operators ones, rather than duplicating the same info.

PR #124, opened 2018-12-16
[WIP] Finalize s3 storage

coreydaley:
Remove the S3 bucket during finalization (if we created it)

### DIFF
--- a/pkg/operator/finalizer.go
+++ b/pkg/operator/finalizer.go
@@ -16,6 +16,7 @@ import (
 
 	regopapi "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1alpha1"
 	regopset "github.com/openshift/cluster-image-registry-operator/pkg/generated/clientset/versioned/typed/imageregistry/v1alpha1"
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage"
 
 	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
 )
@@ -49,7 +50,16 @@ func (c *Controller) finalizeResources(o *regopapi.ImageRegistry) error {
 
 	glog.Infof("finalizing %s", objectInfo(o))
 
-	err := c.RemoveResources(o)
+	driver, err := storage.NewDriver(o.Name, o.Namespace, &o.Spec.Storage)
+	if err != nil {
+		return err
+	}
+	err = driver.RemoveStorage(o)
+	if err != nil {
+		return err
+	}
+
+	err = c.RemoveResources(o)
 	if err != nil {
 		errOp := c.clusterStatus.Update(osapi.OperatorFailing, osapi.ConditionTrue, "unable to remove registry")
 		if errOp != nil {
--- a/pkg/storage/emptydir/emptydir.go
+++ b/pkg/storage/emptydir/emptydir.go
@@ -1,9 +1,11 @@
 package emptydir
 
 import (
+	"fmt"
 	corev1 "k8s.io/api/core/v1"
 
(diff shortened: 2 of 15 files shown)


pkg/storage/emptydir/emptydir.go
```diff
@@ -51,9 +53,25 @@ func (d *driver) Volumes() ([]corev1.Volume, []corev1.VolumeMount, error) {
...
+	return nil
+}
+
+func (d *driver) CreateStorage(cr *opapi.ImageRegistry) error {
+	return nil
+}
+
+func (d *driver) RemoveStorage(cr *opapi.ImageRegistry) error {
+	if !cr.Status.Storage.Managed {
+		return fmt.Errorf("storage is not managed by the image registry operator, so we can't delete it.")
```
bparees: are we currently setting Managed=true for emptydir storage?  Basically we should either always set it to true, or this should not care whether it is managed or not (since removing storage for emptydir is a no-op anyway)

pkg/storage/s3/s3.go
```diff
@@ -129,43 +210,74 @@ func (d *driver) createAndTagBucket(svc *s3.S3, installConfig *installer.Install
...
+	if err != nil {
+		return err
+	}
+	_, err = svc.DeleteBucket(&s3.DeleteBucketInput{
+		Bucket: aws.String(d.Config.Bucket),
+	})
+
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionTrue, aerr.Code(), aerr.Error())
```
bparees: it's probably condition unknown at this point... we don't know if the bucket got deleted or not, given that we got an error.

test/e2e/aws_test.go
```diff
@@ -185,4 +198,69 @@ func TestAWS(t *testing.T) {
...
+	if !foundStorageExists {
+		t.Errorf("condition StorageExists was not found, but should have been. %#v", cr.Status.Conditions)
+	}
+
+	// Clean up the image-registry-private-configuration-user secret
+	err = client.Secrets(testframework.ImageRegistryDeploymentNamespace).Delete(regopapi.ImageRegistryPrivateConfigurationUser, &metav1.DeleteOptions{})
+	if err != nil {
+		t.Errorf("unable to remove %s/%s: %#v", testframework.ImageRegistryDeploymentNamespace, regopapi.ImageRegistryPrivateConfigurationUser, err)
+	}
+
```
bparees: need a test that confirms when we delete the registry resource, the finalizer cleans up all resources + the s3 buckets.

pkg/operator/finalizer.go
```diff
@@ -49,7 +50,16 @@ func (c *Controller) finalizeResources(o *regopapi.ImageRegistry) error {
...
 
-	err := c.RemoveResources(o)
+	driver, err := storage.NewDriver(o.Name, o.Namespace, &o.Spec.Storage)
+	if err != nil {
+		return err
+	}
+	err = driver.RemoveStorage(o)
+	if err != nil {
+		return err
+	}
```
bparees: you've got storage drivers returning an error when storage is not managed, which means for any user who created their own s3 buckets (and thus has unmanaged storage), they'll never be able to finalize the registry.  That's bad.

If the storage is unmanaged, I think RemoveStorage() should just return success.. it did what it's supposed to do, with respect to the storage.

If you want to maintain a distinction between removing unmanaged and removing managed storage, then you need another return value, or you need to check the error type so that we can proceed in the face of "storage is not owned by us so we couldn't remove it" errors.

coreydaley:
Pulled changes into my other PR, closing this one.

PR #127, opened 2018-12-19
Use Cluster Version ClusterID instead of Install Config ClusterID

coreydaley:
This is a temporary patch until #106 merges


coreydaley:
@bparees ptal
This will get refactored some in #106

coreydaley:
Failure due to maximum number of VPCs exceeded on AWS
@smarterclayton @bparees fyi

bparees:
/lgtm 

thanks for jumping on this @coreydaley

PR #131, opened 2019-01-03
Update to openshift/api/v1 deps

coreydaley:


config/config.yaml
```diff
@@ -1,3 +1,3 @@
-apiVersion: imageregistry.operator.openshift.io/v1alpha1
+apiVersion: imageregistry.operator.openshift.io/v1
```
bparees: why wasn't this part of your other PR?  (and how did your other PR pass tests w/o this, or did it?)

bparees: nm, i see this is from the other commit you pulled in.

coreydaley:
@bparees ptal
I split out the vendor bump and the code changes into two commits

pkg/apis/imageregistry/v1/types.go
```diff
@@ -90,7 +90,7 @@ type ImageRegistryConfigRoute struct {
 }
 
 type ImageRegistrySpec struct {
-	operatorsv1alpha1api.OperatorSpec `json:",inline"`
+	operatorsv1api.OperatorSpec `json:",inline"`
```
bparees: what are we still using from this? can we get rid of it entirely?

coreydaley: Checking...

bparees:
squash the last 2 commits and lgtm.

coreydaley:
@bparees squashed and pushed, thanks!

PR #132, opened 2019-01-03
a small refactor / cleanup

coreydaley:
- variable name should not mask imported package name
 - errors should not start with a capital letter or end with punctuation
 - update to new function name (previous was deprecated)
 - refactor storage finalizer to retry on conflict with a new version of the image registry resource and update the test for the s3 finalizer

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -141,7 +141,7 @@ func (d *driver) SyncSecrets(sec *coreapi.Secret) (map[string]string, error) {
 // StorageExists checks if an S3 bucket with the given name exists
 // and we can access it
 func (d *driver) StorageExists(cr *opapi.ImageRegistry) (bool, error) {
-	if d.Config.Bucket == "" {
+	if len(cr.Status.Storage.State.S3.Bucket) == 0 {
 		return false, nil
 	}
 
@@ -150,7 +150,7 @@ func (d *driver) StorageExists(cr *opapi.ImageRegistry) (bool, error) {
 		return false, err
 	}
 	_, err = svc.HeadBucket(&s3.HeadBucketInput{
-		Bucket: aws.String(d.Config.Bucket),
+		Bucket: aws.String(cr.Status.Storage.State.S3.Bucket),
 	})
 
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/storage/s3/s3.go
```diff
@@ -351,7 +351,7 @@ func (d *driver) RemoveStorage(cr *opapi.ImageRegistry) error {
 		return err
 	}
 	_, err = svc.DeleteBucket(&s3.DeleteBucketInput{
-		Bucket: aws.String(d.Config.Bucket),
+		Bucket: aws.String(cr.Spec.Storage.S3.Bucket),
```
bparees: should use cr.Status's storage config info?

coreydaley: yep, updated

pkg/storage/s3/s3.go
```diff
@@ -351,7 +351,7 @@ func (d *driver) RemoveStorage(cr *opapi.ImageRegistry) error {
 		return err
 	}
 	_, err = svc.DeleteBucket(&s3.DeleteBucketInput{
-		Bucket: aws.String(d.Config.Bucket),
+		Bucket: aws.String(cr.Status.Storage.State.S3.Bucket),
```
bparees: what if it's not populated for some reason?

coreydaley: Then the below code would update the condition on the image registry resource with the code & message from aws

coreydaley: actually, when you initiate the driver, d.Config.Bucket is set from cr.Storage.S3.Bucket, so using the Status one is probably wrong here.

bparees: > Then the below code would update the condition on the image registry resource with the code & message from aws

seems silly to ask aws to tear down an empty bucket name and then report AWS' error to the user.  we can do better.

> actually, when you initiate the driver, d.Config.Bucket is set from cr.Storage.S3.Bucket, so using the Status one is probably wrong here.

Well status should represent "the bucket that we actually know exists because we created it".  Spec is "the bucket we intended to create".  So status seems more correct to me.

coreydaley: updated

pkg/storage/s3/s3.go
```diff
@@ -344,14 +344,16 @@ func (d *driver) CreateStorage(cr *opapi.ImageRegistry) error {
 func (d *driver) RemoveStorage(cr *opapi.ImageRegistry) error {
 	if !cr.Status.Storage.Managed {
 		return nil
+	} else if len(cr.Status.Storage.State.S3.Bucket) == 0 {
+		return fmt.Errorf("unable to remove storage, no bucket name available in image registry resource")
```
bparees: is this going to block finalization?  I don't think we want to block finalization over this.

coreydaley: good point, updated

PR #133, opened 2019-01-03
be more judicious when setting modified=true

bparees:
setting "modified=true" everytime we ran syncStorage meant we kept triggering ourselves infinitely.

(we'd update the CR, which would result in an event, which would call syncStorage, which would set modified=true which meant we'd update the CR, and repeat).

Another way to avoid this would be to diff the new/old object and ignore the event if the only change is the resourceVersion, but I think this is the "right" way to solve it.

### DIFF
--- a/pkg/resource/generator.go
+++ b/pkg/resource/generator.go
@@ -97,7 +97,6 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
 //      b.) see if we need to try to create the new storage
 func (g *Generator) syncStorage(cr *regopapi.ImageRegistry, modified *bool) error {
 	var runCreate bool
-	*modified = true
 	// Create a driver with the current configuration
 	driver, err := storage.NewDriver(cr.Name, cr.Namespace, &cr.Spec.Storage)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


bparees:
@sjenning fyi

/assign @coreydaley 

@dmage @legionus I also want us to follow up with a discussion about this:
https://github.com/openshift/cluster-image-registry-operator/blob/6e2fee7776e3fc671cc90fe783000095e6927fad/pkg/resource/generator.go#L233-L248

because it doesn't make sense to me that we are setting "modified=true" (which implies the CR needs to be updated) when all we did is update/create secondary resources, *not* update the CR.

bparees:
I opened https://github.com/openshift/cluster-image-registry-operator/issues/134 to track the more general need to clean up how we are handling "modified"

pkg/resource/generator.go
```diff
@@ -97,30 +97,29 @@ func (g *Generator) list(cr *regopapi.ImageRegistry) ([]Mutator, error) {
 //      b.) see if we need to try to create the new storage
 func (g *Generator) syncStorage(cr *regopapi.ImageRegistry, modified *bool) error {
 	var runCreate bool
-	*modified = true
```
bparees: this was the root cause of the hot looping.

coreydaley:
error: unable to connect to image repository registry.svc.ci.openshift.org/ci-op-9qf57gxr/stable@sha256:86f6104f0b2d69a5c0133fa92f2325bbc78c398ba1000f10927f3a9d3b36e792: Get https://registry.svc.ci.openshift.org/v2/: net/http: TLS handshake timeout

sjenning:
apirequests
This didn't seem to fix the API request flood from this operator

PR #136, opened 2019-01-04
initialize configlistener field

bparees:


bparees:
@dmage this was panic'ing the operator when a configmap change happened

PR #137, opened 2019-01-05
Refactoring the S3 storage driver

coreydaley:


pkg/operator/finalizer.go
```diff
@@ -65,13 +65,21 @@ func (c *Controller) finalizeResources(o *regopapi.ImageRegistry) error {
 		if cr == nil {
 			cr, err = client.ImageRegistries().Get(o.Name, metav1.GetOptions{})
 			if err != nil {
-				return fmt.Errorf("failed to get %s: %s", objectInfo(o), err)
+				return fmt.Errorf("failed to get %s, retrying: %#v", objectInfo(o), err)
```
bparees: %v for errors, not %#v.

pkg/operator/finalizer.go
```diff
@@ -65,13 +65,21 @@ func (c *Controller) finalizeResources(o *regopapi.ImageRegistry) error {
...
+
+		if err := driver.RemoveStorage(o, &modified); err != nil {
 			cr = nil
-			return err
+			return fmt.Errorf("unable to remove storage, retrying: %#v", err)
+		}
+		if modified {
+			_, err := client.ImageRegistries().Update(o)
+			if err != nil {
+				return fmt.Errorf("unable to update %s, retrying: %#v", objectInfo(o), err)
```
bparees: %v

pkg/storage/s3/s3.go
```diff
@@ -176,27 +246,33 @@ func (d *driver) StorageExists(cr *opapi.ImageRegistry, modified *bool) (bool, e
 func (d *driver) StorageChanged(cr *opapi.ImageRegistry, modified *bool) bool {
 	var storageChanged bool
 
-	if cr.Status.Storage.State.S3 == nil || cr.Status.Storage.State.S3.Bucket != cr.Spec.Storage.S3.Bucket {
+	if cr.Status.Storage.State.S3 == nil || cr.Spec.Storage.S3 == nil {
+		storageChanged = true
```
bparees: isn't it only changed if one is nil and one isn't?  if they're both nil, it's unchanged.

pkg/storage/s3/s3.go
```diff
@@ -211,35 +287,67 @@ func (d *driver) CreateStorage(cr *opapi.ImageRegistry, modified *bool) error {
...
+				util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionFalse, aerr.Code(), aerr.Error(), modified)
+			default:
+				util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionUnknown, "Unknown Error Occurred", err.Error(), modified)
+				return err
+			}
 		}
+	}
+	if len(cr.Spec.Storage.S3.Bucket) != 0 && bucketExists {
+		d.Config.Bucket = cr.Spec.Storage.S3.Bucket
+		cr.Status.Storage.State.S3 = d.Config
```
bparees: How did we end up with "Status.Storage.State.S3" instead of just "Status.Storage.S3" to match the spec struct?

pkg/storage/s3/s3.go
```diff
@@ -66,11 +69,37 @@ func NewDriver(crname string, crnamespace string, c *opapi.ImageRegistryConfigSt
 	}
 }
 
-// GetType returns the name of the storage driver
+// NewDriverFromResource creates a new driver from an existing image registry resource
+func NewDriverFromResource(cr *opapi.ImageRegistry) (*driver, error) {
+	if cr.Status.Storage.State.S3 == nil {
```
bparees: why is this looking at status instead of spec?

I would expect the following sequence of events during bootstrap:

1) we identify we want to use s3 storage
2) we invoke NewDriver to create an s3 driver
3) we use the driver to populate/initialize spec.storage.s3 in the CR we are bootstrapping
4) we create the CR
5) the controller sees the CR, the driver gets invoked to turn the requested "spec.storage.s3" into a storage that exists
6) the driver, after creating the storage, updates status.storage.s3 with the details of the storage it created, and updates the condition storagexists to true

And of course in cases where the user creates a CR themselves, the flow would be:
1) the controller sees the CR, the driver gets invoked to turn the requested "spec.storage.s3" into a storage that exists
2) the driver, after creating the storage, updates status.storage.s3 with the details of the storage it created, and updates the condition storagexists to true

pkg/storage/s3/s3.go
coreydaley: If they are both nil, don't we need to go ahead and set up some sort of storage?

pkg/storage/s3/s3.go
coreydaley: We have Status.Storage.Managed, which isn't part of the configuration (but it might make sense for it to be, so I can refactor that)

coreydaley: Updated in current push

pkg/storage/s3/s3.go
coreydaley: I guess technically this should never be the case, as bootstrap should put *something* into at least the state ...

pkg/storage/s3/s3.go
bparees: If the current status.storage is GCS, but the spec.storage is S3, then what's going to happen when you invoke this function with the CR?  How will the operator move forward to transition onto s3 storage?

The one case where you're doing this that makes sense is the finalizer.... when tearing down, you want to tear down the current state, not the requested state.  But in the generator it does not make sense.  

What you should probably do is just pass a "Storage" object to NewDriverFromStorage().  Then depending on context you can invoke NewDriveFromStorage(spec.Storage) or NewDriverFromStorage(status.Storage).

And CreateStorage/etc should just look at the storage object that was passed to NewDriverFromStorage, rather than being hardcoded to look at spec.

pkg/storage/s3/s3.go
coreydaley: This should be sorted out now

pkg/storage/s3/s3.go
coreydaley: ok, this should be sorted out now, thanks!

pkg/operator/controller.go
```diff
@@ -206,7 +206,7 @@ func (c *Controller) eventProcessor() {
 
 			if _, ok := obj.(string); !ok {
 				c.workqueue.Forget(obj)
-				glog.Errorf("expected string in workqueue but got %#v", obj)
+				glog.Errorf("expected string in workqueue but got %v", obj)
```
bparees: here i think we want %#v.

pkg/storage/s3/s3.go
```diff
@@ -330,39 +412,37 @@ func (d *driver) CreateStorage(cr *opapi.ImageRegistry, modified *bool) error {
...
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok {
 			util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionUnknown, aerr.Code(), aerr.Error(), modified)
 		}
+		return err
 	}
 
+	cr.Spec.Storage.S3.Bucket = ""
+	d.Config.Bucket = ""
+	cr.Status.Storage.S3 = d.Config
```
bparees: need to set *modified=true if any of these represent a change in value.

pkg/storage/s3/s3.go
```diff
@@ -372,15 +452,13 @@ func (d *driver) CompleteConfiguration(cr *opapi.ImageRegistry, modified *bool)
...
-
 	if len(d.Config.Region) == 0 {
 		d.Config.Region = cfg.Storage.S3.Region
 	}
 
-	cr.Status.Storage.State.S3 = d.Config
-	*modified = true
+	cr.Status.Storage.S3 = d.Config
+
+	d.CreateStorage(cr, modified)
```
bparees: why are you invoking create storage here?  I think the pattern we want is:

1) completeconfiguration is used to populate a boostrapped CR
2) then the operator sees the bootstrapped CR (or user provided CR) and makes it happen

we shouldn't tie "bootstrapping the CR" together with actually creating the storage, i don't think.

test/e2e/basic_empty_dir_test.go
```diff
@@ -37,6 +37,15 @@ func TestBasicEmptyDir(t *testing.T) {
...
 		},
+		Status: imageregistryapi.ImageRegistryStatus{
+			Storage: imageregistryapi.ImageRegistryConfigStorage{
+				Filesystem: &imageregistryapi.ImageRegistryConfigStorageFilesystem{
+					VolumeSource: corev1.VolumeSource{
+						EmptyDir: &corev1.EmptyDirVolumeSource{},
+					},
+				},
+			},
+		},
```
bparees: why does this need to be passed in?

pkg/operator/controller.go
coreydaley: updated in next push

pkg/storage/s3/s3.go

test/e2e/basic_empty_dir_test.go
coreydaley: I can remove that, i needed it before when i hard coded the NewDriverFromResource, will fix in all the tests

coreydaley: updated in current push

pkg/storage/s3/s3.go
```diff
@@ -439,6 +439,7 @@ func (d *driver) RemoveStorage(cr *opapi.ImageRegistry, modified *bool) error {
 	cr.Spec.Storage.S3.Bucket = ""
 	d.Config.Bucket = ""
 	cr.Status.Storage.S3 = d.Config
+	*modified = true
```
bparees: no.  you need to actually check if you are changing the value.  if you do this, you're creating hotloops again.

pkg/resource/generator.go
```diff
@@ -103,7 +103,9 @@ func (g *Generator) syncStorage(cr *regopapi.ImageRegistry, modified *bool) erro
 		return err
 	}
 
-	if driver.StorageChanged(cr, modified) {
+	if len(driver.GetStorageName()) == 0 {
+		runCreate = true
```
bparees: This is wrong.

since you are looking at spec.Storage, this is always going to be true because spec.Storage.BucketName never gets set, only status.

coreydaley:
@bparees ptal, updated per our conversation

pkg/storage/s3/s3.go
```diff
@@ -211,35 +260,69 @@ func (d *driver) CreateStorage(cr *opapi.ImageRegistry, modified *bool) error {
...
+				default:
+					util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionUnknown, "Unknown Error Occurred", err.Error(), modified)
 					return err
 				}
 			}
 		}
-		break
 	}
+	if len(d.Config.Bucket) != 0 && bucketExists {
+		*cr.Status.Storage.S3 = *d.Config
```
bparees: assuming you're trying to avoid referencing the same thing, you need to do a deepcopy here.  Otherwise any pointer fields embedded in these objects will still point to the same place.

pkg/storage/s3/s3.go
```diff
@@ -211,35 +260,69 @@ func (d *driver) CreateStorage(cr *opapi.ImageRegistry, modified *bool) error {
...
+						continue
+					default:
+						util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionFalse, aerr.Code(), aerr.Error(), modified)
+						return err
+					}
+				}
+			}
+			cr.Status.StorageManaged = true
+			*cr.Status.Storage.S3 = *d.Config
+			*cr.Spec.Storage.S3 = *d.Config
```
bparees: same problem here. you're shallow copying.

pkg/storage/s3/s3.go
```diff
@@ -330,39 +413,40 @@ func (d *driver) CreateStorage(cr *opapi.ImageRegistry, modified *bool) error {
...
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok {
 			util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionUnknown, aerr.Code(), aerr.Error(), modified)
 		}
+		return err
+	}
+
+	cr.Spec.Storage.S3.Bucket = ""
+	d.Config.Bucket = ""
+	*cr.Status.Storage.S3 = *d.Config
```
bparees: and here.

pkg/storage/s3/s3.go
```diff
@@ -372,15 +456,17 @@ func (d *driver) CompleteConfiguration(cr *opapi.ImageRegistry, modified *bool)
...
 	}
-
-	cr.Status.Storage.State.S3 = d.Config
+	if cr.Spec.Storage.S3 == nil {
+		cr.Spec.Storage.S3 = &opapi.ImageRegistryConfigStorageS3{}
+	}
+	if cr.Status.Storage.S3 == nil {
+		cr.Status.Storage.S3 = &opapi.ImageRegistryConfigStorageS3{}
+	}
+	*cr.Spec.Storage.S3 = *d.Config
```
bparees: and here

pkg/storage/s3/s3.go
```diff
@@ -66,11 +70,18 @@ func NewDriver(crname string, crnamespace string, c *opapi.ImageRegistryConfigSt
 	}
 }
 
-// GetType returns the name of the storage driver
+// UpdateFromStorage updates the driver from an image registry resource
+func (d *driver) UpdateFromStorage(cfg opapi.ImageRegistryConfigStorage) {
+	d.Config = cfg.S3
```
bparees: and you should probably consider making this a deepcopy too, lest you accidentally assign it elsewhere.

pkg/storage/s3/s3.go
coreydaley: updated

pkg/storage/s3/s3.go
coreydaley: updated

pkg/storage/s3/s3.go
coreydaley: updated

pkg/storage/s3/s3.go
coreydaley: updated

pkg/storage/s3/s3.go
coreydaley: updated

pkg/resource/nodecadaemon.go
```diff
@@ -128,14 +128,13 @@ func (ds *generatorNodeCADaemonSet) Create() error {
...
-
 	modified := false
 	exists := false
-	for _, env := range daemonSet.Spec.Template.Spec.Containers[0].Env {
+	for i, env := range daemonSet.Spec.Template.Spec.Containers[0].Env {
 		if env.Name == "internalRegistryHostname" {
 			exists = true
 			if env.Value != ds.hostname {
-				env.Value = ds.hostname
+				daemonSet.Spec.Template.Spec.Containers[0].Env[i].Value = ds.hostname
```
bparees: thanks

bparees:
squash and lgtm.

coreydaley:
@bparees squashed

pkg/storage/s3/s3.go
```diff
@@ -330,39 +413,40 @@ func (d *driver) CreateStorage(cr *opapi.ImageRegistry, modified *bool) error {
...
 	_, err = svc.DeleteBucket(&s3.DeleteBucketInput{
-		Bucket: aws.String(cr.Status.Storage.State.S3.Bucket),
+		Bucket: aws.String(d.Config.Bucket),
 	})
 
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok {
 			util.UpdateCondition(cr, opapi.StorageExists, operatorapi.ConditionUnknown, aerr.Code(), aerr.Error(), modified)
 		}
+		return err
```
dmage: What if we get timeout while reading the response from the server?

pkg/testframework/framework.go
```diff
@@ -57,7 +57,7 @@ func DeleteCompletely(getObject func() (metav1.Object, error), deleteObject func
 		return err
 	}
 
-	return wait.Poll(1*time.Second, AsyncOperationTimeout, func() (stop bool, err error) {
+	return wait.Poll(1*time.Second, 5*time.Minute, func() (stop bool, err error) {
```
dmage: When do we need more than 2 minutes to complete our actions?

pkg/storage/s3/s3.go
coreydaley: Yes, i'll put that in the next pr.

test/e2e/unmanaged_test.go
```diff
@@ -41,14 +44,22 @@ func TestUnmanaged(t *testing.T) {
...
+	var cr *imageregistryapi.ImageRegistry
+	var err error
+	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		cr, err = client.ImageRegistries().Get(testframework.ImageRegistryName, metav1.GetOptions{})
+		if err != nil {
+			t.Fatal(err)
+		}
 
-	cr.Spec.ManagementState = operatorapi.Unmanaged
+		cr.Spec.ManagementState = operatorapi.Unmanaged
```
dmage: What can concurrently update our CR after the registry is completely deployed?

pkg/testframework/framework.go
coreydaley: I can set that back in the next pr also

test/e2e/unmanaged_test.go
coreydaley: this may have been due to the hot loop we had that was updating the cr and making api calls, might be able to set this back after this pr merges

pkg/storage/storage.go
```diff
@@ -31,8 +31,9 @@ type Driver interface {
 	StorageExists(*opapi.ImageRegistry, *bool) (bool, error)
 	RemoveStorage(*opapi.ImageRegistry, *bool) error
 	StorageChanged(*opapi.ImageRegistry, *bool) bool
-	GetStorageName(*opapi.ImageRegistry, *bool) (string, error)
+	GetStorageName() string
 	SyncSecrets(*coreapi.Secret) (map[string]string, error)
+	UpdateFromStorage(cfg opapi.ImageRegistryConfigStorage)
```
dmage: Oh, we have so many methods in this interface. We need to document them.

pkg/storage/s3/s3.go
```diff
@@ -330,39 +413,40 @@ func (d *driver) CreateStorage(cr *opapi.ImageRegistry, modified *bool) error {
...
 	}
 
-	svc, err := d.getSVC()
+	svc, err := d.getS3Service()
 	if err != nil {
 		return err
 	}
 	_, err = svc.DeleteBucket(&s3.DeleteBucketInput{
-		Bucket: aws.String(cr.Status.Storage.State.S3.Bucket),
+		Bucket: aws.String(d.Config.Bucket),
```
dmage: This field contains the value from Spec. Is it the bucket that we indented to delete?

coreydaley: When used from the finalizer, this field contains the bucket from Status

pkg/storage/storage.go
coreydaley: They are documented in s3.go, but that can be moved here in the future

coreydaley:
It's a belated authentic holiday miracle ...

PR #145, opened 2019-01-09
Remove unused arguments and fields

coreydaley:
These arguments are no longer needed and better replaced by the appropriate constants.

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -119,7 +119,7 @@ func (c *Controller) Bootstrap() error {
 		cr.Spec.HTTPSecret = fmt.Sprintf("%x", string(secretBytes[:]))
 	}
 
-	driver, err := storage.NewDriver(cr.Name, c.params.Deployment.Namespace, &cr.Spec.Storage)
+	driver, err := storage.NewDriver(&cr.Spec.Storage)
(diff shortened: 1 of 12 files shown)


pkg/operator/bootstrap.go
```diff
@@ -119,7 +119,7 @@ func (c *Controller) Bootstrap() error {
 		cr.Spec.HTTPSecret = fmt.Sprintf("%x", string(secretBytes[:]))
 	}
 
-	driver, err := storage.NewDriver(cr.Name, c.params.Deployment.Namespace, &cr.Spec.Storage)
+	driver, err := storage.NewDriver(&cr.Spec.Storage)
```
bparees: i can still see cases where a driver might need this information.  e.g. the filesystem driver which has to create PVCs in the registry namespace.

coreydaley: Can't we just use a constant in the api for that?  I believe we are doing it elsewhere in the code base.

coreydaley: Do you want me to close this pr then?

bparees: yes let's hold off on further driver interface refactoring until we have more implementations that really tell us what shape we need it to take on.

PR #146, opened 2019-01-09
Add support for syncing registry certs

mrunalp:
@bparees @runcom ptal.

Signed-off-by: Mrunal Patel <mrunalp@gmail.com>

### DIFF
--- a/pkg/resource/nodecadaemon.go
+++ b/pkg/resource/nodecadaemon.go
@@ -63,6 +63,7 @@ spec:
                 fi
                 echo $f
                 ca_file_path="/tmp/serviceca/${f}"
+                f=$(echo $f | sed  -r 's/(.*)\.\./\1:/')
                 reg_dir_path="/etc/docker/certs.d/${f}"
                 if [ -e "${reg_dir_path}" ]; then
                     cp -u $ca_file_path $reg_dir_path/ca.crt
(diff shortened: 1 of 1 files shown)


pkg/resource/nodecadaemon.go
```diff
@@ -49,6 +49,33 @@ spec:
...
+                  continue
+              fi
+              if [ "${d}" == "image-registry.openshift-image-registry.svc.cluster.local:5000" ]; then
+                  continue
+              fi
+              reg_conf_path="/tmp/serviceca/${d}"
+              if [ ! -e "${reg_conf_path}" ]; then
+                  rm -rf /etc/docker/certs.d/$d
+              fi
+          done
```
bparees: you can't do this just once, you need to keep looping and re-checking periodically.  The pod is not going to get restarted just because the configmap content is updated.  That's why the logic below is a continuous while loop with a sleep.

mrunalp: Yeah, I intended to put it in the loop of course :) Fixing.

bparees:
/lgtm

@mrunalp i'll let you follow this up w/ logic to handle the "..port" syntax we discussed.

mrunalp:
@bparees I added a commit here with changes for handling .. since tests were failing anyways. ptal.

PR #147, opened 2019-01-10
Cleanup Items

coreydaley:
- Replace testframework constants with api constants
 - Set DeleteCompletely timeout back to AsyncOperationTimeout
 - Standardize the imageregistry/v1 api alias to be imageregistryv1
 - All CVO status messages should start with a capital letter

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ test-unit:
 	go test ./cmd/... ./pkg/...
 
 test-e2e:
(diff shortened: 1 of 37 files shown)


dmage:
Nice, now we have the same constant for the operator namespace and for the registry namespace.

bparees:
> Nice, now we have the same constant for the operator namespace and for the registry namespace.

not sure if that's sarcasm or not, but if/when we want a separate namespace for the operator we can split the constant back out.

PR #153, opened 2019-01-11
drop tls api

bparees:
also added godoc to our api types and updated the readme.


pkg/apis/imageregistry/v1/types.go
```diff
@@ -10,14 +10,55 @@ import (
...
+	// and the registry itself
 	ImageRegistryOperatorNamespace = "openshift-image-registry"
+
+	// Status Conditions
+
+	// OperatorStatusTypeRemoved denotes that the image-registry instance has been
+	// removed
+	// TODO: does this serve any purpose?  As soon as it's removed we'll bootstrap
+	// a new one.
+	OperatorStatusTypeRemoved = "Removed"
```
coreydaley: I vote +1 to removing this status.

bparees:
> any thoughts on moving https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/clusterconfig/clusterconfig.go#L20-L25 into types.go ?

i'm against it.  in fact even some of the stuff(constants) currently in types.go should probably be removed.. it's for external/user facing APIs/values only.  clusterconfig.go is an internal struct and internal constants, right?

pkg/apis/imageregistry/v1/types.go
```diff
@@ -77,63 +172,33 @@ type ImageRegistryConfigStorage struct {
...
-	Requests        ImageRegistryConfigRequests    `json:"requests,omitempty"`
-	TLS             bool                           `json:"tls,omitempty"`
-	CAConfigName    string                         `json:"caConfigName,omitempty"`
-	DefaultRoute    bool                           `json:"defaultRoute,omitempty"`
-	Routes          []ImageRegistryConfigRoute     `json:"routes,omitempty"`
-	Replicas        int32                          `json:"replicas,omitempty"`
-	LogLevel        int64                          `json:"logging,omitempty"`
-}
+	// MaxRunning sets the maximum in flight api requests to the registry
+	MaxRunning int `json:"maxrunning,omitempty"`
```
coreydaley: Are we not doing CamelCase for these for the json? maxrunning vs maxRunning (same for the other two below)

pkg/migration/migration.go
```diff
@@ -302,7 +302,8 @@ storage:
 		return fail("unable to make configuration for the storage: %s", err)
 	}
 
-	tls, tlsSecret, err := migrateTLS(config, podFileGetter)
+	//tls, tlsSecret, err := migrateTLS(config, podFileGetter)
```
coreydaley: Did you mean to just comment this out and not delete it? (same with a few lines below)

pkg/apis/imageregistry/v1/types.go
bparees: they should have been.  updated.

pkg/apis/imageregistry/v1/types.go
legionus: @coreydaley This status is needed at least for tests as there is no other way to know whether all resources have been deleted or not.

pkg/migration/migration.go
legionus: @bparees Since you deleted the rest of the code and did not comment out, this comment does not make sense.

legionus:
It is so nice to come on Monday and see that PR merged at the weekend. Looks like review requests are completely pointless.

bparees:
> It is so nice to come on Monday and see that PR merged at the weekend. Looks like review requests are completely pointless.

@legionus I requested a review from you, corey, and oleg.  I did not intend that all 3 of you must review it before it merges, only that any one of you could do it.  But of course if you have concerns they can be addressed in a follow up as well.

PR #154, opened 2019-01-11
[WIP] add resources+nodeselector config fields

bparees:


pkg/resource/podtemplatespec.go
```diff
@@ -218,6 +218,16 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, params *parameters.
 
 	image := os.Getenv("IMAGE")
 
+	resources := corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceCPU:    resource.MustParse("100m"),
+			corev1.ResourceMemory: resource.MustParse("256Mi"),
```
bparees: @legionus @dmage i'm not sure where these defaults came from originally (i copied them from below as you can see), but they look a little low to me?  Any opinions on raising them?  Granted, an admin can now easily set them to something more appropriate as needed.

pkg/testframework/imageregistry.go
```diff
@@ -342,3 +342,24 @@ func MustEnsureClusterOperatorStatusIsSet(t *testing.T, client *Clientset) {
 		t.Fatal(err)
 	}
 }
+
+func MustEnsureOperatorIsNotHotLooping(t *testing.T, client *Clientset) {
```
bparees: @coreydaley fyi

bparees:
/assign @dmage @legionus @coreydaley 

I am going to put a hold on this so all 3 of you have a chance to weigh in before it merges.  If there are no changes required, the last person to review can remove the hold.

/hold

pkg/testframework/imageregistry.go
coreydaley: +1

pkg/testframework/imageregistry.go
```diff
@@ -342,3 +342,24 @@ func MustEnsureClusterOperatorStatusIsSet(t *testing.T, client *Clientset) {
...
+	// Allow the operator a few seconds to stabilize
+	time.Sleep(15 * time.Second)
```
dmage: This function is called after MustEnsureImageRegistryIsAvailable, which waits for Progressing == False && Available == True. Doesn't that condition mean that the operator has done its job and now it's in the stable state?

coreydaley: The registry can be in the available state and there still be a looping issue in say, syncStorage or syncSecrets.

bparees: yeah, the hotlooping condition we've seen doesn't cause the registry itself to redeploy, it just causes us to keep updating the CR repeatedly, which in turn triggers all the resource rendering logic.  The generated resources don't change (hence no re-deploy) but it hammers the api server.

bparees:
/hold cancel

merging based on https://coreos.slack.com/archives/CE2HALN2W/p1547650725472800

pkg/resource/podtemplatespec.go
dmage: It was a long time ago, so it's a bit outdated. And it's just a request for the scheduler, not the limit that the registry should be fit into. Maybe in 4.0 we'll be able to get better numbers from telemetry.

bparees: ok.  we'll leave it for now.  I think Clayton said the registry was actually only using like 50megs on api.ci right now.

PR #156, opened 2019-01-15
Update Get calls to use the lister cache

coreydaley:
Updates all(most) Get calls to use the lister cache.
ClusterVersion and ClusterOperator are excluded as they provided a steady stream of events which caused a hotloop like issue of constant events being processed.


pkg/clusterconfig/clusterconfig.go
```diff
@@ -111,22 +107,22 @@ func GetAWSConfig() (*Config, error) {
...
 	// Look for a user defined secret to get the AWS credentials from first
-	sec, err := client.Secrets(imageregistryv1.ImageRegistryOperatorNamespace).Get(imageregistryv1.ImageRegistryPrivateConfigurationUser, metav1.GetOptions{})
+	sec, err := listers.Secrets.Get(imageregistryv1.ImageRegistryPrivateConfigurationUser)
 	if err != nil && errors.IsNotFound(err) {
 		// If no user defined secret is found, use the system one
-		sec, err = client.Secrets(installerConfigNamespace).Get(installerAWSCredsName, metav1.GetOptions{})
+		sec, err = listers.InstallerSecrets.Get(installerAWSCredsName)
 		if err != nil {
-			return nil, fmt.Errorf("unable to get secret %q: %v", fmt.Sprintf("%s/%s", installerConfigNamespace, installerAWSCredsName), err)
+			return nil, fmt.Errorf("unable to get secret %q: %v", fmt.Sprintf("%s/%s", "kube-system", installerAWSCredsName), err)
```
bparees: why is this moving away from being a constant?

pkg/operator/controller.go
```diff
@@ -382,6 +400,11 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 		}
 	}
 
+	err = c.clusterStatus.Create(c.listers)
+	if err != nil {
+		glog.Errorf("unable to create cluster operator resource: %s", err)
+	}
```
bparees: i know you just moved this code here, but isn't it going to report an error on every subsequent startup because the resource already exists?  we should probably check if it's an already exists error and not log that.

pkg/testframework/mock/listers/configmaps.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	coreapi "k8s.io/api/core/v1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	coreset "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type MockConfigMapNamespaceLister struct {
```
bparees: doesn't look like this is used anywhere?

pkg/clusterconfig/clusterconfig.go
coreydaley: leftover from another change I had tested.  Fixed in next push

pkg/testframework/mock/listers/configmaps.go
coreydaley: Looks like that was a leftover from when I tried doing listers for the ClusterVersion and ClusterOperator.  Removed in next push.

coreydaley:
@bparees ptal
Updates made per your feedback in a separate commit.
https://github.com/openshift/cluster-image-registry-operator/pull/156/commits/95e5500017006b50efd48363bbb6bbd721bdc8fa

bparees:
lgtm, let's see what @legionus and @dmage have to say.

bparees:
> I'm not happy how pkg/resource has been refactored (I used individual listers to track dependencies). 

I assume @coreydaley's intent was to simplify the number of parameters/fields being passed around and just share a single collection of listers.  Can you talk more about the need to track dependencies?  When does it come up and perhaps there is another way to achieve that?  I'm not against splitting the listers back out, but at least at first blush it does seem a little cleaner to strip the arguments down to a single lister collection.

dmage:
It could be useful for writing unit tests (it's easier to mock a couple of listers, rather than all of them). But I don't have a strong opinion on this. Having less arguments makes sense to me.

coreydaley:
@bparees @dmage. Can I get an lgtm then?

bparees:
@coreydaley i'm invoking team lead privilege and holding this until 

https://github.com/openshift/cluster-image-registry-operator/pull/154

and

https://github.com/openshift/cluster-image-registry-operator/pull/159

land as they are critical for declaring feature complete on the registry operator and i'm pretty sure this PR is going to cause a conflict w/ at least one those.  Sorry.

But otherwise i think it's g2g.

/hold

coreydaley:
> It could be useful for writing unit tests (it's easier to mock a couple of listers, rather than all of them). But I don't have a strong opinion on this. Having less arguments makes sense to me.

You only have to mock the ones that actually get used, the rest can stay nil.

coreydaley:
@bparees I would be happy if ANYTHING would merge at this point ;)

coreydaley:
@bparees ptal, i rebased on top of your other two pull requests

bparees:
our flakes are bad, but they aren't this bad..please show at least one aws job passing and then i'll review.

bparees:
@coreydaley have you built this operator image locally and run it on a cluster?

coreydaley:
@bparees yes, but that was awhile ago, I'll check it again today

coreydaley:
@bparees I think this is ready for another look.  
There were enough changes from Oleg's pull request that got merged that I was not able to rebase.  I just started with a fresh branch and copied things in from the github diff.  Just an FYI.

pkg/client/listers.go
```diff
@@ -1,27 +1,30 @@
...
 	Routes              routelisters.RouteNamespaceLister
-	ClusterRoles        rbaclisters.ClusterRoleLister
-	ClusterRoleBindings rbaclisters.ClusterRoleBindingLister
-	OpenShiftConfig     corelisters.ConfigMapNamespaceLister
+	ClusterRoles        krbaclisters.ClusterRoleLister
+	ClusterRoleBindings krbaclisters.ClusterRoleBindingLister
+	OpenShiftConfig     kcorelisters.ConfigMapNamespaceLister
 	ImageConfigs        configlisters.ImageLister
-	ImageRegistry       regoplisters.ConfigLister
+	Config              regoplisters.ConfigLister
```
bparees: `Configs`?  (to be consistent w/ the other ones)

I also kinda lean towards RegistryConfigs for clarity, but i can understand this name since it matches the resource type name.

pkg/operator/finalizer.go
```diff
@@ -66,7 +65,7 @@ func (c *Controller) finalizeResources(o *imageregistryv1.Config) error {
 	cr := o
 	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
 		if cr == nil {
-			cr, err := client.Configs().Get(o.Name, metav1.GetOptions{})
+			cr, err := c.listers.Config.Get(o.Name)
```
bparees: when we're try to fetch the latest object because of a conflict, we probably want to go to the souce, not the cache.

pkg/storage/util/util.go
```diff
@@ -83,7 +79,7 @@ func CreateOrUpdateSecret(name string, namespace string, data map[string]string)
 	var updatedSecret *coreapi.Secret
 
 	if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		cur, err := client.Secrets(namespace).Get(name, metaapi.GetOptions{})
+		cur, err := listers.Secrets.Get(name)
```
bparees: same here about going to the source vs the cache since we're trying to ensure we have the latest content

coreydaley: The retry is technically for the create/update that happens towards the end of the function, not the Get, as I don't believe that would return a conflict type error.

pkg/operator/finalizer.go
coreydaley: The retry is technically for the create/update that happens towards the end of the function, not the Get, as I don't believe that would return a conflict type error.

bparees: yes but the point of this get is to ensure we have the latest version of the object before we start making changes and invoke update on it, right?

if the update fails due to a conflict we'll loop back up to here, fetch a "newer" version of the object and try again... except if now if the cache isn't updated, the "newer" version won't be any newer and we'll fail on the update again, until the cache is updated.

coreydaley:
@bparees updated per your comments

pkg/storage/util/util.go
coreydaley: Updated per our conversation

pkg/operator/finalizer.go
coreydaley: updated per our conversation

pkg/client/listers.go
coreydaley: updated

pkg/apis/imageregistry/v1/types.go
```diff
@@ -76,7 +76,7 @@ type ConfigList struct {
 // +genclient:nonNamespaced
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// Config is the configuration object for a registry instance managed by
+// RegistryConfigs is the configuration object for a registry instance managed by
```
bparees: this shouldn't have been changed

pkg/apis/imageregistry/v1/zz_generated.deepcopy.go
```diff
@@ -19,7 +19,7 @@ func (in *Config) DeepCopyInto(out *Config) {
 	return
 }
 
-// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new Config.
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new RegistryConfigs.
```
bparees: nor this

pkg/apis/imageregistry/v1/types.go
coreydaley: fixed

pkg/apis/imageregistry/v1/zz_generated.deepcopy.go
coreydaley: fixed

pkg/client/client.go
```diff
@@ -22,10 +22,10 @@ const (
 	OperatorNameEnvVar = "OPERATOR_NAME"
 )
 
-// GetConfig creates a *rest.Config for talking to a Kubernetes apiserver.
+// GetConfig creates a *rest.RegistryConfigs for talking to a Kubernetes apiserver.
 // Otherwise will assume running in cluster and use the cluster provided kubeconfig.
 //
-// Config precedence
+// RegistryConfigs precedence
```
bparees: should be Config?

pkg/resource/utils.go
```diff
@@ -17,7 +17,7 @@ func asOwner(cr *v1.Config) metav1.OwnerReference {
 	isController := true
 	return metav1.OwnerReference{
 		APIVersion:         v1.SchemeGroupVersion.String(),
-		Kind:               "Config",
+		Kind:               "RegistryConfigs",
```
bparees: Config

test/e2e/basic_empty_dir_test.go
```diff
@@ -21,7 +21,7 @@ func TestBasicEmptyDir(t *testing.T) {
 	cr := &imageregistryv1.Config{
 		TypeMeta: metav1.TypeMeta{
 			APIVersion: imageregistryv1.SchemeGroupVersion.String(),
-			Kind:       "Config",
+			Kind:       "RegistryConfigs",
```
bparees: Config here and a few other places.

pkg/client/client.go
coreydaley: fixed

pkg/resource/utils.go
coreydaley: fixed

test/e2e/basic_empty_dir_test.go
coreydaley: fixed

PR #158, opened 2019-01-15
don't create default route unless it is requested

bparees:


bparees:
/assign @dmage @legionus 

@legionus @dmage ptal, i think this was a bug... we were always creating the default route, regardless of spec.DefaultRoute's value.

PR #159, opened 2019-01-16
set external registry hostnames on cluster image config

bparees:
Set the external registry hostname in the cluster configuration (this drives the imagestream public docker repository value) when routes are configured for the registry.

### DIFF
--- a/pkg/apis/imageregistry/v1/types.go
+++ b/pkg/apis/imageregistry/v1/types.go
@@ -11,6 +11,9 @@ import (
 
 const (
 
+	// DefaultRouteName is the name of the default route created for the registry
+	// when a default route is requested from the operator
+	DefaultRouteName = "default-route"
 	// ImageRegistryName is the name of the image-registry workload resource (deployment)
 	ImageRegistryName = "image-registry"
 
--- a/pkg/operator/complete.go
+++ b/pkg/operator/complete.go
@@ -24,7 +24,7 @@ func verifyResource(cr *imageregistryv1.Config, p *parameters.Globals) error {
 	}
 
 	names := map[string]struct{}{
-		imageregistryv1.ImageRegistryName + "-default-route": {},
+		imageregistryv1.DefaultRouteName: {},
 	}
 
 	for _, routeSpec := range cr.Spec.Routes {
--- a/pkg/resource/imageconfig.go
+++ b/pkg/resource/imageconfig.go
@@ -1,12 +1,18 @@
 package resource
 
 import (
+	"reflect"
+	"sort"
+	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	configapi "github.com/openshift/api/config/v1"
 	configset "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	configlisters "github.com/openshift/client-go/config/listers/config/v1"
+	routelisters "github.com/openshift/client-go/route/listers/route/v1"
 
 	imageregistryv1 "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1"
 	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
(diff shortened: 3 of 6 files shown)


bparees:
@coreydaley @dmage @legionus ptal

note that this is built on top of https://github.com/openshift/cluster-image-registry-operator/pull/154, so only the second commit is really relevant.

Doing this also left me with a strong desire to refactor pkg/testframework into test/framework or something, but i'm going to hold off for now.

/assign @coreydaley @dmage @legionus 

i will settle for 2/3 reviews :)

/hold

test/e2e/configuration_test.go
```diff
@@ -70,5 +75,113 @@ func TestConfiguration(t *testing.T) {
...
+		}
+		if cfg == nil {
+			return false, nil
+		}
+		externalHosts = cfg.Status.ExternalRegistryHostnames
+
+		foundDefaultRoute := false
+		foundUserRoute := false
+		for _, h := range externalHosts {
+			if strings.HasPrefix(h, "default-route-openshift-image-registry") {
```
coreydaley: Can we build default-route-openshift-image-registry so that this test doesn't have to be updated if we change that name?

test/e2e/configuration_test.go
```diff
@@ -70,5 +75,113 @@ func TestConfiguration(t *testing.T) {
...
+		}
+		if routes == nil || len(routes.Items) < 2 {
+			t.Logf("insuffient routes found: %#v", routes)
+			return false, nil
+		}
+
+		foundDefaultRoute := false
+		foundUserRoute := false
+		for _, r := range routes.Items {
+			if strings.HasPrefix(r.Spec.Host, "default-route-openshift-image-registry") {
```
coreydaley: Same as above

pkg/operator/complete.go
```diff
@@ -24,7 +24,7 @@ func verifyResource(cr *imageregistryv1.Config, p *parameters.Globals) error {
 	}
 
 	names := map[string]struct{}{
-		imageregistryv1.ImageRegistryName + "-default-route": {},
+		"default-route": {},
```
coreydaley: default-route is used several times in this pull request, should it be a const somewhere instead of using a string every time? (would also help with building the default-route-openshift-image-registry in the tests)

pkg/resource/imageconfig.go
```diff
@@ -62,26 +71,78 @@ func (gic *generatorImageConfig) Create() error {
...
+			hostname := ingress.Host
+			if len(hostname) == 0 {
+				continue
+			}
+			if strings.HasPrefix(hostname, "default-route-"+gic.namespace) {
+				// prepend the default hostname because it has special meaning to the cluster
+				// so we want it first in the list.
+				externalHostnames = append([]string{hostname}, externalHostnames...)
+			} else {
+				externalHostnames = append(externalHostnames, hostname)
```
dmage: Will they be always in the same order?

bparees: I assumed so (that a list retrieval of objects always returns those objects in the same order) but maybe @deads2k can confirm.   @deads2k ?

deads2k: > I assumed so (that a list retrieval of objects always returns those objects in the same order) but maybe @deads2k can confirm. @deads2k ?

Yes, we store slices as ordered items and keep their serialization consistent to avoid extra writes.

bparees: @deads2k not talking about the string slice here though...  asking about the list of Route objects we're getting back from our List API call to the route api.

pkg/operator/complete.go
bparees: updated

test/e2e/configuration_test.go
bparees: updated

test/e2e/configuration_test.go
bparees: updated

bparees:
/hold

while we get resolution on whether "List" returns a stable list of objects.

pkg/apis/imageregistry/v1/types.go
```diff
@@ -11,6 +11,9 @@ import (
 
 const (
 
+	// DefaultRouteName is the name of the default route created for the registry
+	// when a default route is requested from the operator
+	DefaultRouteName = "default-route"
```
coreydaley: Isn't this just a prefix, not the whole route name?

bparees: it is the route name.  it is not the route hostname.

bparees: ("oc get route default-route")

pkg/resource/imageconfig.go
bparees: i spoke with david, List results are not guaranteed to return in the same order, so i'll need to add a sort here.

pkg/resource/imageconfig.go
```diff
@@ -62,26 +72,89 @@ func (gic *generatorImageConfig) Create() error {
...
+		if routeOwner == nil || routeOwner.UID != gic.owner.UID {
+			continue
+		}
+		for _, ingress := range route.Status.Ingress {
+			hostname := ingress.Host
+			if len(hostname) == 0 {
+				continue
+			}
+			if strings.HasPrefix(hostname, imageregistryv1.DefaultRouteName+"-"+gic.namespace) {
+				// prepend the default hostname because it has special meaning to the cluster
```
coreydaley: This comment doesn't make sense anymore

coreydaley:
one nit, but otherwise lgtm

pkg/resource/imageconfig.go
bparees: whoops. fixed and pushed and re-tagged.

pkg/testframework/imageregistry.go
```diff
@@ -342,3 +342,24 @@ func MustEnsureClusterOperatorStatusIsSet(t *testing.T, client *Clientset) {
 		t.Fatal(err)
 	}
 }
+
+func MustEnsureOperatorIsNotHotLooping(t *testing.T, client *Clientset) {
+	// Allow the operator a few seconds to stabilize
+	time.Sleep(15 * time.Second)
```
coreydaley: I wonder if 15 seconds isn't enough (based on the ci infrastructure flaking so much) https://github.com/openshift/cluster-image-registry-operator/pull/159/files#diff-a8d53698d228697ddce3c9426f371e5b
Do you think we could have it check every 5 seconds for a minute with a wait.Poll ?

bparees: what purpose would wait.poll serve?  wait until we think it's stable, then wait another 15s to confirm it is stable?

pkg/resource/imageconfig.go
```diff
@@ -62,26 +72,88 @@ func (gic *generatorImageConfig) Create() error {
...
 	return true, err
 }
 
 func (gic *generatorImageConfig) Delete(opts *metav1.DeleteOptions) error {
-	return gic.client.Images().Delete(gic.GetName(), opts)
+	return gic.configClient.Images().Delete(gic.GetName(), opts)
+}
+
+func (gic *generatorImageConfig) getRouteHostnames() ([]string, error) {
+	var externalHostnames []string
```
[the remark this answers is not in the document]
coreydaley: +1

bparees:
passed all tests this morning.  flakes look irrelevant, merging.

PR #162, opened 2019-01-17
Do not encrypt the bucket if we did not create it

### LINKED ISSUE #160: encryption being applied to user provided buckets
based on this:

https://github.com/openshift/cluster-image-registry-operator/blob/7f6ffaf58e512d193e8de4d683783921f59f3797/pkg/storage/s3/s3.go#L365-L377

it looks like we apply our own encryption settings to the s3 bucket even if the bucket was supplied by the user, even if the bucket already existed.  This means a user can't provide their own s3 bucket w/ their own encryption settings.

coreydaley:
- Do not tag the bucket if we did not create it
- Do not encrypt the bucket if we did not create it
- Do not enable cleanup of incomplete multipart uploads if we did not create the bucket

Fixes #160

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -338,79 +338,85 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config, modified *bool) error
 
 	// Tag the bucket with the openshiftClusterID
 	// along with any user defined tags from the cluster configuration
-	if ic.Platform.AWS != nil {
-		var tagSet []*s3.Tag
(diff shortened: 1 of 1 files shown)


coreydaley:
@bparees @dmage @legionus ptal

coreydaley:
Sheesh, one extra level of indentation sure made a mess of the diff on github ...
All I added was the if cr.Status.StatusManaged{} around the three sections of code ...

dmage:
@coreydaley `Diff settings -> Hide whitespace changes` is a nice option to review such changes.

/lgtm

PR #163, opened 2019-01-17
Set operator cpu request

adambkaplan:


adambkaplan:
/assign @bparees 
/cc @dmage 
fyi @legionus @derekwaynecarr

adambkaplan:
looks like the operand CPU request is already set for new installs: https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/podtemplatespec.go#L249

bparees:
> looks like the operand CPU request is already set for new installs: https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/podtemplatespec.go#L249

for the operand, yes.

PR #164, opened 2019-01-17
Remove storage medium if ManagementState set to Removed

coreydaley:
Remove the storage medium (if we manage it):
 - When the finalizer runs
 - When the ManagementState is set to Removed

Also moved objectInfo to pkg/util/util.go (ObjectInfo) so that it can easily be used in more places.

### DIFF
--- a/pkg/resource/generator.go
+++ b/pkg/resource/generator.go
@@ -3,6 +3,8 @@ package resource
 import (
 	"fmt"
 	"github.com/golang/glog"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"time"
 
 	routeset "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 
@@ -23,7 +25,7 @@ import (
 
 	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
 	"github.com/openshift/cluster-image-registry-operator/pkg/storage"
-	"github.com/openshift/cluster-image-registry-operator/pkg/storage/util"
+	storageutil "github.com/openshift/cluster-image-registry-operator/pkg/storage/util"
 )
 
 func NewGenerator(kubeconfig *rest.Config, listers *client.Listers, params *parameters.Globals) *Generator {
@@ -156,7 +158,7 @@ func (g *Generator) syncSecrets(cr *imageregistryv1.Config, modified *bool) erro
 		glog.Infof("Updating secret %q with updated credentials.", fmt.Sprintf("%s/%s", g.params.Deployment.Namespace, imageregistryv1.ImageRegistryPrivateConfiguration))
 
 		// Update the image-registry-private-configuration secret
-		_, err := util.CreateOrUpdateSecret(imageregistryv1.ImageRegistryPrivateConfiguration, g.params.Deployment.Namespace, data)
+		_, err := storageutil.CreateOrUpdateSecret(imageregistryv1.ImageRegistryPrivateConfiguration, g.params.Deployment.Namespace, data)
 		if err != nil {
 			return err
 		}
@@ -283,6 +285,27 @@ func (g *Generator) Remove(cr *imageregistryv1.Config, modified *bool) error {
 		}
 		glog.Infof("object %s deleted", Name(gen))
 		*modified = true
+	}
+
+	driver, err := storage.NewDriver(cr.ObjectMeta.Name, cr.ObjectMeta.Namespace, &cr.Status.Storage)
+	if err != nil {
+		return err
+	}
+
+	var derr error
+	var retriable bool
+	err = wait.PollImmediate(1*time.Second, 5*time.Minute, func() (stop bool, err error) {
+		if retriable, derr = driver.RemoveStorage(cr, modified); derr != nil {
+			if retriable {
+				return false, nil
+			} else {
+				return true, derr
+			}
+		}
+		return true, nil
+	})
+	if err != nil {
+		return fmt.Errorf("unable to remove storage: %s, %s", err, derr)
 	}
 
 	return nil
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -16,7 +16,6 @@ package e2e
 
 import (
 	"fmt"
-	"k8s.io/client-go/util/retry"
 	"reflect"
 	"testing"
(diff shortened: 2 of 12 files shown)


coreydaley:
@dmage @legionus @bparees ptal

pkg/resource/generator.go
```diff
@@ -284,6 +286,41 @@ func (g *Generator) Remove(cr *imageregistryv1.Config, modified *bool) error {
...
+			o, err = rclient.Configs().Get(cr.Name, metav1.GetOptions{})
+			if err != nil {
+				return fmt.Errorf("failed to get %s, retrying: %v", util.ObjectInfo(cr), err)
+			}
+			driver.UpdateFromStorage(o.Status.Storage)
+		}
+		modified := false
+
+		if err := driver.RemoveStorage(cr, &modified); err != nil {
+			o = nil
```
bparees: i'm not seeing the point of this assignment.

pkg/resource/generator.go
```diff
@@ -284,6 +286,41 @@ func (g *Generator) Remove(cr *imageregistryv1.Config, modified *bool) error {
...
+	rclient, err := regopset.NewForConfig(g.kubeconfig)
+	if err != nil {
+		return err
+	}
+	driver, err := storage.NewDriver(cr.ObjectMeta.Name, cr.ObjectMeta.Namespace, &cr.Status.Storage)
+	if err != nil {
+		return err
+	}
+
+	o := cr
```
bparees: not seeing the point of this either... looks like you can just keep using the cr variable

bparees: ok, nm, i do see why you need the copy.  still don't think you need the nil assignment below though.

coreydaley: updated

pkg/resource/generator.go
coreydaley: updated

coreydaley:
@bparees ptal, updated per your comments

pkg/resource/generator.go
```diff
@@ -285,5 +287,22 @@ func (g *Generator) Remove(cr *imageregistryv1.Config, modified *bool) error {
...
+	driver, err := storage.NewDriver(cr.ObjectMeta.Name, cr.ObjectMeta.Namespace, &cr.Status.Storage)
+	if err != nil {
+		return err
+	}
+
+	var derr error
+	err = wait.PollImmediate(1*time.Second, 5*time.Minute, func() (stop bool, err error) {
+
+		if derr = driver.RemoveStorage(cr, modified); derr != nil {
+			return false, nil
```
bparees: please do a follow up to make this more fine-grained.  We do not want to wait 5 minutes to remove storage in the face of fatal errors.

test/e2e/aws_test.go
```diff
@@ -339,11 +338,11 @@ func TestAWSUpdateCredentials(t *testing.T) {
 	testframework.MustEnsureClusterOperatorStatusIsSet(t, client)
 
 	// Create the image-registry-private-configuration-user secret using the invalid credentials
-	err = retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+	err = wait.PollImmediate(1*time.Second, testframework.AsyncOperationTimeout, func() (stop bool, err error) {
 		if _, err := util.CreateOrUpdateSecret(imageregistryv1.ImageRegistryPrivateConfigurationUser, imageregistryv1.ImageRegistryOperatorNamespace, fakeAWSCredsData); err != nil {
-			return fmt.Errorf("unable to create secret %q: %#v", fmt.Sprintf("%s/%s", imageregistryv1.ImageRegistryOperatorNamespace, imageregistryv1.ImageRegistryPrivateConfigurationUser), err)
+			return false, nil
```
bparees: you should log this error so we can see why it was failing when it eventually times out.

pkg/resource/generator.go
coreydaley: fine-grained as in check the type of error that we get? or fine-grained as in a lower timeout?

bparees: check the type of error.  So we don't spend 5 minutes retrying this operation if we (should) know it's never going to succeed.

coreydaley: updated

test/e2e/aws_test.go
coreydaley: updated

coreydaley:
@bparees ptal, updated per your suggestions.

pkg/resource/generator.go
coreydaley: If an aws error returns, it's not retriable, else it is retriable (updated in the RemoveStorage function)

pkg/resource/generator.go
```diff
@@ -285,5 +287,25 @@ func (g *Generator) Remove(cr *imageregistryv1.Config, modified *bool) error {
...
+		return err
+	}
+
+	var derr error
+	err = wait.PollImmediate(1*time.Second, 5*time.Minute, func() (stop bool, err error) {
+		if retriable, derr := driver.RemoveStorage(cr, modified); derr != nil {
+			if retriable {
+				return false, nil
+			} else {
+				return true, err
```
dmage: derr?

bparees: yeah this logic is a little borked now.  returning derr is probably the simplest fix.  As it stands now you're risking returning a nil error when you actually got a fatal error.

coreydaley: yep, that was a typo, updated in current push

pkg/resource/generator.go
```diff
@@ -285,5 +287,25 @@ func (g *Generator) Remove(cr *imageregistryv1.Config, modified *bool) error {
...
 	}
 
+	driver, err := storage.NewDriver(cr.ObjectMeta.Name, cr.ObjectMeta.Namespace, &cr.Status.Storage)
+	if err != nil {
+		return err
+	}
+
+	var derr error
+	err = wait.PollImmediate(1*time.Second, 5*time.Minute, func() (stop bool, err error) {
+		if retriable, derr := driver.RemoveStorage(cr, modified); derr != nil {
```
bparees: this isn't going to preserve derr the way you want, you've created a shadow copy of it.
https://play.golang.org/p/hPuFAc7DVYl

PR #165, opened 2019-01-18
[WIP] add monitoring rule for prometheus metrics

bparees:


bparees:
/hold

@dmage @legionus I need to do one final round of local testing on this to ensure i got the RBAC config correct (I refactored it a bit after I got it working). and it needs release process approval to merge, but ptal now so we have all the pieces lined up.  Thanks.

bparees:
/hold cancel

this has been approved and i've finished my local testing

deploy/08-monitoring.yaml
```diff
@@ -0,0 +1,79 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: image-registry
+  namespace: openshift-monitoring
```
brancz: this should be your own namespace

brancz:
In order for your ServiceMonitor to be picked up by the cluster monitoring stack, you also need to label your namespace appropriately as described here: https://docs.google.com/document/d/1rCKAYTrYMESjJDyJ0KvNap05NNukmNXVVi6MShISnOw/edit#heading=h.37avim0ervi

bparees:
/hold

i get the distinct impression this PR is the cause of the job failures though i'm not sure how..my only theory is the CVO is having issues applying the service monitoring resources and that prevents the installer from completing.

Issue #161, opened 2019-01-16
Support user KMS key


pierreozoux:
As a cluster operator in an enterprise env, I'm required to encrypt all the things.

It would be nice if I could supply my own kms key.

bparees:
@pierreozoux can you open this as a formal RFE so we can track it properly?  we are using github issues just for very lightweight items.

pierreozoux:
Should I open a RedHat support case? Or what is this RFE, is it on the coreos Jira? Or RedHat bugzilla? (I'm a bit confused, and that's why I thought GitHub would be the best place)

bparees:
Going through a support case is best(and support can then open the RFE), that way the RFE is tied to a specific customer request which helps w/ prioritization.

That said, our support team may be slightly confused when you ask for an RFE against an unreleased product :)  You can point them to this issue discussion.

I will also say we may choose not to implement this...  basically i'd expect your choices to be:

1) use the bucket we create, w/ whatever encryption we choose
or
2) supply your own bucket w/ whatever encryption you choose

(you'll be able to do either of those in the 4.0GA.. in fact you can do them in the beta today as well)

we are trying to move away from the hybrid model of us managing components that admins/end-users are customizing significantly.  We are trying to simplify the configuration surface area.

pierreozoux:
> we are trying to move away from the hybrid model of us managing components that admins/end-users are customizing significantly. We are trying to simplify the configuration surface area.

Sounds good :) I'll close this issue then!

bparees:
actually @coreydaley understood this RFE better than I did and I now understand why we still need support on our side for KMS keys.  Corey is implementing the support here:
https://github.com/openshift/cluster-image-registry-operator/pull/166

PR #166, opened 2019-01-18
Initial plumbing to support Encrypt, and KeyID (KMS Encryption)

coreydaley:
Initial plumbing to support the following options on S3:

 - Encrypt - Enable or disable KMS encryption on S3
 - KeyID - The KMS Key ID for the S3 bucket

fixes https://github.com/openshift/cluster-image-registry-operator/issues/161


bparees:
/hold

i think we need to have some discussions about exactly how deep we want to get into the s3 bucket management business.

pkg/storage/s3/s3.go
```diff
@@ -88,7 +89,19 @@ func (d *driver) ConfigEnv() (envs []corev1.EnvVar, err error) {
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_BUCKET", Value: d.Config.Bucket},
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_REGION", Value: d.Config.Region},
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_REGIONENDPOINT", Value: d.Config.RegionEndpoint},
-		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_ENCRYPT", Value: fmt.Sprintf("%v", d.Config.Encrypt)},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_SECURE", Value: strconv.FormatBool(d.Config.Secure)},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_ENCRYPT", Value: strconv.FormatBool(d.Config.Encrypt)},
+		corev1.EnvVar{
```
coreydaley: I'm not 100% sure (according to the documentation) whether or not the KMS KeyID is sensitive data or not, so maybe it doesn't need to be stored in a secret ...
https://docs.docker.com/registry/storage-drivers/s3/

coreydaley: I confirmed that the KMS Key ID is not sensitive information, so refactoring to just use the field

coreydaley:
@bparees Do I need to find a way to automate testing this KMS setup?

pkg/apis/imageregistry/v1/types.go
```diff
@@ -150,10 +150,24 @@ type ImageRegistryConfigProxy struct {
...
+	// Region is the AWS region in which your bucket exists
+	// Optional, will be set based on the installed AWS Region
+	Region string `json:"region,omitempty"`
+	// RegionEndpoint is the endpoint for S3 compatible storage services
+	// Optional, defaults based on the Region that is provided
 	RegionEndpoint string `json:"regionEndpoint,omitempty"`
-	Encrypt        bool   `json:"encrypt,omitempty"`
+	// Secure indicates whether to use HTTPS instead of HTTP
+	// Optional, defaults to true
+	Secure bool `json:"secure,omitempty"`
```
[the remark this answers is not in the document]
coreydaley: Updated to remove the ability to configure "Secure".  The S3 driver defaults to https so we should be good

pkg/apis/imageregistry/v1/types.go
```diff
@@ -150,10 +150,21 @@ type ImageRegistryConfigProxy struct {
...
+	// RegionEndpoint is the endpoint for S3 compatible storage services
+	// Optional, defaults based on the Region that is provided
 	RegionEndpoint string `json:"regionEndpoint,omitempty"`
-	Encrypt        bool   `json:"encrypt,omitempty"`
+	// Encrypt specifies whether the registry stores the image in encrypted format or not
+	// Optional, defaults to false
+	Encrypt bool `json:"encrypt,omitempty"`
+	// KeyID is the KMS key ID to use for encryption
+	// Optional, Encrypt must be true, or this parameter is ignored
+	KeyID string `json:"keyID,omitempty"`
```
bparees: what is the behavior if someone changes Encrypt or KeyID later?

coreydaley: Ultimately, we would want the syncStorage to pick up the change and re-deploy the image-registry with the new env vars

bparees: so today the behavior is.....?  i just want to make sure someone won't end up in a bad state if they set this after the fact (where we might end up half-configuring the system, ie setting up the registry config but not configuring the bucket properly).

What's going to happen when we see that spec storage config doesn't match status storage config currently?

bparees: @coreydaley bump ^^

coreydaley: I'm working on this, it may be part of a larger change to deal with what should happen when things change like this.

coreydaley:
@bparees @dmage @legionus This is ready for another look (if you want to even though CI is down)

pkg/apis/imageregistry/v1/types.go
bparees: "if i add a keyID to the custom resource, it changes the encryption on the bucket,and if i remove it, it changes it back "

i take it that's the answer to my question on this?

coreydaley: Yes, that is the answer now :) I'm not sure what it was before ...

pkg/storage/s3/s3.go
```diff
@@ -76,6 +76,7 @@ func (d *driver) ConfigEnv() (envs []corev1.EnvVar, err error) {
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_REGION", Value: d.Config.Region},
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_REGIONENDPOINT", Value: d.Config.RegionEndpoint},
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_ENCRYPT", Value: fmt.Sprintf("%v", d.Config.Encrypt)},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_S3_KEYID", Value: fmt.Sprintf("%v", d.Config.KeyID)},
```
bparees: why not %s ?

bparees: also i'd skip appending this if d.Config.KeyID is empty

bparees: in fact you probably don't need the fmt.Sprintf at all since it is a string?

pkg/storage/s3/s3.go
```diff
@@ -322,7 +337,10 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config, modified *bool) error
 				util.UpdateCondition(cr, imageregistryv1.StorageEncrypted, operatorapi.ConditionFalse, "Unknown Error Occurred", err.Error(), modified)
 			}
 		} else {
-			util.UpdateCondition(cr, imageregistryv1.StorageEncrypted, operatorapi.ConditionTrue, "Encryption Successful", "Default encryption was successfully enabled on the S3 bucket", modified)
+			util.UpdateCondition(cr, imageregistryv1.StorageEncrypted, operatorapi.ConditionTrue, "Encryption Successful", fmt.Sprintf("Default %s encryption was successfully enabled on the S3 bucket", encryptionType), modified)
+			d.Config.Encrypt = true
+			cr.Status.Storage.S3 = d.Config.DeepCopy()
+			cr.Spec.Storage.S3 = d.Config.DeepCopy()
```
bparees: per our chat, this needs to be done even in cases when we don't manage the bucket, otherwise we'll think the storage has changed everytime we re-sync, so we'll keep churning through this logic.  (not quite a hotloop,but unnecessary).

pkg/storage/s3/s3.go
coreydaley: Updated in next push

pkg/storage/s3/s3.go

coreydaley:
@bparees ptal

bparees:
AWS limit issues

/retest

bparees:
@coreydaley squash and lgtm
/approve

coreydaley:
@bparees squashed.

coreydaley:
had to rename testframework to framework in aws_test.go, weird

coreydaley:
Forgot that these only exist if set now, fixed in current push:
```
aws_test.go:307: unable to find environment variable: wanted REGISTRY_STORAGE_S3_REGIONENDPOINT
	aws_test.go:307: unable to find environment variable: wanted REGISTRY_STORAGE_S3_KEYID
```

coreydaley:
@bparees thanks, i was going to tag you after the tests passed

PR #171, opened 2019-01-24
make e2e tests more flake tolerant

bparees:


pkg/testframework/imageregistry.go
```diff
@@ -346,17 +346,33 @@ func MustEnsureClusterOperatorStatusIsSet(t *testing.T, client *Clientset) {
...
 	// Allow the operator a few seconds to stabilize
 	time.Sleep(15 * time.Second)
-	cfg, err := client.Configs().Get(imageregistryv1.ImageRegistryResourceName, metav1.GetOptions{})
-	if err != nil || cfg == nil {
+	var cfg imageregistryv1.Config
+	var err error
+	err = wait.Poll(1*time.Second, 30*time.Second, func() (stop bool, err error) {
+		cfg, err = client.Configs().Get(imageregistryv1.ImageRegistryResourceName, metav1.GetOptions{})
+		if err != nil || cfg == nil {
+			logger.Logf("failed to retrieve registry operator config: %v", err)
```
coreydaley: Isn't this the registry configuration? Not the operator configuration? (same below also)

bparees:
@dmage @legionus @coreydaley ptal.  this will hopefully make our e2es a little more resilient to flakes caused by the api server being flaky.

pkg/testframework/imageregistry.go
bparees: it's the registry config that is consumed by the registry operator and converted into configuration inputs to the registry.

it is not, for example, the config.yaml file which is consumed directly by the registry, which i would consider "registry configuration".

coreydaley:
/retest
imageregistry.go:158: unable to stop the operator: timed out waiting for the condition

PR #174, opened 2019-01-29
Add PVC storage driver

legionus:


coreydaley:
This gets repeated several times, can it be refactored into a non-exported function?

```
namespace, err := regopclient.GetWatchNamespace()
	if err != nil {
		return false, fmt.Errorf("failed to get watch namespace: %s", err)
	}

	kubeconfig, err := regopclient.GetConfig()
	if err != nil {
		return false, err
	}

	client, err := coreset.NewForConfig(kubeconfig)
	if err != nil {
		return false, err
	}
```

bparees:
> This gets repeated several times, can it be refactored into a non-exported function?

alternatively perhaps the client should just be created as part of the NewDriver() implementation.

legionus:
@bparees @dmage please review again

pkg/storage/pvc/pvc.go
```diff
@@ -0,0 +1,161 @@
...
+	if !cr.Status.StorageManaged {
+		return false, nil
+	}
+
+	err := d.Client.PersistentVolumeClaims(d.Namespace).Delete(d.ClaimName, &metav1.DeleteOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return false, err
+	}
+
+	return false, nil
```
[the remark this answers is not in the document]
legionus: @dmage [retriable](https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/generator.go#L256)

test/e2e/pvc_test.go
```diff
@@ -0,0 +1,200 @@
...
+	deploy, err := client.Deployments(imageregistryv1.ImageRegistryOperatorNamespace).Get(imageregistryv1.ImageRegistryName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if deploy.Status.AvailableReplicas == 0 {
+		framework.DumpObject(t, "deployment", deploy)
+		t.Errorf("error: the deployment doesn't have available replicas")
+	}
+
+	logs, err := framework.GetOperatorLogs(client)
```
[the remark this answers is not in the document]
legionus: @dmage why ?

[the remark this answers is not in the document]
legionus: @dmage removed.

pkg/storage/pvc/pvc.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, err
+	}
+
+	client, err := coreset.NewForConfig(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+
+	return &driver{
+		ClaimName: fmt.Sprintf("%s-storage", imageregistryv1.ImageRegistryName),
```
bparees: how does a user customize the PVC so they can control the PV they get?

pkg/apis/imageregistry/v1/types.go
```diff
@@ -169,12 +170,17 @@ type ImageRegistryConfigStorageFilesystem struct {
 	VolumeSource corev1.VolumeSource `json:"volumeSource,omitempty"`
 }
 
+type ImageRegistryConfigStoragePVC struct {
+	Size resource.Quantity `json:"size,omitempty"`
```
bparees: what's the behavior if/when a user updates this value?

legionus: @bparees You can not change it. Otherwise, we will need to re-create PVC and you will lose the current data since another PV is likely to be selected.

pkg/storage/pvc/pvc.go
legionus: @bparees Before that, you told me not to generate this name. If we do not generate a name, then it is under our control in our namespace. User can customize only size.

legionus:
> at a minimum i think they need to be able to point the storage configuration to a PVC they have created, similar to how the s3 driver allows them to point to an S3 bucket they created.

What should I do if the user does not specify a PVC name and the `operator-defined-pvc-name` is already exist and it does not meet the size requirement?

pkg/apis/imageregistry/v1/types.go
bparees: yes, this is the same issue we discussed extensively around how to deal w/ s3 bucket changes.  Ultimately (After a few false starts) we decided to let people do it because the alternatives are not great (you install a new cluster, we gave you a 100gig PVC by default. You know that's not going to be enough, so you want to change it...what's the process for doing so?  Basically I have to remove the registry, disable the operator, delete the CR, create a new CR w/ the right storage value, and then re-enable the operator... or something along those lines, right?)


In this case, perhaps a better experience would be to not allow them to specify the size at all, just require them to provide the name of the PVC they want to use.  If they provide no PVC name, we create a PVC w/ a size we choose (as you are doing).  If they change the PVC name in the future, just like s3, they lose their storage, that was their choice.

That is similar to the current S3 experience.

legionus:
I no longer understand what behavior you want from this driver.

pkg/apis/imageregistry/v1/types.go
dmage: Why do we need to ask the PVC name? As you told before, its our own namespace and there is no need to have more than one PVC in our namespace.

bparees:
> What should I do if the user does not specify a PVC name and the `operator-defined-pvc-name` is already exist and it does not meet the size requirement?

why would it exist?  because the user created it themselves?  I think you can safely report an error condition in that case.  ("could not create default PVC, it already exists and is not owned by the operator").  We should be able to distinguish (via ownerref or annotation) whether the operator created the PVC or not.  If the operated created it, then it should just use it...

> I no longer understand what behavior you want from this driver.

I want the following:

1) by default a cluster running a platform where we use PVs for storage, should create a PVC and bring up a registry that mounts the PV claimed by the PVC.  The PVC should define sane defaults in terms of RW-many, size, etc.

2) a user should be able to change the configuration of the PVC in case they want to use a different PV, different attributes, different size, etc.  This can be done by either:
  a) exposing parts of the PVC api in the registry storage api and relying on the operator to update the PVC/create a new PVC with those attributes
  b) allowing the user to specify the name of the PVC they want to use and relying on them to create an appropriate PVC
  c) ???

of those, (b) seems easier than deciding which PVC fields we want to expose and managing the PVC.

I apologize if i've given mixed/conflicting suggestions about how to implement this(ie around naming) but those are the fundamental requirements as I see them.  I certainly want to have more conversations if you think those requirements are inappropriate/insufficient/inaccurate.

legionus:
> I certainly want to have more conversations if you think those requirements are inappropriate/insufficient/inaccurate.

No, I just implement what you ask for as I have been doing for a quite long time in case of the operator.

legionus:
@dmage ptal

dmage:
¯\_(ツ)_/¯
I don't get the idea why we need to check ownerrefs. And I barely understand why we need to parametrise the claim name (to copy data from an old PV to a new PV, I guess?).

legionus:
> to copy data from an old PV to a new PV, I guess?

Yes. I guess you will need to mount old and new PVC in some pod and copy data from one to another. Only after that you can use the new claim name. But I'm also not sure why.

legionus:
@bparees ptal

pkg/storage/pvc/pvc.go
```diff
@@ -0,0 +1,223 @@
...
+
+	for _, claimMode := range claim.Spec.AccessModes {
+		for _, mode := range requiredModes {
+			if claimMode == mode {
+				return nil
+			}
+		}
+	}
+
+	return fmt.Errorf("PVC %s does not contain the necessary access modes", d.Config.Claim)
```
bparees: print the required modes?

test/e2e/pvc_test.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+	framework.MustRemoveImageRegistry(t, client)
+}
+
+func createPV(t *testing.T, storageClass string) error {
+	client := framework.MustNewClientset(t, nil)
+	name := ""
+
+	for i := 0; i < 100; i++ {
```
bparees: why do we need to create 100?

test/e2e/pvc_test.go
```diff
@@ -0,0 +1,255 @@
...
+		pvName := fmt.Sprintf("pv-%s", rand.String(64))
```
bparees: including `i` in the name would further ensure uniqueness

test/e2e/pvc_test.go
```diff
@@ -0,0 +1,255 @@
...
+		localPath := fmt.Sprintf("/tmp/%s", pvName)
+
+		if err := os.MkdirAll(localPath, 0777); err != nil {
```
bparees: this code(the test code) isn't running on the cluster nodes, so how does this work?  I think you're creating PVs that aren't actually backed by real directories on the cluster nodes (which i guess is fine since as you noted, these tests don't push/pull to the registry anyway right now).

test/e2e/pvc_test.go
```diff
@@ -0,0 +1,255 @@
...
+func createPVWithStorageClass(t *testing.T) error {
+	client := framework.MustNewClientset(t, nil)
+
+	storageClassList, err := client.StorageClasses().List(metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("unable to list storage classes: %s", err)
+	}
+
+	for _, storageClass := range storageClassList.Items {
+		if err := createPV(t, storageClass.Name); err != nil {
```
bparees: since a storageclass is required to have a provisioner, why do we need to do this?

shouldn't we only need to create PVs manually if there is no storageclass/provisioner defined on the cluster?

test/e2e/pvc_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+	defer testDefer(t, client)
+
+	if err := createPV(t, ""); err != nil {
+		t.Fatal(err)
+	}
+
+	if err := createPVWithStorageClass(t); err != nil {
+		t.Fatal(err)
+	}
```
bparees: why do we need to do both createPV and createPVWithStorageClass ?

test/e2e/pvc_test.go
legionus: @bparees You do not quite understand. This is the number of retries.

bparees: i see now, my bad.

test/e2e/pvc_test.go
legionus: @bparees I use 64 random bits with 100 retries. If you think that adding one number will greatly increase randomization, ok I will add it.

bparees: it was only relevant when i thought you were creating 100 at once.  since you're only trying to create one, it's less important (and yes, it wasn't that important in the first place, but given that you had the entropy available, it seemed to make sense to use it).

test/e2e/pvc_test.go
legionus: @bparees Tests pass. It works. PV change status.phase into the `available` state. OK, will remove `mkdir`.

test/e2e/pvc_test.go
legionus: > since a storageclass is required to have a provisioner, why do we need to do this?

The provisioner did not create PV. I have to create them manually.

> shouldn't we only need to create PVs manually if there is no storageclass/provisioner defined on the cluster?

@bparees I tried different configurations and only this works.

test/e2e/pvc_test.go
legionus: @bparees I need it to run locally without provisioner.

test/e2e/pvc_test.go
bparees: I assume this is a callback to our discussion from yesterday, so i'll just say:

If you would like to add a basic image push/pull to our operator suite to ensure the various storage setups work, that would be great.  If not, when we get "e2e-xxxx" where xxxx defaults to PV storage, we will find out whether this works end to end, but assuming the PVC is being created and a PV is being claimed, it I guess I would expect that it is, barring some sort of permissions issue writing to the PV or incorrect configuration setting in the registry pod.

test/e2e/pvc_test.go
coreydaley: Why not use uuid like we do in s3? (you can truncate it to the specified length)

test/e2e/pvc_test.go
legionus: In addition, if at some point someone turns provisioner off, then I want the test to work without storageclass.

test/e2e/pvc_test.go
bparees: > The provisioner did not create PV. I have to create them manually.

did you talk to the storage team about it to understand why it didn't do so?

> @bparees I tried different configurations and only this works.

so just creating a PV and then a PVC to claim it, does not work?  That seems very surprising, i'm pretty sure we do exactly that elsewhere.  I'd like you to discuss this with the storage team to understand why you had to do it this way, unless you understand it yourself.

test/e2e/pvc_test.go
legionus: @coreydaley I think the kubernetes module is more useful here. It was created specifically for such cases and I'm sure everything is done right there.

test/e2e/pvc_test.go
bparees: > @bparees I need it to run locally without provisioner.

why not just run it that way always then?

(if the autoprovisioner were working, then i think it would be reasonable to prefer to run the test using autoprovisioned PVs where possible, especially since those PVs will actually work, unlike the hostvolume ones you are creating which aren't actually backed by anything, but since you're creating the no-storage-class PVs manually either way, I don't see what purpose creating the ones with the storageclass serves?)

I think this comes back to: you should talk to the storage team to learn why PVs aren't being autoprovisioned on AWS.

test/e2e/pvc_test.go
legionus: >  If not, when we get "e2e-xxxx" where xxxx defaults to PV storage, we will find out whether this works end to end, but assuming the PVC is being created and a PV is being claimed, it I guess I would expect that it is, barring some sort of permissions issue writing to the PV or incorrect configuration setting in the registry pod.

@bparees I dumped all PVs / PVCs  and did not see any other than those created by me. The deployment cannot change status to `processing=false` if the PVC failed to bound PV.

test/e2e/pvc_test.go
bparees: You don't have to learn everything, just enough to answer my questions and convince yourself+me that we are doing this the right/expected way.  

Right now it sounds like you found a way that works(and that's probably sufficient for what we specifically need right now), but you don't know why it works/has to be done that way/is correct.  I'd like to get to the bottom of that, not just for our own purposes, but because my one of my understandings(perhaps incorrect) at a product level is that we should have PV autoprovisioning that works out of the box on an AWS cluster.  I think it's a useful exercise.

test/e2e/pvc_test.go
bparees: > @bparees Because when I create a PVC in a cluster, the admission plugin automatically adds a storageclass to it and my PV without a storageclass does not bound. I can't control it. That is why I create all kinds of PV (with and without storageclass).

thanks, that helps (I do recall you mentioning part of this issue on scrum, now that you bring it up again here).

But i take it you've seen some clusters (your own local ones) that do not have the admission plugin behavior setting a storageclass on the PVC.

Putting a couple comments in the code explaining why you're creating 2 PVs/what you've observed would be helpful for the next person who's looking at this.

test/e2e/pvc_test.go
legionus: > but you don't know why it works/has to be done that way/is correct.

I need to know how our test cluster was configured and I will answer to you. I am almost 100% sure that we do not have a autoprovisioner. But for now I can not prove it.

test/e2e/pvc_test.go
legionus: > But i take it you've seen some clusters (your own local ones) that do not have the admission plugin behavior setting a storageclass on the PVC.

@bparees For example `oc cluster up` (v3.11).

bparees: wow.  that is brave/dedicated to keep making oc cluster up work.

legionus: > I'm pretty sure we don't have a autoprovisioner. Only admission plugin.

I was wrong. We have an autoprovisioner there, but for AWS. AWSElasticBlockStore do not support ReadWriteMany access mode [1].

[1] https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

legionus:
> aws_s3_bucket.ignition: Error creating S3 bucket: TooManyBuckets: You have attempted to create more buckets than allowed"
/retest

legionus:
@bparees ptal
/hold cancel

bparees:
@legionus bump me on this when you're done investigating the test flakes and are ready for a review/merge.  thanks

legionus:
@bparees I just made rebase once again. All other fails are just flakes.

PR #177, opened 2019-01-29
bump(*) update to kube 1.12 

bparees:


bparees:
@coreydaley the latest commit removes a couple places where we were setting modified=true that i don't think were correct.

As i understand it, we should only set modified=true if we modifed the CR itself (and thus need to call update).  Generating/updating other resources (like routes, the deployment, etc) do not require updating the CR itself.

coreydaley:
@bparees That is also how i understand it, as a modification to the CR should trigger refreshing / rebuilding other resources that need that information

bparees:
/hold

@coreydaley i'm suspicious that the kube 1.12 update is causing some of the weird client behavior i'm seeing in these failures, so i've split the fixes out to a separate PR:
https://github.com/openshift/cluster-image-registry-operator/pull/179

bparees:
example of weird failure:
```
logs.go:77: E0130 01:21:26.807626       1 reflector.go:134] github.com/openshift/client-go/route/informers/externalversions/factory.go:101: Failed to list *v1.Route: the server could not find the requested resource (get routes.route.openshift.io)
```

in: https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-image-registry-operator/177/pull-ci-openshift-cluster-image-registry-operator-master-e2e-aws-operator/803/

bparees:
these flakes are happening in 1.11 based operator PRs also, so
/hold cancel

PR #184, opened 2019-01-31
remove unnecessary status.internalregistryhostname field

bparees:
After further consideration i think setting the internal registry hostname in the registry resource status was unnecessary, and not doing it simplifies the code and makes the generator more like the other generators.

As part of this PR I also fixed up the node-ca daemon to actually use the internalregistryhostname value, which it previously was not doing, and switched it to fetch the value from the image config resource instead of the registry resource.


bparees:
@mrunalp fyi since this modifies the node-ca logic a bit.

/assign @dmage 

you'll probably want to review w/ whitespace ignored.

bparees:
and apparently we're not running a make verify job on our PRs :(

PR #185, opened 2019-02-01
[WIP] Avoid repeated message in wait loops

dmage:
Just an idea for a small improvement of our e2e tests logs.

Another option is to have something like this:
```
func TestFoo(tt *testing.T) {
    t := testframework.NewT(tt)
    defer t.Close()
```
Having a wrapper will allow us to have a timestamp for each t.Logf.

/cc @legionus @bparees @coreydaley 

What do you think?

### DIFF
--- a/pkg/testframework/framework.go
+++ b/pkg/testframework/framework.go
@@ -1,6 +1,7 @@
 package testframework
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dmage:
Example of output:
```
    imageregistry.go:207: 16:05:17 waiting for the registry: available (<unset>), progressing (<unset>), failing (<unset>), removed (<unset>)
    imageregistry.go:207: 16:05:21 waiting for the registry: available (<unset>), progressing (<unset>), failing (<unset>), removed (<unset>) (x4)
    imageregistry.go:207: 16:05:22 waiting for the registry: available (False), progressing (True), failing (False), removed (False)
    imageregistry.go:207: 16:05:28 waiting for the registry: available (False), progressing (True), failing (False), removed (False) (x6)
    imageregistry.go:207: 16:05:29 waiting for the registry: available (True), progressing (True), failing (False), removed (False)
    imageregistry.go:207: 16:05:31 waiting for the registry: available (True), progressing (True), failing (False), removed (False) (x2)
    imageregistry.go:207: 16:05:32 waiting for the registry: available (True), progressing (False), failing (False), removed (False)
```

bparees:
Seems like a reasonable idea to me.  I was also thinking (while going through our logs) that we should reduce the wait.Poll interval from 1s to 5 or 10s.  That would reduce the spam volume significantly.

legionus:
Looks good for me

PR #186, opened 2019-02-01
create a clusteroperator empty object so the CVO waits for the registry operator  ------------------------ >8 ------------------------ # Do not modify or remove the line above. # Everything below it will be ignored.

bparees:
Requesting a pull to openshift:master from bparees:operator

Write a message for this pull request. The first block
of text is the title and the rest is the description.


bparees:
/hold

i'm concerned about this potentially causing install failures (in cases where the registry operator does not successfully get to available=true), so i want to sit on it for a bit.

PR #187, opened 2019-02-03
Refactoring bootstrap.go to only look for one custom resource

coreydaley:
The bootstrap process only needs to look for a single custom resource, not a list.

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -6,110 +6,56 @@ import (
 
 	"github.com/golang/glog"
 
-	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


pkg/operator/bootstrap.go
```diff
@@ -33,30 +32,35 @@ func resourceName(namespace string) string {
...
-	default:
-		return fmt.Errorf("only one registry custom resource expected in %s namespace, got %d", c.params.Deployment.Namespace, len(crList))
 	}
 
+	// If no registry resource exists,
+	// let's create one
 	var spec imageregistryv1.ImageRegistrySpec
 
-	dc, err := c.listers.DeploymentConfigs.Get(resourceName(c.params.Deployment.Namespace))
+	dc, err := c.listers.DeploymentConfigs.Get(deploymentConfigName(c.params.Deployment.Namespace))
```
bparees: the adoption/migration code is increasingly out of sync and unnecessary, this is probably a good opportunity to remove it.

coreydaley: updated in current push

coreydaley:
@bparees ptal

PR #188, opened 2019-02-03
Reorganizing some of the testframework

coreydaley:


dmage:
To be honest, I don't see the point of these changes.

bparees:
> To be honest, I don't see the point of these changes.

@dmage to me the point is to separate purely test-related code, from product code.  There's a different bar for what test helpers/utilities/etc look like, vs shipping code.  It also helps someone working in the code know what is relevant to the actual operator function and what is not.

bparees:
(I don't know that some of the refactoring out into separate files for things like checking hostname+routes was strictly necessary or even a pattern we want to continue to grow as it's a bit heavy, but i'm definitely on board w/ the bulk of the change to move the test framework helpers into the test package).

/lgtm

dmage:
> separate files for things like checking hostname+routes was strictly necessary or even a pattern we want to continue to grow as it's a bit heavy

For me the rule of thumb is if we use the function in two tests or more, it should be extracted in a separate package. Otherwise it's better to keep it in the _test.go file.

bparees:
> For me the rule of thumb is if we use the function in two tests or more, it should be extracted in a separate package. Otherwise it's better to keep it in the _test.go file.

That seems like a reasonable rule to me.  Of course sometimes people end up re-inventing things because they don't realize it already exists in some other test package, so there's something to be said for putting it in a util package immediately so it's more discoverable.  [shrug].

not something i'm going to get hung up about one way or another.

PR #192, opened 2019-02-06
Remove modified

dmage:


bparees:
thanks @dmage i think this is a great cleanup/simplification

/lgtm

PR #199, opened 2019-02-08
add credential request for aws

bparees:
replaces https://github.com/openshift/cluster-image-registry-operator/pull/190


bparees:
@dgoodwin you can remove this from the cred minter operator now.

dgoodwin:
On it, thanks

PR #201, opened 2019-02-14
Renaming image registry resource

coreydaley:
Renaming image registry resource from instance to cluster


coreydaley:
@bparees ptal

bparees:
https://github.com/openshift/cluster-image-registry-operator/blob/master/manifests/registry-config.yaml#L5 needs to be updated too.

bparees:
(used by people manually creating the config resource)

coreydaley:
@bparees update pushed

PR #202, opened 2019-02-14
openstack: Add temporary OpenStack suport on FS

flaper87:
This commit adds OpenStack as a supported platform using an empty dir.
Note that this is not production ready and it's being done as a way to
allow OpenStack based deployments to succeed and move on.

Eventually this should be switched to swift.

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -107,6 +107,15 @@ func getPlatformStorage() (imageregistryv1.ImageRegistryConfigStorage, error) {
 		}
 	case installConfig.Platform.AWS != nil:
 		cfg.S3 = &imageregistryv1.ImageRegistryConfigStorageS3{}
+	case installConfig.Platform.OpenStack != nil:
+		// TODO(flaper87): This should be switch to swift as soon as support for
+		// it is complete. Using Emptydir for now so that OpenStack deployments
+		// (and work) can move forward for now. Not production ready!
+		cfg.Filesystem = &imageregistryv1.ImageRegistryConfigStorageFilesystem{
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		}
 	default:
 		// if we can't determine what platform we're on, fallback to creating
 		// a PVC for the registry.


eparis:
Isn't this the wrong fix? Didn't we discuss the default for the operator needed to create a PVC, but not block upgrades/CVO/install if the PVC wasn't satisfied?  This is going to create completely unusable cluster and we'll unlikely remember to come back and remove this block....

Ben, how hard is it to make the operator handle the PVC mechanism with (greatly) delayed PVC satisfaction.

bparees:
> Isn't this the wrong fix? 

Not at the moment, no.

> Didn't we discuss the default for the operator needed to create a PVC, but not block upgrades/CVO/install if the PVC wasn't satisfied? 

it's been discussed but there's been no agreement on the right mechanism for implementing that yet in terms of what conditions the registry should be reporting or what the CVO will respect.  

> This is going to create completely unusable cluster and we'll unlikely remember to come back and remove this block....

It's very likely we will remember.  When we go to support openstack as a platform, we will be implementing swift as the default storage that is configured.  this isn't going to get forgotten when we start supporting openstack as a platform.


> Ben, how hard is it to make the operator handle the PVC mechanism with (greatly) delayed PVC satisfaction.

Handle it how?  We can create the PVC and then sit on our thumbs until the PV is created (in fact that's what we're in the process of doing) at which point the registry would report available, but that doesn't help the situation w/ the CVO expecting the registry to report available=true when in fact it is not available because there's no PV yet.

Openstack is an unsupported platform, you get emptydir storage just like libvirt, until it becomes a supported platform at which point we can revisit the proper default storage configuration(swift) and availability reporting that ensures it doesn't block the install.  (This is a more urgent problem for baremetal where we will be creating PVCs, they will go unfulfilled, and the CVO will block, until we decide the right rules for when/how to lie to the CVO about our availability)

And in case it's not super clear:  this is the meets min fix that unblocks the openstack folks while not exerting resources on something that is not a 4.0(or even 4.1) problem.

tomassedovic:
For what it's worth, the OpenStack folks (@flaper87 and myself) have put Swift on list of things to do to graduate from being an experimental provider for the installer -- and we're happy to develop it in time or lend a hand.

This is just to fix the current OpenStack breakage, but absolutely aware of it not being the final solution.

PR #204, opened 2019-02-15
confirm internal registry hostname propagation to api server

bparees:
per discussion on aos-devel, block the registry operator from reporting available
until the internal registry hostname has been picked up by the openshift apiserver

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ test-unit: verify
 	./hack/test-go.sh ./cmd/... ./pkg/...
 
 test-e2e:
-	GOCACHE=off ./hack/test-go.sh -timeout 20m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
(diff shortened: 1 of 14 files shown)


bparees:
```
level=fatal msg="failed to initialize the cluster: Cluster operator image-registry has not yet reported success"
```

well this definitely managed to block startup :/

bparees:
ok this looks to be passing at least locally.

/assign @legionus @dmage

bparees:
/hold
need to look into the hotloop test failure.  I don't think this should have introduced any hotlooping that i can see.

smarterclayton:
Good sign that this has passed 3 times in a row.

bparees:
> Good sign that this has passed 3 times in a row.

unfortunately it appears to be wreaking havoc on our operator e2e test because it takes longer to report available when we tear it down and deploy it.  May have to bump some timeouts in our tests.

bparees:
/hold

bumped timeout to 40mins just so we can get logs out of the tests to see where we're spending so much time.

coreydaley:
@bparees bump

PR #209, opened 2019-02-18
Enable subresource status

### LINKED ISSUE #200: use a status subresource on the registry operator resource
I thought we'd already done this..is there a reason the registry resource doesn't define and use a Status subresource?

dmage:
Closes #200

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -153,21 +153,41 @@ func (c *Controller) sync() error {
 
 	metadataChanged := strategy.Metadata(&prevCR.ObjectMeta, &cr.ObjectMeta)
 	specChanged := !reflect.DeepEqual(prevCR.Spec, cr.Spec)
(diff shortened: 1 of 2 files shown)


pkg/operator/controller.go
```diff
@@ -172,6 +170,23 @@ func (c *Controller) sync() error {
...
+		glog.Infof("object changed: %s (status=%t)", util.ObjectInfo(cr), statusChanged)
+
+		client, err := regopset.NewForConfig(c.kubeconfig)
+		if err != nil {
+			return err
+		}
+
+		_, err = client.ImageregistryV1().Configs().UpdateStatus(cr)
+		if err != nil {
+			if !errors.IsConflict(err) {
```
bparees: I'm concerned that since may have just updated spec above, this is very likely to get a conflict.  Can we do something to avoid that scenario?

PR #210, opened 2019-02-18
Add ReadOnly flag

dmage:


bparees:
@dmage i think you missed some files on this commit

test/e2e/readonly_test.go
```diff
@@ -0,0 +1,65 @@
...
+		Spec: imageregistryv1.ImageRegistrySpec{
+			ManagementState: operatorapi.Managed,
+			Storage: imageregistryv1.ImageRegistryConfigStorage{
+				Filesystem: &imageregistryv1.ImageRegistryConfigStorageFilesystem{
+					VolumeSource: corev1.VolumeSource{
+						EmptyDir: &corev1.EmptyDirVolumeSource{},
+					},
+				},
+			},
+			ReadOnly: 1,
```
bparees: why not a boolean?

bparees:
/retest

there is definitely some flakes in our operator test that need to be hunted down, seems like we either aren't stabilizing into a running state quickly, or there's some issue w/ teardown, or there's a race between teardown+create

PR #211, opened 2019-02-22
Add Swift support

Fedosin:


coreydaley:
@bparees are we going to be able to write any tests for this using our existing infrastructure?

adambkaplan:
@coreydaley it looks like the swift library supports a fake embedded server, which may be suitable for unit/integration tests. I assume e2e testing will have to wait until CI supports OpenStack.

bparees:
/hold

at a minimum this can't merge until we have a 4.1 branch, since we can't risk destabilizing 4.0 at this point.

Fedosin:
@coreydaley hi! I'm going to write tests and finalize the code for this change next week. Currently it has "work in progress" status and is not ready for merging - so, no rush with 4.0, I think

pkg/storage/swift/swift.go
```diff
@@ -1,17 +1,76 @@
...
+		}
+	} else {
+		// Try to read credentials from the native secret
+		err := setSwiftCredentials(d.Listers, cr.Spec.Storage.Swift)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	opts := gophercloud.AuthOptions{
```
flaper87: I'm not a huge fan of building `AuthOpts` like this because clouds have different credentials. Some of them use regions, others don't. Others use trusts, etc. If you have the `clouds` object, you can tell `gophrecloud` to build the opts object itself.

Take a look here:

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/clients/machineservice.go#L91

https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/master/pkg/cloud/openstack/clients/machineservice.go#L146

pkg/storage/swift/utils.go
```diff
@@ -0,0 +1,94 @@
...
+	if v, ok := sec.Data["REGISTRY_STORAGE_SWIFT_PASSWORD"]; ok {
+		cr.Password = string(v)
+	} else {
+		return fmt.Errorf("secret %q does not contain required key \"REGISTRY_STORAGE_SWIFT_PASSWORD\"", fmt.Sprintf("%s/%s", imageregistryv1.ImageRegistryOperatorNamespace, imageregistryv1.ImageRegistryPrivateConfigurationUser))
+	}
+
+	return nil
+}
+
+func parseClouds(data []byte) (map[string]clientconfig.Cloud, error) {
```
flaper87: you can do away with this function if you use the internal `gophercloud` utilities.

pkg/apis/imageregistry/v1/types.go
```diff
@@ -204,10 +204,16 @@ type ImageRegistryConfigStorageGCS struct {
 type ImageRegistryConfigStorageSwift struct {
 	AuthURL   string `json:"authURL,omitempty"`
 	Container string `json:"container,omitempty"`
+	Username  string `json:"username,omitempty`
```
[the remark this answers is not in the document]
Fedosin: Not neccessary, but more convenient... If we read data from clouds.yaml, then username and password are stored there along with all other parameters. In that case we are able to set them to the config in one place. But we can separate them, if you think it should be better.

pkg/storage/swift/swift.go
```diff
@@ -1,17 +1,76 @@
...
+)
+
+// getSwiftClient returns a client that allows us to interact with the OpenStack Swift service
+func (d *driver) getSwiftClient(cr *imageregistryv1.Config) (*gophercloud.ServiceClient, error) {
+	if swiftClient != nil {
+		return swiftClient, nil
+	}
+
+	// First read config from the secret with clouds.yaml
+	if cr.Spec.Storage.Swift.CloudsSecret == nil || cr.Spec.Storage.Swift.CloudName == "" {
```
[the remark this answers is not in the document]
Fedosin: yes, thanks :)

pkg/storage/swift/utils.go
Fedosin: I couldn't find a function there for this usecase. all of them try to find clouds.yaml file in the filesystem first and read data from there: findAndReadCloudsYAML()  https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/utils.go#L99
But we need to parse data from the secret, and LoadCloudsYAML can't do that https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/requests.go#L62

pkg/storage/swift/swift.go
Fedosin: Okay, makes sense.

pkg/apis/imageregistry/v1/types.go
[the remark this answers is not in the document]
flaper87: I think we should be fine with just the secret reference. No need to separate the username/password from it. The configs are stored in a secret anyways

pkg/storage/swift/utils.go
flaper87: See how I did it in the 2 links I posted in my previous comment. That should have an example of how you can do this.

adambkaplan:
/hold

We are now in feature freeze for the 4.1 release.

pkg/storage/swift/utils.go
```diff
@@ -0,0 +1,71 @@
...
+
+	yaml "gopkg.in/yaml.v2"
+)
+
+// getSwiftCredentialsFromSecret reads credentials from the operator's native secret
+func getSwiftCredentialsFromSecret(listers *regopclient.Listers) (credentials, error) {
+	creds := credentials{}
+
+	// Look for a user defined secret to get the Swift credentials
+	sec, err := listers.Secrets.Get(imageregistryv1.ImageRegistryPrivateConfigurationUser)
```
legionus: You need to save data from the `ImageRegistryPrivateConfigurationUser` to `ImageRegistryPrivateConfiguration` as well.

pkg/storage/swift/swift.go
```diff
@@ -28,6 +137,10 @@ func (d *driver) ConfigEnv() (envs []corev1.EnvVar, err error) {
 		corev1.EnvVar{Name: "REGISTRY_STORAGE", Value: "swift"},
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_AUTHURL", Value: d.Config.AuthURL},
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_CONTAINER", Value: d.Config.Container},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_DOMAIN", Value: d.Config.Domain},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_DOMAINID", Value: d.Config.DomainID},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_TENANT", Value: d.Config.Tenant},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_TENANTID", Value: d.Config.TenantID},
 		corev1.EnvVar{
 			Name: "REGISTRY_STORAGE_SWIFT_USERNAME",
```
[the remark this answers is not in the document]
Fedosin: Yeah, they are different. Fixed.

pkg/storage/swift/swift.go
```diff
@@ -51,34 +164,55 @@ func (d *driver) ConfigEnv() (envs []corev1.EnvVar, err error) {
...
 }
 
 func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
+	client, err := d.getSwiftClient(cr)
+	if err != nil {
+		util.UpdateCondition(cr, imageregistryv1.StorageExists, operatorapi.ConditionFalse, err.Error(), err.Error())
+		return err
+	}
+
+	_, err = containers.Create(client, cr.Spec.Storage.Swift.Container, containers.CreateOpts{}).Extract()
```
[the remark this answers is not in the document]
Fedosin: Done.

pkg/storage/swift/swift.go
```diff
@@ -51,34 +164,55 @@ func (d *driver) ConfigEnv() (envs []corev1.EnvVar, err error) {
...
-}
 
-func (d *driver) StorageExists(cr *imageregistryv1.Config) (bool, error) {
-	return false, nil
+	util.UpdateCondition(cr, imageregistryv1.StorageExists, operatorapi.ConditionTrue, "Swift container Exists", "")
+	return true, nil
 }
 
 func (d *driver) StorageChanged(cr *imageregistryv1.Config) bool {
+	if !reflect.DeepEqual(cr.Status.Storage.Filesystem, cr.Spec.Storage.Filesystem) {
```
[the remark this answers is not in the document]
Fedosin: Thanks :)

pkg/storage/swift/utils.go
Fedosin: I don't quite get it... Do you want me to copy data from ImageRegistryPrivateConfigurationUser secret to ImageRegistryPrivateConfiguration secret?

Fedosin: Okay, done!

pkg/apis/imageregistry/v1/types.go
```diff
@@ -208,6 +208,10 @@ type ImageRegistryConfigStorageSwift struct {
 	DomainID  string `json:"domainID,omitempty"`
 	Tenant    string `json:"tenant,omitempty"`
 	TenantID  string `json:"tenantID,omitempty"`
+	// The name of the secret containing the openstack clouds.yaml file
+	CloudsSecret *corev1.SecretReference `json:"cloudsSecret"`
+	// The name of the cloud to use from the clouds.yaml file
+	CloudName string `json:"cloudName"`
```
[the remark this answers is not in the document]
Fedosin: @legionus yes, now this value is hardcoded to "openstack-cloud"

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,294 @@
...
+		if err != nil {
+			// TODO(Fedosin): an error should also be returned in this case
+			return nil
+		}
+		kubeClient, err := kubeset.NewForConfig(kubeconfig)
+		if err != nil {
+			// TODO(Fedosin): ditto
+			return nil
+		}
+		kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClient, 10*time.Minute, kubeinformers.WithNamespace(c.CloudsSecret.Namespace))
```
dmage: In general you also need to start your factory. But I'm against of creating a new factory here. We create a new driver instance on every event from the master API, so we expect this operation to be cheap. Having new informers aren't cheap (+1 watcher, +1 store).

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,294 @@
...
+			return nil, err
+		}
+
+		envs = append(envs,
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_AUTHURL", Value: cloud.AuthInfo.AuthURL},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_DOMAIN", Value: cloud.AuthInfo.DomainName},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_DOMAINID", Value: cloud.AuthInfo.DomainID},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_TENANT", Value: cloud.AuthInfo.ProjectName},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_TENANTID", Value: cloud.AuthInfo.ProjectID},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_USERNAME", Value: cloud.AuthInfo.Username},
```
dmage: You shouldn't expose secrets there. That's the reason why we use EnvVarSource from ImageRegistryPrivateConfiguration.

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,294 @@
...
-		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_AUTHURL", Value: d.Config.AuthURL},
 		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_CONTAINER", Value: d.Config.Container},
-		corev1.EnvVar{
-			Name: "REGISTRY_STORAGE_SWIFT_USERNAME",
-			ValueFrom: &corev1.EnvVarSource{
-				SecretKeyRef: &corev1.SecretKeySelector{
-					LocalObjectReference: corev1.LocalObjectReference{
-						Name: imageregistryv1.ImageRegistryPrivateConfiguration,
+	)
+	if d.Config.CloudsSecret != nil {
```
dmage: You don't need this condition. You should always use ImageRegistryPrivateConfiguration.

pkg/apis/imageregistry/v1/types.go
```diff
@@ -208,6 +208,8 @@ type ImageRegistryConfigStorageSwift struct {
 	DomainID  string `json:"domainID,omitempty"`
 	Tenant    string `json:"tenant,omitempty"`
 	TenantID  string `json:"tenantID,omitempty"`
+	// The name of the secret containing the openstack clouds.yaml file
+	CloudsSecret *corev1.SecretReference `json:"cloudsSecret"`
```
dmage: We don't want to have other namespaces: they require additional informers, changing RBAC rules, etc. It's better to use corev1.SecretKeySelector.

But we also need to bootstrap ourselves. When the operator gets installed to a new cluster, it should inspect ConfigMap kube-config/cluster-config-v1 and configure the storage automatically. Will the operator get this secret from the installer or will it somehow to create it? How will it distinguish secrets that are created automatically and that are provided by a user?

PR #215, opened 2019-02-25
do not setup ownerrefs for clusterscoped/cross-namespace objects

bparees:
fixes https://github.com/openshift/cluster-image-registry-operator/issues/216

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ test-unit: verify
 	./hack/test-go.sh ./cmd/... ./pkg/...
 
 test-e2e:
-	GOCACHE=off ./hack/test-go.sh -timeout 20m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
+	GOCACHE=off ./hack/test-go.sh -timeout 30m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
 
 verify:
 	hack/verify.sh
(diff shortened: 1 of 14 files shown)


bparees:
/assign @dmage 

i'm surprised this wasn't breaking us, setting up ownerrefs between objects that are not all within a single namespace results in very unpredictable results.

dmage:
@bparees our flakes are the result of deleting the image config?

/lgtm

bparees:
> @bparees our flakes are the result of deleting the image config?

potentially yes, i think in at least some cases the issue is the imageconfig being deleted/updated (to remove the internal registry hostname) which causes a control-plane restart.  That shouldn't fail our tests, but if it takes long enough it can cause them to timeout.

That said, just not removing the imageconfig won't totally fix it, because even deleting the registry instance causes an update to the imageconfig (to remove the internal registry hostname) which will still restart the control plane.

I didn't create this PR as a direct fix for the flakes so much as something else I noticed while investigating them that definitely needs to be fixed.  But..fingers crossed it makes it better.

bparees:
@dmage i had to add another commit to clean up how we were handling route ownership identification, ptal.

bparees:
/hold

removeObsoleteRoutes also need to be fixed up to check for the owner annotation instead of the owner ref. 

@dmage if you know what I'm referring to and can fix this up and merge it in a new pr in the morning that'd be great, if not I'll fix it when I get in.

dmage:
@bparees I've pushed the fix to your branch.

bparees:
thanks @dmage 
/hold cancel

bparees:
passed tests, manually merging to get us hopefully slightly less blocked.

PR #218, opened 2019-02-26
PVC: Use annotation instead of ownerref

### LINKED ISSUE #216: don't use GC for PVC cleanup
I'm removing most of our GC usage here:
https://github.com/openshift/cluster-image-registry-operator/pull/215

but I don't want to get into the PVC logic.  Right now it appears to be using GC to ensure the PVC gets removed when the CR is removed.  That won't work because the CR is cluster-scoped and GC can't span namespaces.  The PVC needs to be cleaned up via finalization.

legionus:
Fixes #216


legionus:
> aws_route.igw_route: Error creating route: timeout while waiting for state to become 'success' (timeout: 2m0s)
/retest

bparees:
this will get pulled in as part of https://github.com/openshift/cluster-image-registry-operator/pull/222

PR #219, opened 2019-02-26
WIP: operator: Report cluster operator version as per the new standard

smarterclayton:
The cluster operator should report the release image version as its
operator version when it reaches level (Available=true). This operator
version covers the entire functional area for coverage. More specific
versions may be allowed later.

If no release version is set we reset versions to empty.

This is the model we will be implementing across all operators for 4.0.

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -35,6 +35,8 @@ spec:
             requests:
               cpu: 10m
           env:
+            - name: RELEASE_VERSION
+              value: "0.0.1-snapshot"
             - name: WATCH_NAMESPACE
               valueFrom:
                 fieldRef:
--- a/manifests/08-clusteroperator.yaml
+++ b/manifests/08-clusteroperator.yaml
@@ -3,3 +3,7 @@ kind: ClusterOperator
 metadata:
   name: image-registry
 spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
--- a/pkg/clusteroperator/clusteroperator.go
+++ b/pkg/clusteroperator/clusteroperator.go
@@ -2,6 +2,7 @@ package clusteroperator
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -11,8 +12,6 @@ import (
 
 	osapi "github.com/openshift/api/config/v1"
 	osset "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
-
-	"github.com/openshift/cluster-image-registry-operator/version"
 )
 
 type StatusHandler struct {
@@ -42,14 +41,7 @@ func (s *StatusHandler) Create() error {
 		ObjectMeta: metaapi.ObjectMeta{
 			Name: s.Name,
 		},
-		Status: osapi.ClusterOperatorStatus{
-			Versions: []osapi.OperandVersion{
-				{
-					Name:    "operator",
-					Version: version.Version,
-				},
-			},
-		},
+		Status: osapi.ClusterOperatorStatus{},
 	}
 
 	_, err = client.ClusterOperators().Create(state)
@@ -104,12 +96,23 @@ func (s *StatusHandler) Update(condtype osapi.ClusterStatusConditionType, status
 			LastTransitionTime: metaapi.Now(),
 		})
 
-		desiredVersions := []osapi.OperandVersion{
-			{
-				Name:    "operator",
-				Version: version.Version,
-			},
+		// when we are at the available level, reset versions in status to the input
+		// release version
+		desiredVersions := state.Status.Versions
+		if releaseVersion := os.Getenv("RELEASE_VERSION"); len(releaseVersion) > 0 {
+			// an available operator resets release version
+			if condtype == osapi.OperatorAvailable && status == osapi.ConditionTrue {
+				desiredVersions = []osapi.OperandVersion{
+					{
+						Name:    "operator",
+						Version: releaseVersion,
+					},
+				}
+			}
+		} else {
+			desiredVersions = nil
 		}
+
 		if !reflect.DeepEqual(state.Status.Versions, desiredVersions) {
 			state.Status.Versions = desiredVersions
 			modified = true
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -48,6 +48,12 @@ func isDeploymentStatusAvailable(deploy *appsapi.Deployment) bool {
 	return deploy.Status.AvailableReplicas > 0
 }
 
+func isDeploymentStatusAvailableAndUpdated(deploy *appsapi.Deployment) bool {
+	return deploy.Status.AvailableReplicas > 0 &&
+		deploy.Status.ObservedGeneration >= deploy.Generation &&
+		deploy.Status.UpdatedReplicas == deploy.Status.Replicas
+}
+
 func isDeploymentStatusComplete(deploy *appsapi.Deployment) bool {
 	replicas := int32(1)
 	if deploy.Spec.Replicas != nil {
@@ -60,6 +66,7 @@ func isDeploymentStatusComplete(deploy *appsapi.Deployment) bool {
 }
 
 func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Deployment, applyError error, removed bool) {
+	setOperatorAvailable := true
 	operatorAvailable := osapi.ConditionFalse
 	operatorAvailableMsg := ""
 	if deploy == nil {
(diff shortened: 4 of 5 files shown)


smarterclayton:
@abhinavdahiya @derekwaynecarr this is the design that abhinav and I agreed on that minimizes scope for 4.0.  All components except core. control plane must report "operator" version which is the release version.  The first version in the list is what will be shown in status UIs.  Only components which have an exception will be allowed to use name != `operator` or have more than one.

bparees:
do we care about operand versions right now?

smarterclayton:
Will be enforced with https://github.com/openshift/origin/pull/22156

smarterclayton:
The first version in the list will be interpreted as the "functional area version".  The only component version we care about right now is "kubernetes" and even that may not make 4.0.

manifests/08-clusteroperator.yaml
```diff
@@ -3,3 +3,7 @@ kind: ClusterOperator
 metadata:
   name: image-registry
 spec: {}
+status:
+  versions:
+    - name: operator
+      version: "0.0.1-snapshot"
```
deads2k: this is the indication the CVO consumes to look for a matching level?

manifests/07-operator.yaml
```diff
@@ -35,6 +35,8 @@ spec:
             requests:
               cpu: 10m
           env:
+            - name: RELEASE_VERSION
+              value: "0.0.1-snapshot"
```
deads2k: Given a payload substitution from the CVO, is there any way for @bparees (@adambkaplan now?) to go from this version to a level of code?  Navigate via the payload metadata?

adambkaplan: @deads2k Or reference a sha in the version number (4.0.z-sha)?

manifests/08-clusteroperator.yaml
smarterclayton: This is replaced by the release image version - so for CI becomes `4.0.0-0.ci-XXXX` or `0.0.1-2019-blah-blah`.

manifests/07-operator.yaml
smarterclayton: The payload already identifies the git commit.  So `oc adm release extract --git PAYLOAD` will check out the right code, and `oc adm release info` will show the right. commit

smarterclayton:
```
		    message: "Unable to apply resources: unable to sync storage configuration: InvalidAccessKeyId:
		      The AWS Access Key Id you provided does not exist in our records.\n\tstatus
		      code: 403, request id: 9F7CF2E209519A53, host id: KmxOvvgpisxuwNCxf8xm3HQSKHryoECU6n10p2YUQBiW3sytZYtUmny006BW5ypF47nYQTzVSzE="
		    status: "True"
```

/retest

bparees:
@coreydaley we need to get to the bottom of the flake @smarterclayton just hit in https://github.com/openshift/cluster-image-registry-operator/pull/219#issuecomment-467609287

i've seen it as well.

bparees:
(I am wondering if it's a credential minter issue?)

coreydaley:
@bparees looking into it

deads2k:
I still need a way to represent the kube-apiserver version, but other than disliking naming our software versions on our payloads I could make this work in the master operators.

smarterclayton:
kube-apiserver version will be you reporting `kubernetes=0.0.1-snapshot-kubernetes` in your status versions.  The value is already in the payload.

manifests/07-operator.yaml
bparees: So all operators will have the same version, rather than unique versioning per operator?

smarterclayton: Correct.

smarterclayton:
TestCustomPVC seems to be failing on multiple PRs, not just this one.  New?

bparees:
> TestCustomPVC seems to be failing on multiple PRs, not just this one. New?

yeah I opened https://github.com/openshift/cluster-image-registry-operator/issues/223 for it. Since it doesn't seem like @legionus has a fix today i'm going to disable the test.

bparees:
(it's not introduced by your PR if that's what you're asking)

bparees:
our tests should be (relatively) reliable now.

/retest

bparees:
Passing tests... Still a wip?

manifests/07-operator.yaml
adambkaplan: @smarterclayton What about operand version numbers? cluster-openshift-controller-manager-operator comes to mind - my understanding is the CVO sets the tag for the referenced images, ex origin-hypershift, origin-docker-builder, and origin-deployer

pkg/clusteroperator/clusteroperator.go
```diff
@@ -104,12 +105,23 @@ func (s *StatusHandler) Update(condtype osapi.ClusterStatusConditionType, status
...
+			if condtype == osapi.OperatorAvailable && status == osapi.ConditionTrue {
+				desiredVersions = []osapi.OperandVersion{
+					{
+						Name:    "operator",
+						Version: releaseVersion,
+					},
+				}
+			}
+		} else {
+			desiredVersions = nil
```
bparees: So to make sure i understand..this logic means that basically as long as the registry is considered available (pods running, regardless of image version), the operator is going to report its version as being the current operator version.

Meaning when we go from registry operator v4.0 to registry operator v4.1, the operator will basically immediately report itself as being at v4.1 (assuming the registry was available already).

At some point subsequent to that, the registry operator will redeploy the registry (to move it to the new v4.1 image), at which point the registry might go unavailable(if for some reason it can't come up), at which point the registry operator will still report being at v4.1, but not unavailable.

Is that our meets min here, or do we need tighter logic in the registry operator that ensures the registry pods are running at the new version before the operator reports itself as being at the new version?

smarterclayton: That doesn't meet the min.  I'm trying to interpret how your operator knows that it has reached level.

bparees: right now i'd say it probably does not know, at least not in a tightly coupled way.

It knows it has updated the deployment to use the new image, and it knows whether the deployment is available or not.  But it doesn't know whether all pods in the deployment are running the new image level or not.  That will require more logic around looking at the deployment.

at a minimum your logic above should be gating on available==true && progressing==false.  That still doesn't get us 100% of the way there i don't think, but it's closer to reflecting the reality of "we have achieved the new level"

smarterclayton: Actually, looking at operator/controller it does the following:

1. update the deployment
2. load the deployment object
3. pass the deployment object to syncStatus
4. check if the deployment is available
5. sets available true if the deployment is available

The condition that needs to be checked is `isDeploymentStatusAvailable` but with ensuring the generation is updated (like the CVO does).

I believe with that change it would be sufficient to state that you are now available for the version.

bparees: > The condition that needs to be checked is `isDeploymentStatusAvailable` but with ensuring the generation is updated (like the CVO does)

yeah that's roughly what was in my head.  However we are still "available=true" even if the generation is not updated.  however we would be progressing=true.

we need to:

1) mark ourselves progressing=true and some marker about the generation the deployment is at so we know when the deployment achieves the new generation
2) update the deployment to the new image
3) wait for the deployment achieved generation to be updated
4) now we can mark progressing=false and set the new version

smarterclayton: we don't really need the marker - it all happens in the same sync loop.  we only need to reset the versions when we hit "level" and what we should really be doing is avoiding changing the available condition status (skip updateCondition) when we have available replicas but it's not up to date

smarterclayton: Updated with a slight tweak

pkg/operator/status.go
```diff
@@ -48,6 +48,12 @@ func isDeploymentStatusAvailable(deploy *appsapi.Deployment) bool {
 	return deploy.Status.AvailableReplicas > 0
 }
 
+func isDeploymentStatusAvailableAndUpdated(deploy *appsapi.Deployment) bool {
+	return deploy.Status.AvailableReplicas > 0 &&
+		deploy.Status.ObservedGeneration >= deploy.Generation &&
+		deploy.Status.UpdatedReplicas == deploy.Status.Replicas
```
bparees: I think it's still possible that you see the old deployment version here, since we use lister caches.  (or for that matter if we just whiff completely on updating the deployment to the new image version).

I'm going to pull this commit into a new PR and iterate on it.  We need to actually confirm that the deployment which is available also reflects the new version.

smarterclayton:
We clarified:

1. "at version level" is is a superset of "is available".   It involves making sure we have correctly observed the rolled out version of the code that came with this operator, which usually means: at least one sync succeeded, we look at the result of the sync and verify deployments and daemonsets are updated and rolled out (ensure happens before of the specs).
2. if you have two separate loops (one for sync and one for status) or if you read from a cache, you need something to guarantee that you observed the latest spec.  For the network PR I chose to write an annotation to the deployments/daemonsets and then read them back.  Ben was going to do similar.

bparees:
carrying this forward in https://github.com/openshift/cluster-image-registry-operator/pull/225

bparees:
merged in https://github.com/openshift/cluster-image-registry-operator/pull/225

Issue #221, opened 2019-02-26
TestAWSUnableToCreateBucketOnStartup InvalidAccessKeyId errors during tests


coreydaley:
message: "Unable to apply resources: unable to sync storage configuration: InvalidAccessKeyId:
		      The AWS Access Key Id you provided does not exist in our records.\n\tstatus
		      code: 403, request id: 9F7CF2E209519A53, host id: KmxOvvgpisxuwNCxf8xm3HQSKHryoECU6n10p2YUQBiW3sytZYtUmny006BW5ypF47nYQTzVSzE="
		    status: "True"

bparees:
seen here:

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_cluster-image-registry-operator/222/pull-ci-openshift-cluster-image-registry-operator-master-e2e-aws-operator/1199/

bparees:
I am disabling this test in https://github.com/openshift/cluster-image-registry-operator/pull/222/commits/fe0ec949846cfe03c2d125174126cfa4c5048a01 so we can unblock our queue.  Please fix+reenable the test asap.

PR #222, opened 2019-02-26
disable kube/openshift api server operators before running tests

bparees:


bparees:
/assign @dmage 

I feel good about this one.

bparees:
flakes:
https://github.com/openshift/cluster-image-registry-operator/issues/221
https://github.com/openshift/cluster-image-registry-operator/issues/223

both seem new/unrelated to the the GC/+apiserver issues that i think this PR fixes.

bparees:
Added a commit to disable the two flaky tests.

PR #225, opened 2019-02-28
operator: Report cluster operator version as per the new standard

bparees:


bparees:
@smarterclayton ok i think this aligns with what we discussed now.  ptal.

/cc @dmage 

i'm going to try to add an e2e to configuration_test that updates RELEASE_VERSION and confirms we see clusteroperator report that version, but what's here is ready for review.

bparees:
very basic clusteroperator version upgrade/reporting test added.

A better test would ensure we don't report the new version until the operand deployment is at the new version, but i think what's here is a sufficient starting point.

pkg/apis/imageregistry/v1/types.go
```diff
@@ -66,6 +66,10 @@ const (
 	// StorageIncompleteUploadCleanupEnabled denotes whethere or not the registry storage
 	// medium is configured to automatically cleanup incomplete uploads
 	StorageIncompleteUploadCleanupEnabled = "StorageIncompleteUploadCleanupEnabled"
+
+	// VersionAnnotation reflects the version of the registry that this deployment
+	// is running.
+	VersionAnnotation = "imageregistry.operator.openshift.io/version"
```
smarterclayton: I used “release.openshift.io/version” in the network operator.  Better to have a convention.

pkg/clusteroperator/clusteroperator.go
```diff
@@ -97,22 +88,24 @@ func (s *StatusHandler) Update(condtype osapi.ClusterStatusConditionType, status
...
+			newVersions := []osapi.OperandVersion{
+				{
+					Name:    "operator",
+					Version: newVersion,
+				},
+			}
+			if !reflect.DeepEqual(state.Status.Versions, newVersions) {
+				state.Status.Versions = newVersions
+				modified = true
+			}
```
smarterclayton: You need to reset versions to empty if new version is unset.

smarterclayton: Hrm.  I would not pass empty string to mean “not available”

pkg/operator/status.go
```diff
@@ -71,9 +82,16 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
 		operatorAvailableMsg = "Deployment has minimum availability"
+		setOperatorVersion = isDeploymentStatusAvailableAndUpdated(deploy)
 	}
 
-	err := c.clusterStatus.Update(osapi.OperatorAvailable, operatorAvailable, operatorAvailableMsg)
+	deploymentVersion := ""
+	// if the current deployment has achieved availability at the new level, set the operator reported
+	// version to whatever level the deployment has achieved.
+	if setOperatorVersion {
+		deploymentVersion = deploy.Annotations[imageregistryv1.VersionAnnotation]
```
smarterclayton: Hrm... ok, I guess.  Seems a bit weird but I can see why.

pkg/apis/imageregistry/v1/types.go
bparees: works for me, i'll update it in the morning.

bparees:
@smarterclayton updated the annotation name.  My interpretation of your other comments was everything was ok if slightly odd, but let me know if i missed something.

bparees:
> I'm a bit worried about the special case newVersion = "" (which means that the version is not changed). But it looks good to merge.

`newVersion=""` means "we are not in a position to publish what version level we are at (e.g. because we are unavailable)"

PR #229, opened 2019-03-04
Fix namespace for errors for installer-cloud-credentials

dmage:
https://bugzilla.redhat.com/show_bug.cgi?id=1684738

/cc @coreydaley

### DIFF
--- a/pkg/clusterconfig/clusterconfig.go
+++ b/pkg/clusterconfig/clusterconfig.go
@@ -120,18 +120,18 @@ func GetAWSConfig(listers *regopclient.Listers) (*Config, error) {
 		})
 
 		if sec == nil || pollErr != nil {
-			return nil, fmt.Errorf("unable to get cluster minted credentials %q: %v", fmt.Sprintf("%s/%s", installerConfigNamespace, cloudCredentialsName), pollErr)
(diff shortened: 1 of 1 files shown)


pkg/clusterconfig/clusterconfig.go
```diff
@@ -108,7 +108,7 @@ func GetAWSConfig(listers *regopclient.Listers) (*Config, error) {
 	sec, err := listers.Secrets.Get(imageregistryv1.ImageRegistryPrivateConfigurationUser)
 	if err != nil && errors.IsNotFound(err) {
 		pollErr := wait.PollImmediate(1*time.Second, 5*time.Minute, func() (stop bool, err error) {
-			sec, err = client.Secrets(imageregistryv1.ImageRegistryOperatorNamespace).Get(cloudCredentialsName, metav1.GetOptions{})
+			sec, err = client.Secrets(installerConfigNamespace).Get(cloudCredentialsName, metav1.GetOptions{})
```
coreydaley: @dmage it's actually the errors that are incorrect (https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/clusterconfig/clusterconfig.go#L122-L134), they should have the imageregistryv1.ImageRegistryOperatorNamespace instead of the installerConfigNamespace, as that was the old location for the kube-system/aws-creds secret that we no longer use.

dmage: I misread the bug. Thanks, Corey. I've updated the PR.

PR #230, opened 2019-03-04
resource: specify system-cluster-critical priority for registry

derekwaynecarr:
specifies system-cluster-critical priority for operands that run on masters.
blocks passing smoke tests for ensuring control plane pods always schedule.

see: openshift/origin#22217


bparees:
@derekwaynecarr needs rebase, but otherwise lgtm.

PR #234, opened 2019-03-05
[4.1] Add cloudfront support

legionus:
/assign @dmage 
/hold


adambkaplan:
@legionus I've created the `master-4.1` branch. You can update the PR to merge there.

@dmage I believe our tooling should fast forward this branch on top of `master` - you may want to keep an eye on that just to be sure.

dmage:
As we don't have a long merge queue, I'd prefer just to use master when it will be open for 4.1.

PR #235, opened 2019-03-08
add missing s3 endpoint

FlorinPeter:
Currently it is not possible to change the s3 storage to another region by setting the RegionEndpoint. This PR will fix that issue.


coreydaley:
@FlorinPeter Currently Region is set based on the cluster install configuration, however RegionEndpoint is not guaranteed to be set.  Will this throw an error if RegionEndpoint is blank?

FlorinPeter:
> @FlorinPeter Currently Region is set based on the cluster install configuration, however RegionEndpoint is not guaranteed to be set. Will this throw an error if RegionEndpoint is blank?

@coreydaley no because RegionEndpoint (https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/apis/imageregistry/v1/types.go#L168) is set to omitempty and the s3 config will not use it if not set.

coreydaley:
@dmage Can you approve this please?

PR #237, opened 2019-03-09
openstack: add domain and tenant config

FlorinPeter:
this is needed for openstack to make the registry run on swift

### DIFF
--- a/pkg/apis/imageregistry/v1/types.go
+++ b/pkg/apis/imageregistry/v1/types.go
@@ -185,6 +185,10 @@ type ImageRegistryConfigStorageGCS struct {
 type ImageRegistryConfigStorageSwift struct {
 	AuthURL   string `json:"authURL,omitempty"`
 	Container string `json:"container,omitempty"`
+	Domain    string `json:"domain,omitempty"`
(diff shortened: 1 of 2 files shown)


FlorinPeter:
@coreydaley can you please approve or do we need someone else?

pkg/apis/imageregistry/v1/types.go
```diff
@@ -185,6 +185,10 @@ type ImageRegistryConfigStorageGCS struct {
 type ImageRegistryConfigStorageSwift struct {
 	AuthURL   string `json:"authURL,omitempty"`
 	Container string `json:"container,omitempty"`
+	Domain    string `json:"domain,omitempty"`
+	DomainId  string `json:"domainId,omitempty"`
```
dmage: style: in Go it's better to use [DomainID instead of DomainId][1]. The same works for [Kubernetes json tags][2].

[1]: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
[2]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#containerstatus-v1-core

FlorinPeter: thx, I changed it

PR #246, opened 2019-03-25
Add reasons to conditions

dmage:


pkg/operator/status.go
```diff
@@ -69,19 +76,56 @@ func isDeploymentStatusComplete(deploy *appsapi.Deployment) bool {
...
+		operatorAvailable.Reason = "DeploymentDeleted"
 	} else if !isDeploymentStatusAvailable(deploy) {
-		operatorAvailableMsg = "Deployment does not have available replicas"
+		operatorAvailable.Message = "Deployment does not have available replicas"
+		operatorAvailable.Reason = "NoReplicasAvailable"
 	} else {
-		operatorAvailable = osapi.ConditionTrue
-		operatorAvailableMsg = "Deployment has minimum availability"
+		operatorAvailable.Status = configapiv1.ConditionTrue
+		operatorAvailable.Message = "Deployment has minimum availability"
```
bparees: can we distinguish between "minimal availability" and "we're fully deployed" and convey that to the user?  As an admin i'd be nervous seeing "deployment has minimum availability" because it sounds like a warning.  I'd be wondering why it's only at the minimum.

pkg/operator/status.go
```diff
@@ -91,85 +135,83 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
-		operatorProgressingMsg = "Unmanaged"
+	operatorProgressing := clusteroperator.ConditionState{
+		Status:  configapiv1.ConditionTrue,
+		Message: "",
+		Reason:  "",
+	}
+	if cr.Spec.ManagementState == operatorapiv1.Unmanaged {
+		operatorProgressing.Status = configapiv1.ConditionFalse
+		operatorProgressing.Message = "Unmanaged"
+		operatorProgressing.Reason = "Unmanaged"
```
bparees: "The registry configuration is set to unmanaged mode"

pkg/operator/status.go
```diff
@@ -91,85 +135,83 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
 	} else if removed {
 		if deploy != nil {
-			operatorProgressingMsg = "The deployment still exists"
+			operatorProgressing.Message = "The deployment still exists"
```
bparees: "The deployment is being removed" ?

pkg/operator/status.go
```diff
@@ -91,85 +135,83 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
+			operatorProgressing.Reason = "DeletingDeployment"
 		} else {
-			operatorProgressing = osapi.ConditionFalse
-			operatorProgressingMsg = "Everything is removed"
+			operatorProgressing.Status = configapiv1.ConditionFalse
+			operatorProgressing.Message = "Everything is removed"
```
bparees: "All registry resources are removed" ?

pkg/operator/status.go
```diff
@@ -91,85 +135,83 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
+		operatorProgressing.Reason = "FinalizingDeployment"
 	} else if !isDeploymentStatusComplete(deploy) {
-		operatorProgressingMsg = "The deployment has not completed"
+		operatorProgressing.Message = "The deployment has not completed"
+		operatorProgressing.Reason = "DeploymentNotCompleted"
 	} else {
-		operatorProgressing = osapi.ConditionFalse
-		operatorProgressingMsg = "Everything is ready"
+		operatorProgressing.Status = configapiv1.ConditionFalse
+		operatorProgressing.Message = "Everything is ready"
```
bparees: "The registry is ready" ?

dmage:
@bparees thanks, "Everything" indeed sounds not so clear. All comments are addressed.

PR #248, opened 2019-04-04
Show false values in config's spec

dmage:
To have the default route users need to set spec.defaultRoute to true. But by default you won't see this field when you try to `kubectl edit` our custom resource because its default value is False. Changing `false` to `true` is more convenient than remembering the exact name of the field.

/assign @adambkaplan @legionus


pkg/apis/imageregistry/v1/types.go
```diff
@@ -108,23 +108,23 @@ type ImageRegistrySpec struct {
...
 	// should be created using the default generated hostname
-	DefaultRoute bool `json:"defaultRoute,omitempty"`
+	DefaultRoute bool `json:"defaultRoute"`
 
 	// Routes defines additional external facing routes which should be created for the registry
 	Routes []ImageRegistryConfigRoute `json:"routes,omitempty"`
 
 	// Replicas determines the number of registry instances to run
-	Replicas int32 `json:"replicas,omitempty"`
+	Replicas int32 `json:"replicas"`
```
adambkaplan: I'm concerned about this defaulting to 0 - does the operator force this to be at least 1?

legionus: @adambkaplan No.

dmage: It doesn't change the default value. It will show the field in `kubectl edit` and `kubectl get -o yaml` even if its value is zero.

PR #249, opened 2019-04-10
fix config resource name in readme

bparees:


coreydaley:
@bparees 
cc @smarterclayton 

This seems like a HUGE waste of time/money/resources running this test umpteen (I live in Kentucky, that is a word here) times for a simple README change (assuming there are no tests that actually check the content of the document).  Can we add something to CI to not test on markdown file changes or something like that?

Just my $.02 ....

bparees:
Probably not that simple. Markdown could be used by read by code, or tests,
for example.

Anyway things like this are a sign our Ci is unhealthy. The effort should
be put into resolving why a pr like this can't sail through ci, because
every pr is impacted by those problems.

The testing cost in dollars for PRs like this is non zero but it's
relatively miniscule against our pr testing expenditure in general.


Ben Parees | OpenShift

PR #252, opened 2019-04-11
Fix e2e tests

dmage:


Makefile
```diff
@@ -21,7 +21,7 @@ test-unit: verify
 	./hack/test-go.sh ./cmd/... ./pkg/...
 
 test-e2e:
-	GOCACHE=off ./hack/test-go.sh -timeout 30m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
+	./hack/test-go.sh -count 1 -timeout 30m -v$${WHAT:+ -run="$$WHAT"} ./test/e2e/
```
dmage: [The idiomatic way to disable test caching explicitly is to use -count=1.](https://golang.org/pkg/cmd/go/internal/test/)

dmage: Go 1.12 fails when it sees GOCACHE.

PR #255, opened 2019-04-16
Fix PVC access mode validation 

legionus:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1700294
/assign @dmage

### DIFF
--- a/pkg/storage/pvc/pvc.go
+++ b/pkg/storage/pvc/pvc.go
@@ -3,7 +3,6 @@ package pvc
 import (
 	"fmt"
 	"reflect"
-	"strings"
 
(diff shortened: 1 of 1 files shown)


alvaroaleman:
@legionus RWX is not supported by most provisioners, hence this change basically makes PVC as source useless. What about instead setting `.Spec.Strategy: Recreate` on the registry deployment?

dmage:
@alvaroaleman we don't want to have downtimes during upgrades and configuration changes. Which cloud platform do you use and what volume plugin do you want to use for your persistent volume?

alvaroaleman:
@dmage I am using vsphere and would like to use its volumes.

PR #257, opened 2019-04-16
Remove unsupported storages

### LINKED ISSUE #256: [api-freeze] Verify operator API is frozen
*Backwards compatibility of GA APIs*

What constitutes a backwards compatible API?

1. Fields can not be removed (this also means you can’t rename a field)
2. Serialized defaults can not change.  Unserialized defaults in operators can change over time.
3. Behavior (meaning) of a field can not change.  This means that old fields must be maintained even if new fields supercede them.
4. New fields can be introduced that supercede existing fields, but values in the existing fields must be respected when present.

Other things that should be considered “API”

1. Names of singleton CRs in openshift-config (user-provided) and openshift-config-managed (operator-provided) namespaces that are used for cross-communication of components
2. Names / namespaces of Services used for cross-communication of components
3. Annotations and labels used for cross-communication of components

/assign @dmage

dmage:
As a part of API stabilization we need to remove unsupported storage backends. Fixes #256.


pkg/apis/imageregistry/v1/types.go
```diff
@@ -191,38 +191,17 @@ type ImageRegistryConfigStorageS3 struct {
...
 }
 
 type ImageRegistryConfigStorage struct {
-	Azure      *ImageRegistryConfigStorageAzure      `json:"azure,omitempty"`
-	Filesystem *ImageRegistryConfigStorageFilesystem `json:"filesystem,omitempty"`
-	GCS        *ImageRegistryConfigStorageGCS        `json:"gcs,omitempty"`
-	S3         *ImageRegistryConfigStorageS3         `json:"s3,omitempty"`
-	Swift      *ImageRegistryConfigStorageSwift      `json:"swift,omitempty"`
-	PVC        *ImageRegistryConfigStoragePVC        `json:"pvc,omitempty"`
+	EmptyDir *ImageRegistryConfigStorageEmptyDir `json:"emptydir,omitempty"`
```
adambkaplan: /s/emptydir/emptyDir/

pkg/apis/imageregistry/v1/types.go
```diff
@@ -191,38 +191,17 @@ type ImageRegistryConfigStorageS3 struct {
...
-}
-
-type ImageRegistryConfigStorageSwift struct {
-	AuthURL   string `json:"authURL,omitempty"`
-	Container string `json:"container,omitempty"`
-	Domain    string `json:"domain,omitempty"`
-	DomainID  string `json:"domainID,omitempty"`
-	Tenant    string `json:"tenant,omitempty"`
-	TenantID  string `json:"tenantID,omitempty"`
-}
```
adambkaplan: Since swift storage was added by a customer, we may mark this as experimental

adambkaplan:
@FlorinPeter we may need to remove the swift storage configuration that you contributed, as we are not ready to support this until a future release. Is it OK if we move this to a feature branch, or are you using it to install the registry on OpenStack?

FlorinPeter:
@adambkaplan yes I am using this part in OCP4 HTB.
What is the problem with this part? 

@tomassedovic I guess you also use this part?

pkg/apis/imageregistry/v1/types.go
soltysh: yeah, if there are users, it's better to have it as experimental rather than breaking them entirely

dmage: @soltysh how to mark them as experimental?

Fedosin:
@adambkaplan hi! I think it's okay to remove Swift storage for a while. I'll rebase my PR #211 in that case.

pkg/apis/imageregistry/v1/types.go
soltysh: In the comments like here: https://github.com/openshift/api/blob/master/build/v1/types.go#L654-L659

pkg/apis/imageregistry/v1/types.go
```diff
@@ -191,38 +191,17 @@ type ImageRegistryConfigStorageS3 struct {
...
 type ImageRegistryConfigStorage struct {
-	Azure      *ImageRegistryConfigStorageAzure      `json:"azure,omitempty"`
-	Filesystem *ImageRegistryConfigStorageFilesystem `json:"filesystem,omitempty"`
-	GCS        *ImageRegistryConfigStorageGCS        `json:"gcs,omitempty"`
-	S3         *ImageRegistryConfigStorageS3         `json:"s3,omitempty"`
-	Swift      *ImageRegistryConfigStorageSwift      `json:"swift,omitempty"`
-	PVC        *ImageRegistryConfigStoragePVC        `json:"pvc,omitempty"`
+	EmptyDir *ImageRegistryConfigStorageEmptyDir `json:"emptydir,omitempty"`
+	S3       *ImageRegistryConfigStorageS3       `json:"s3,omitempty"`
+	PVC      *ImageRegistryConfigStoragePVC      `json:"pvc,omitempty"`
```
adambkaplan: Need godoc for all three (and Swift as we may keep it as Experimental)

FlorinPeter: Would be great to have Swift marked as Experimental 👍

dmage:
@adambkaplan ptal

pkg/apis/imageregistry/v1/types.go
```diff
@@ -208,21 +201,26 @@ type ImageRegistryConfigStorageSwift struct {
...
-	GCS        *ImageRegistryConfigStorageGCS        `json:"gcs,omitempty"`
-	S3         *ImageRegistryConfigStorageS3         `json:"s3,omitempty"`
-	Swift      *ImageRegistryConfigStorageSwift      `json:"swift,omitempty"`
-	PVC        *ImageRegistryConfigStoragePVC        `json:"pvc,omitempty"`
+	// EmptyDir represents an experimental storage that uses a temporary
+	// directory that shares a pod's lifetime. This storage is not suitable
+	// for production use. It can't be used with more than 1 replica. When the
+	// pod is removed from a node for any reason, the data in the emptyDir is
+	// deleted forever.
+	EmptyDir *ImageRegistryConfigStorageEmptyDir `json:"emptyDir,omitempty"`
```
adambkaplan: Suggested godoc:

```suggestion
	// EmptyDir represents ephemeral storage on the pod's host node.
	// This storage cannot be used with more than 1 replica and is not suitable
	// for production use. When the pod is removed from a node for any reason, 
	// the data in the emptyDir is deleted forever.
	EmptyDir *ImageRegistryConfigStorageEmptyDir `json:"emptyDir,omitempty"`
```

pkg/apis/imageregistry/v1/types.go
```diff
@@ -208,21 +201,26 @@ type ImageRegistryConfigStorageSwift struct {
...
+	// S3 represents configuration that uses Amazon Simple Storage Service.
+	S3 *ImageRegistryConfigStorageS3 `json:"s3,omitempty"`
+	// Swift represents experimental configuration that uses OpenStack Object
+	// Storage.
```
adambkaplan:
```suggestion
	// Swift represents configuration that uses OpenStack Object Storage.
	// This configuration is EXPERIMENTAL and is subject to change without notice.
```

pkg/apis/imageregistry/v1/types.go
dmage: You've dropped "experimental" from the description. This storage may be removed from the next releases (or we may want to remove it right now).

adambkaplan:
Per conversation with @dmage and comments in this PR, Swift and EmptyDir storage will _both_ be marked as EXPERIMENTAL APIs, subject to modification or removal in a future release.

PR #258, opened 2019-04-20
migrate to glide and bump(*)

coreydaley:
I know that glide isn't being actively developed anymore, but the majority of the go projects under the openshift organization are using it, and maybe there will be a future migration effort that this can be a part of to a new tool.


adambkaplan:
/hold

I want to make sure that we don't introduce any instabilities given that a large number of deps got updated.

wking:
We need openshift/api bumped so we can migrate off the deprecated cluster-config-v1 and over to `InfrastructureName` from openshift/api#266 for [rhbz#1685089][1].  Should I wait on this PR, or file a bump for just openshift/api that sticks with `dep`?

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1685089#c11

coreydaley:
@wking this one will probably be on hold for awhile, so doing your own bump is advised.

PR #259, opened 2019-04-22
Bug 1701422: use port 443 for registry service

adambkaplan:
Use the default HTTPS port for the internal registry service.
This lets us drop the port when referencing images on cluster.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -66,6 +66,8 @@ func NewController(kubeconfig *restclient.Config) (*Controller, error) {
 
 	p := parameters.Globals{}
 
+	p.Service.Name = imageregistryv1.ImageRegistryName
+
 	p.Deployment.Namespace = namespace
 	p.Deployment.Labels = map[string]string{"docker-registry": "default"}
 
@@ -75,7 +77,6 @@ func NewController(kubeconfig *restclient.Config) (*Controller, error) {
 	p.Healthz.Route = "/healthz"
 	p.Healthz.TimeoutSeconds = 5
 
-	p.Service.Name = imageregistryv1.ImageRegistryName
 	p.ImageConfig.Name = "cluster"
 	p.CAConfig.Name = imageregistryv1.ImageRegistryCertificatesName
 
--- a/test/framework/imageregistry.go
+++ b/test/framework/imageregistry.go
@@ -271,8 +271,17 @@ func ensureInternalRegistryHostnameIsSet(logger Logger, client *Clientset) error
 		} else if err != nil {
 			return false, err
 		}
-		if cfg == nil || cfg.Status.InternalRegistryHostname != "image-registry.openshift-image-registry.svc:5000" {
+		if cfg == nil {
 			return false, nil
+		}
+		if len(cfg.Status.InternalRegistryHostname) == 0 {
+			return false, nil
+		}
+		// Bug 1701422: drop port number since we are using HTTPS default port on service
+		if cfg.Status.InternalRegistryHostname != "image-registry.openshift-image-registry.svc" {
+			return false, fmt.Errorf("expected internal registry hostname %s; got %s",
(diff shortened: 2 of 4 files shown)


pkg/operator/controller.go
```diff
@@ -66,6 +66,10 @@ func NewController(kubeconfig *restclient.Config) (*Controller, error) {
 
 	p := parameters.Globals{}
 
+	p.Service.Name = imageregistryv1.ImageRegistryName
+	p.Service.Port = 443
```
dmage: Having this as a parameter may give wrong feeling that this a parameter that can be changed. But at the same time we have `fmt.Sprintf("%s.%s.svc", svc.Name, svc.Namespace)` that makes this value impossible to change without additional modifications. It worth mention that we rely on the fact that this port is the default port Docker.

adambkaplan: I can drop this (in which case I can drop the TargetPort as well, since that should be the port on the container)

test/framework/imageregistry.go
```diff
@@ -271,9 +271,18 @@ func ensureInternalRegistryHostnameIsSet(logger Logger, client *Clientset) error
...
 		}
+		if len(cfg.Status.InternalRegistryHostname) == 0 {
+			return false, nil
+		}
+		// Bug 1701422: drop port number since we are using HTTPS default port on service
+		if cfg.Status.InternalRegistryHostname != "image-registry.openshift-image-registry.svc" {
+			return false, fmt.Errorf("expected internal registry hostname %s; got %s",
+				"image-registry.openshift-image-registry.svc",
+				cfg.Status.InternalRegistryHostname)
+		}
```
adambkaplan: Fail fast if the internal registry host name doesn't match what we expect.

adambkaplan:
/hold

Not clear why we are consistently failing

adambkaplan:
got past the auth issue, now seeing a service name collision on imagestream tag import:
```
STEP: trying to tag a docker image below limit map[openshift.io/image-tags:{{0 0} {<nil>} 0 DecimalSI}]
Apr 26 16:44:41.848: INFO: Running 'oc import-image --config=/tmp/configfile546157238 --namespace=e2e-test-limitrange-admission-ls5dk stream:dockerimage --confirm --insecure --from image-registry.openshift-image-registry.svc:5000/e2e-test-limitrange-admission-ls5dk/src@sha256:90adb61c4cefbfc6de612caff4c5d964bba789cc3244ccc50cdd3603041ef9d0'
error: tag  failed: Internal error occurred: Get http://image-registry.openshift-image-registry.svc:5000/v2/: dial tcp 172.30.138.105:5000: connect: no route to host
imagestream.image.openshift.io/stream imported with errors

Name:			stream
Namespace:		e2e-test-limitrange-admission-ls5dk
Created:		Less than a second ago
Labels:			<none>
Annotations:		openshift.io/image.dockerRepositoryCheck=2019-04-26T16:44:47Z
Image Repository:	image-registry.openshift-image-registry.svc/e2e-test-limitrange-admission-ls5dk/stream
Image Lookup:		local=false
Unique Images:		0
Tags:			1

dockerimage
  tagged from image-registry.openshift-image-registry.svc:5000/e2e-test-limitrange-admission-ls5dk/src@sha256:90adb61c4cefbfc6de612caff4c5d964bba789cc3244ccc50cdd3603041ef9d0
    will use insecure HTTPS or HTTP connections

  ! error: Import failed (InternalError): Internal error occurred: Get http://image-registry.openshift-image-registry.svc:5000/v2/: dial tcp 172.30.138.105:5000: connect: no route to host
      Less than a second ago
STEP: waiting for an is importer to import a tag dockerimage into a stream stream
```

adambkaplan:
Per thread in aos-devel, we don't plan on landing this until 4.2

adambkaplan:
/close

Replaced by #322

PR #260, opened 2019-04-23
pkg: Migrate from cluster-config-v1 to Infrastructure.config.openshift.io

wking:
Mostly, this is replacing `cluster-config-v1` (which [is deprecated][1]) with the `Infrastructure` config object.  As part of that, I'm dropping `UserTags` support (see discussion [here][2] and [here][3]).  I'm also replacing the `openshiftClusterID` tag with a `kubernetes.io/cluster/{infraName}: owned` tag, see [here][4] and [here][5].

This also replaces `OperatorFailing` with `OperatorDegraded`, catching up with openshift/api#287.

I've taken the vendor bump in two steps, since `dep` sees a lot in master's `vendor/` that it doesn't think we need.  Let me know if you'd rather have me squash that down into a single commit, or split the second vendor bump out into individual pivots.

[EDIT: expanding on the above motivation a bit, the `OperatorFailing` -> `OperatorDegraded` bit is going to be a 4.1 blocker soon (registry Bugzilla should be landing tonight, for now, there is [this][6]).  Dropping cluster-config-v1 in favor of Infrastructure simplifies user-facing docs because all you need to tear down an AWS cluster is the `kubernetes.io/cluster/{infrastructureName}: owned` tag.  And it's [tied up in the 4.1-targetted rhbz#1685089][5].  There may be some slush about delivering that bug, although [at one point it was a beta-blocker][7]].

[1]: https://github.com/openshift/installer/issues/680
[2]: https://github.com/openshift/api/pull/231#discussion_r264390588
[3]: https://github.com/openshift/api/pull/266
[4]: https://github.com/openshift/installer/pull/1280
[5]: https://bugzilla.redhat.com/show_bug.cgi?id=1685089#c11
[6]: https://bugzilla.redhat.com/show_bug.cgi?id=1702087
[7]: https://bugzilla.redhat.com/show_bug.cgi?id=1685089#c6

PR #267, opened 2019-04-26
Infrastructure name

coreydaley:
Replaces #260 
I hope that the few updates that I made will resolve the issues that we were seeing with #260, but the tests will tell.


coreydaley:
/assign @dmage @adambkaplan 
cc @wking

coreydaley:
@bparees @smarterclayton Do we need to disable checking for OWNER files in the vendor directory?
https://github.com/openshift/cluster-image-registry-operator/pull/267#issuecomment-486912225

wking:
> Do we need to disable checking for OWNER files...

Or prune non-Go, see my commit message [here][1].

[1]: https://github.com/openshift/cluster-image-registry-operator/pull/260/commits/788ebbc433e00db83f0798a9c0f86184d4f6803e

dmage:
@coreydaley 

```
I0426 07:29:33.590569       1 bootstrap.go:38] generating registry custom resource
E0426 07:29:33.592481       1 controller.go:240] unable to sync: unable to get storage configuration from cluster install config: infrastructures "cluster" is forbidden: User "system:serviceaccount:openshift-image-registry:cluster-image-registry-operator" cannot get resource "infrastructures" in API group "" at the cluster scope, requeuing
```

It's weird that in the error message the API group is empty.

coreydaley:
@wking Any thoughts on the error? 
https://github.com/openshift/cluster-image-registry-operator/pull/267#issuecomment-487012535

coreydaley:
/hold 
Need to resolve this before it merges:
```
if infra.Status.Platform == configv1.AWSPlatformType {
		cfg.Storage.S3.Region = "us-east-1" // FIXME: installConfig.Platform.AWS.Region
	}
```

coreydaley:
> > Do we need to disable checking for OWNER files...
> 
> Or prune non-Go, see my commit message [here](https://github.com/openshift/cluster-image-registry-operator/pull/260/commits/788ebbc433e00db83f0798a9c0f86184d4f6803e).

Shouldn't need to do that, it needs to be updates for all CI/CD

lavalamp:
> @bparees @smarterclayton Do we need to disable checking for OWNER files in the vendor directory?

Yes, you do :)

coreydaley:
Blocked by https://github.com/openshift/api/pull/300

wking:
openshift/installer#1725 is in flight to unblock this.

adambkaplan:
per @wking - we can push this to 4.2.
On upgrade we should do the following:
1. Look for the existing `openshiftClusterID` tag, remove if present
2. Add the new `kubernetes.io/cluster/{infraName}` tag

wking:
I dunno if you can adjust both tags in a single, atomic operation.  If you can't, you'll want to add the `kubernetes.io/cluster/{infraName}` tag first and then remove the `openshiftClusterID` tag to avoid potentially leaking the bucket if you die mid-retag and the cluster is subsequently destroyed.

coreydaley:
@wking @adambkaplan 
Is there a specific reason that we need to remove the old tag?  It shouldn't hurt anything to leave it alone I would think.

wking:
> Is there a specific reason that we need to remove the old tag?

Just reducing possible user confusion by removing a tag that no longer had semantic meaning.

coreydaley:
Blocked by https://github.com/openshift/installer/pull/1725

coreydaley:
@dmage ptal

coreydaley:
/retest
VPC limit exceeded

coreydaley:
@dmage ptal, tests are passing, Skipping baremetal test as we discussed.

PR #272, opened 2019-04-30
Bug 1703941: use image-registry-certificates for system and user certificates

dmage:
A config map with the annotation `service.beta.openshift.io/inject-cabundle` can contain only `service-ca.crt`. We need to combine user-provided certificates with this service-ca.crt save the result as `image-registry-certificates`.

https://bugzilla.redhat.com/show_bug.cgi?id=1703941


bparees:
> A config map with the annotation service.beta.openshift.io/inject-cabundle can contain only service-ca.crt

was this a recent change?  we had tested this previously right?

@adambkaplan does this affect the configmap that the buildcontroller creates for builds, also?

adambkaplan:
/hold

@bparees this affects builds as well. Observing this in my testing.

@openshift/sig-auth is the API for the annotation frozen? If possible we shouldn't ship GA code with beta APIs.

pkg/resource/caconfig.go
```diff
@@ -60,17 +66,34 @@ func (gcac *generatorCAConfig) GetName() string {
...
+		return cm, err
+	} else {
+		if cert, ok := serviceCA.Data["service-ca.crt"]; ok {
+			internalHostname, err := getServiceHostname(gcac.serviceLister, gcac.serviceName)
+			if err != nil {
+				return cm, err
+			}
+
+			cm.Data[internalHostname] = cert
+		} else {
```
adambkaplan: +1 - @dmage this will also help us drop the port on the internal registry's service.

bparees:
This looks like the right direction, thanks @dmage 
Can you confirm for me that when the service-ca.crt key gets updated the CM, we will 

1) observe the change 
2) update the merged CM
3) rollout both the registry pods

(the node-ca pods will pick up the updated value from (2) automatically since they just reread it periodically)

adambkaplan:
/hold cancel

Annotation will be using `service.beta.openshift.io` for GA.

dmage:
@bparees yes

1. we will receive an event and run the entire re-sync cycle
2. it will run generatorCaConfig.Update, which will update the config map image-registry-certificates accordingly to the result of generatorCaConfig.expected
3. and we have this resource as [a dependency for the Deployment](https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/podtemplatespec.go#L215)
4. we inject into the pod template [the checksum of all dependencies](https://github.com/openshift/cluster-image-registry-operator/blob/204b55f15b9c100403f40a87615b2c581dc526cd/pkg/resource/deployment.go#L71-L79)

bparees:
> we will receive an event and run the entire re-sync cycle

if we miss the event(for example, because we were down at the time), will we also detect the change on a relist?

everything else sounds good.

dmage:
@bparees yes, we will.

dmage:
This PR removes support of `image-registry.openshift-image-registry.svc.cluster.local:5000`. Maybe we want it back, though.

bparees:
> This PR removes support of image-registry.openshift-image-registry.svc.cluster.local:5000. Maybe we want it back, though.

yeah let's not rock the boat if we don't have to.

pkg/resource/nodecadaemon.go
```diff
@@ -74,25 +64,12 @@ spec:
...
                     rm -rf /etc/docker/certs.d/$d
                 fi
             done
-            if [ -e /tmp/serviceca/service-ca.crt ]; then
-              cp -u /tmp/serviceca/service-ca.crt /etc/docker/certs.d/image-registry.openshift-image-registry.svc.cluster.local:5000
-              cp -u /tmp/serviceca/service-ca.crt /etc/docker/certs.d/${internalRegistryHostname}
-            else 
-              rm /etc/docker/certs.d/image-registry.openshift-image-registry.svc.cluster.local:5000/service-ca.crt
-              rm /etc/docker/certs.d/${internalRegistryHostname}/service-ca.crt
-            fi
```
adambkaplan: @dmage If I read the PR correctly - we don't need this any more because the ServiceCA ConfigMap is created first, and then the certificate data is copied into the registry's "global" CA keyed on the internal registry hostname.

adambkaplan:
@dmage I assume we already have logic that forces a rollout of the nodeca dameon when the ConfigMap containing the CAs is updated?

pkg/resource/nodecadaemon.go
dmage: right, but we don't put `image-registry.openshift-image-registry.svc.cluster.local:5000` to this configmap.

dmage:
@adambkaplan afaik kubelet will update pods filesystem and we do not need to restart it.

pkg/resource/nodecadaemon.go
adambkaplan: per @bparees - let's restore this for now. We can revisit in 4.2 when we hopefully are able to drop the port number.

bparees:
@dmage is correct.  the node-ca run-script just re-reads the filesystem every minute or so.  And kube will update the filesystem when the configmap changes.

adambkaplan:
@dmage so we're only worried about the registry pods, which per your previous comment is already taken care of.

Latest update adds the local service name to the mounted ConfigMap.

/lgtm

PR #276, opened 2019-05-04
Add azure platform support, with emptydir as storage backend

awesomenix:


adambkaplan:
/hold

We are _not_ going to merge anything Azure related into `master` until 4.1 code freeze is lifted.

Furthermore we intend on supporting native Azure blob storage by default.

FYI @dmage

awesomenix:
@adambkaplan unfortunately without these changes installer never completes, because it keeps waiting for success from cluster-image-registry operator.

@crawford @derekwaynecarr for guidance on next steps.

pkg/storage/storage.go
```diff
@@ -90,12 +90,14 @@ func getPlatformStorage() (imageregistryv1.ImageRegistryConfigStorage, error) {
...
-	case installConfig.Platform.Libvirt != nil:
+	switch installConfig.Platform.Name() {
+	case "libvirt":
 		cfg.EmptyDir = &imageregistryv1.ImageRegistryConfigStorageEmptyDir{}
-	case installConfig.Platform.AWS != nil:
+	case "aws":
 		cfg.S3 = &imageregistryv1.ImageRegistryConfigStorageS3{}
-	case installConfig.Platform.OpenStack != nil:
+	case "azure":
+		cfg.PVC = &imageregistryv1.ImageRegistryConfigStoragePVC{}
```
bparees: i think this needs to be emptydir, unless you are guaranteeing read/write/many PVs will be available on the Azure environment when the registry is created.  As it is you will mostly still hang the install waiting for the registry to be available because the registry won't be able to claim a PV successfully.

bparees:
@derekwaynecarr and i spoke offline..we can proceed w/ this, but please make the azure storage configuration use emptydir, i think if you use PVCs you'll find your are still blocked.

derekwaynecarr:
@awesomenix thank you for the pr.

@adambkaplan - we can accept the change if it is localized to azure only.  prior to 4.2 GA, you are free to change the storage, but we need to demonstrate and initial install flow asap.

adambkaplan:
/hold cancel

@awesomenix please switch the storage to use `EmptyDir` unless you are confident Azure has an automatic storage provisioner for RWX volumes. If no storage is available IPI installs will time out waiting for the registry to be available.

pkg/storage/storage.go
awesomenix: EmptyDir sounds good to me. Didnt realize that it requires ReadWriteMany PVC, azure default storage class creates a AzureDisk which is just ReadWriteOnce mode. 

```

Volume Plugin | ReadWriteOnce | ReadOnlyMany | ReadWriteMany
-- | -- | -- | --
AWSElasticBlockStore | ✓ | - | -
AzureFile | ✓ | ✓ | ✓
AzureDisk | ✓ | - | -

```

I think blob store matches the requirement more than AzureFile/Disk. Sending an update to use EmptyDir and change later based on requirements (also emptydir is the least risk option right now)

awesomenix:
Thanks @bparees @adambkaplan for the detailed explanations. Went with the least risky option for quick unblock and changed the storage backend to emptydir

pkg/storage/storage.go
```diff
@@ -90,12 +90,14 @@ func getPlatformStorage() (imageregistryv1.ImageRegistryConfigStorage, error) {
...
+	switch installConfig.Platform.Name() {
+	case "libvirt":
 		cfg.EmptyDir = &imageregistryv1.ImageRegistryConfigStorageEmptyDir{}
-	case installConfig.Platform.AWS != nil:
+	case "aws":
 		cfg.S3 = &imageregistryv1.ImageRegistryConfigStorageS3{}
-	case installConfig.Platform.OpenStack != nil:
+	case "azure":
+		cfg.EmptyDir = &imageregistryv1.ImageRegistryConfigStorageEmptyDir{}
+	case "openstack":
```
bparees: I think this case statement was better w/ the field nil check since I don't know what assurances we have that the platform "name" won't change in the future, but the field should certainly not be nil when we're on a given platform.

that is, i'm not sure to what degree we can consider Platform.Name() to be a stable api.

if @abhinavdahiya or @sdodson or @wking says it's ok, we can, but otherwise i'd prefer to keep it as it was.

bparees:
i'm going to proactively lgtm this to get the azure folks unblocked, but depending on the feedback from the install team, we may need set the case statement back the way it was.  It should not be an issue for 4.1 but in terms of long term safety we should use something we can rely on (which means if nothing else, those strings should be coming from constants we are getting from the installer)

/lgtm

PR #277, opened 2019-05-05
fix: change switch to use objects instead of names

awesomenix:
- change switch to use objects instead of names
- vendor new installer changes from master


Gopkg.lock
```diff
@@ -65,8 +65,8 @@
     "service/sts",
   ]
   pruneopts = ""
-  revision = "4e2e7c3c80cfb77d22990cadb2c17096d2614938"
-  version = "v1.16.35"
+  revision = "86dc5d0ebe9f30ac9e939da21c240a98519343f2"
+  version = "v1.19.24"
```
adambkaplan: @dmage If I'm not mistaken you consider a bump to `aws-sdk-go` too risky.
@bparees @derekwaynecarr do we go back to having @awesomenix run `dep ensure -update github.com/openshift/installer`?

adambkaplan: Alternative is that we pin aws-sdk-go to 1.16.35 in the constraint.

dmage: @adambkaplan touching any dependency is risky, aws-sdk-go is not special. There are also protobuf, x/crypto, x/net, k8s.io/apiextensions-apiserver, k8s.io/apimachinery, etc. It's unlikely that this will break the operator, but I'd prefer to change as less as possible right before the release.

adambkaplan:
@awesomenix my apologies for sending you down the wrong path.

Please restore what you had previously (upgrading only openshift/installer), and I'll lgtm

adambkaplan:
/lgtm

Note that the bump is only for openshift/installer (`dep ensure -update github.com/openshift/installer`)

PR #279, opened 2019-05-13
Add swift support

Fedosin:
This PR adds support of OpenStack Swift for the image registry operator.


pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,206 @@
...
+		}
+	} else {
+		opts = &gophercloud.AuthOptions{
+			IdentityEndpoint: cr.Spec.Storage.Swift.AuthURL,
+			Username:         cfg.Storage.Swift.Username,
+			Password:         cfg.Storage.Swift.Password,
+			DomainID:         cr.Spec.Storage.Swift.DomainID,
+			DomainName:       cr.Spec.Storage.Swift.Domain,
+			TenantID:         cr.Spec.Storage.Swift.TenantID,
+			TenantName:       cr.Spec.Storage.Swift.Tenant,
```
dmage: @adambkaplan you were advocating for keeping this options and not to replacing them with cloud.yaml, and I don't know your vision of their interaction when you have cloud.yaml and some fields in the custom resource, so please review this.

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,206 @@
...
+	if cfg.Storage.Swift.Cloud != nil {
+		cloud := cfg.Storage.Swift.Cloud
+		envs = append(envs,
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_AUTHURL", Value: cloud.AuthInfo.AuthURL},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_DOMAIN", Value: cloud.AuthInfo.DomainName},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_DOMAINID", Value: cloud.AuthInfo.DomainID},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_TENANT", Value: cloud.AuthInfo.ProjectName},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_TENANTID", Value: cloud.AuthInfo.ProjectID},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_USERNAME", Value: cloud.AuthInfo.Username},
+			corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_PASSWORD", Value: cloud.AuthInfo.Password},
```
dmage: You must not have secret information outside of the Kubernetes secrets. See https://github.com/openshift/cluster-image-registry-operator/blob/afd711132340f353f4f66e44e0949888b23324ac/pkg/storage/s3/s3.go#L108-L118

Fedosin: This information comes from clouds.yaml, stored in "installer-cloud-credentials" secret. We read it with GetSwiftConfig function from clusterconfig.go

dmage: These EnvVars go to the pod's spec.template.spec, i.e. they go to the deployment resource. So this code will expose passwords to users that have the cluster-reader role (they can't get secrets).

pkg/clusterconfig/clusterconfig.go
```diff
@@ -44,10 +48,19 @@ type S3 struct {
 	Region    string
 }
 
+type Swift struct {
+	// UPI crednetials
+	Username string
+	Password string
+	// IPI auth data from clouds.yaml file
+	Cloud *clientconfig.Cloud
```
adambkaplan: I'm not keen on storing the full `clouds.yaml` content here, especially if credentials are reused.
I'd rather all the data needed to create the openstack driver be added here as separate fields.

See the s3 config struct [1] - we only store what we need, and read in data provided by the cred minter for IPI.
[1] https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/clusterconfig/clusterconfig.go#L40-L45

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,223 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.Storage.Swift.Cloud == nil {
+		return map[string]string{
+			"REGISTRY_STORAGE_SWIFT_USERNAME": cfg.Storage.Swift.Username,
+			"REGISTRY_STORAGE_SWIFT_PASSWORD": cfg.Storage.Swift.Password,
+		}, nil
+	}
```
adambkaplan: I believe (and @dmage correct me if I'm wrong) we always want to publish the swift username/password in this map.

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,223 @@
...
-			ValueFrom: &corev1.EnvVarSource{
-				SecretKeyRef: &corev1.SecretKeySelector{
-					LocalObjectReference: corev1.LocalObjectReference{
-						Name: imageregistryv1.ImageRegistryPrivateConfiguration,
-					},
-					Key: "REGISTRY_STORAGE_SWIFT_PASSWORD",
-				},
-			},
-		},
-	)
```
adambkaplan: Restore this - we need to read the swift usernames and passwords from a secret.

Review @dmage 's doc on the credential flow - we want to apply the same pattern for OpenStack: https://github.com/openshift/cluster-image-registry-operator/blob/master/docs/credentials-flow.md

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,223 @@
...
+	)
+
 	return
 }
 
 func (d *driver) StorageExists(cr *imageregistryv1.Config) (bool, error) {
-	return false, nil
+	client, err := d.getSwiftClient(cr)
+	if err != nil {
+		util.UpdateCondition(cr, imageregistryv1.StorageExists, operatorapi.ConditionUnknown, err.Error(), err.Error())
```
adambkaplan: Print a summary for the condition ("Could not connect to registry storage")

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,223 @@
...
+		return false, err
+	}
+
+	_, err = containers.Get(client, cr.Spec.Storage.Swift.Container, containers.GetOpts{}).Extract()
+	if err != nil {
+		util.UpdateCondition(cr, imageregistryv1.StorageExists, operatorapi.ConditionFalse, err.Error(), err.Error())
```
adambkaplan: Ditto ("Storage does not exist")

pkg/storage/swift/swift.go
```diff
@@ -1,84 +1,223 @@
...
+		corev1.EnvVar{Name: "REGISTRY_STORAGE", Value: "swift"},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_CONTAINER", Value: d.Config.Container},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_AUTHURL", Value: d.Config.AuthURL},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_DOMAIN", Value: d.Config.Domain},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_DOMAINID", Value: d.Config.DomainID},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_TENANT", Value: d.Config.Tenant},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_TENANTID", Value: d.Config.TenantID},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_USERNAME", Value: username},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_PASSWORD", Value: password},
+		corev1.EnvVar{Name: "REGISTRY_STORAGE_SWIFT_REGION", Value: d.Config.RegionName},
```
adambkaplan: Can't pass credentials in the clear (security issue)

pkg/apis/imageregistry/v1/types.go
```diff
@@ -192,13 +192,17 @@ type ImageRegistryConfigStorageS3 struct {
...
-	DomainID  string `json:"domainID"`
-	Tenant    string `json:"tenant"`
-	TenantID  string `json:"tenantID"`
+	AuthURL    string `json:"authURL"`
+	Container  string `json:"container"`
+	Domain     string `json:"domain"`
+	DomainID   string `json:"domainID"`
+	Tenant     string `json:"tenant"`
+	TenantID   string `json:"tenantID"`
+	RegionName string `json:"region-name"`
```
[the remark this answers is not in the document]
Fedosin: Done

pkg/clusterconfig/clusterconfig.go
Fedosin: Done. I added all the required fields to the struct.

pkg/storage/swift/swift.go
Fedosin: Done

pkg/storage/swift/swift.go
Fedosin: Done

pkg/storage/swift/swift.go
Fedosin: Done

pkg/storage/swift/swift.go
Fedosin: Done

pkg/storage/swift/swift.go
Fedosin: Done

pkg/clusterconfig/clusterconfig.go
```diff
@@ -156,3 +172,74 @@ func GetGCSConfig() (*Config, error) {
...
+				return false, err
+			}
+			return true, nil
+		})
+
+		if sec == nil || pollErr != nil {
+			return nil, fmt.Errorf("unable to get cluster minted credentials %q: %v", fmt.Sprintf("%s/%s", imageregistryv1.ImageRegistryOperatorNamespace, cloudCredentialsName), pollErr)
+		}
+
+		if cloudsData, ok := sec.Data["clouds.yaml"]; ok {
```
adambkaplan: I'd add a comment here that the cloud credential minter is responsible for generating the `clouds.yaml` file and placing it in the cloud creds secret. This will be good to reference if down the line the install team/auth team changes this behavior.

pkg/storage/swift/swift.go
```diff
@@ -50,35 +139,71 @@ func (d *driver) ConfigEnv() (envs []corev1.EnvVar, err error) {
...
+	// Example of a generated name: image_registry_FHEIBGDDGBLWPXFR
+	if cr.Spec.Storage.Swift.Container == "" {
+		bytes := make([]byte, 16)
+		for i := 0; i < 16; i++ {
+			bytes[i] = byte(65 + rand.Intn(25)) // A=65 and Z=65+25
+		}
+		cr.Spec.Storage.Swift.Container = "image_registry_" + string(bytes)
+	}
+
+	fmt.Println(cr.Spec.Storage.Swift.Container)
```
adambkaplan: Use glog for log messages (github.com/golang/glog) - for debug messages our convention is to use a log level of 5 or higher:
`glog.V(5).Infof("Attempting to create swift container %s", cr.Spec.Storage.Swift.Container)`

pkg/storage/swift/swift.go
```diff
@@ -87,13 +212,34 @@ func (d *driver) RemoveStorage(cr *imageregistryv1.Config) (bool, error) {
...
+
+	return true, nil
 }
 
 func (d *driver) Volumes() ([]corev1.Volume, []corev1.VolumeMount, error) {
 	return nil, nil, nil
 }
 
 func (d *driver) CompleteConfiguration(cr *imageregistryv1.Config) error {
+	cr.Status.Storage.Swift = d.Config.DeepCopy()
```
adambkaplan: @dmage can you please weigh in on what this function is supposed to do in the storage driver's lifecycle?
Comparing the `s3`, `pvc`, and `emptyDir` implementations I see very different things going on.

pkg/storage/swift/swift.go
```diff
@@ -1,33 +1,122 @@
...
 import (
+	"fmt"
+	"math/rand"
+	"reflect"
+
 	corev1 "k8s.io/api/core/v1"
 
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack"
+	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers"
```
adambkaplan: Nit - imports should be grouped as follows (separated by empty lines):
1. golang std lib
2. 3rd party
3. k8s.io
4. openshift

We don't always adhere to this, as IDEs can't really enforce this.

pkg/clusterconfig/clusterconfig.go
```diff
@@ -16,6 +18,8 @@ import (
 
 	imageregistryv1 "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1"
 	regopclient "github.com/openshift/cluster-image-registry-operator/pkg/client"
+
+	yamlv2 "gopkg.in/yaml.v2"
```
adambkaplan: Imports nit

pkg/storage/swift/swift.go
dmage: It can be just `return nil`. When this PR is merged, I'll create a follow-up PR to remove this function from the interface, we already have all defaultings in CreateStorage.

pkg/clusterconfig/clusterconfig.go
Fedosin: done

pkg/clusterconfig/clusterconfig.go
Fedosin: done

pkg/storage/swift/swift.go
Fedosin: done

pkg/storage/swift/swift.go
Fedosin: I decided to completely remove this line - I used it for local debugging and forgot to remove.

pkg/storage/swift/swift.go
Fedosin: thanks, I deleted this line since it's not necessary.

PR #286, opened 2019-05-22
WIP - Default to EmptyDir storage

adambkaplan:
* If cloud provider is unknown, default to EmptyDir storage.
* Report registry as degraded if EmptyDir is used for storage.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -158,6 +158,10 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
 		operatorDegraded.Status = operatorapiv1.ConditionTrue
 		operatorDegraded.Message = applyError.Error()
 		operatorDegraded.Reason = e.Reason
+	} else if cr.Status.Storage.EmptyDir != nil {
+		operatorDegraded.Status = operatorapiv1.ConditionTrue
+		operatorDegraded.Message = "Registry is not using persistent storage - image data may be lost"
+		operatorDegraded.Reason = "StorageNotPersisted"
 	}
 
 	updateCondition(cr, operatorapiv1.OperatorStatusTypeDegraded, operatorDegraded)
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -107,6 +107,10 @@ func getPlatformStorage() (imageregistryv1.ImageRegistryConfigStorage, error) {
 		// to obtain system credentials. Until it's done we'll use EmptyDir for
 		// OpenStack deployment.
 		cfg.EmptyDir = &imageregistryv1.ImageRegistryConfigStorageEmptyDir{}
+	default:
+		// To faciliate unknown cloud providers, default to EmptyDir storage
+		// This will result in the operator reporting the `Degraded=true` condition
+		cfg.EmptyDir = &imageregistryv1.ImageRegistryConfigStorageEmptyDir{}
 	}
 
 	return cfg, nil
--- a/test/e2e/configuration_test.go
+++ b/test/e2e/configuration_test.go
@@ -33,10 +33,7 @@ func TestPodResourceConfiguration(t *testing.T) {
 		},
 		Spec: imageregistryapiv1.ImageRegistrySpec{
 			ManagementState: operatorapiv1.Managed,
-			Storage: imageregistryapiv1.ImageRegistryConfigStorage{
-				EmptyDir: &imageregistryapiv1.ImageRegistryConfigStorageEmptyDir{},
-			},
-			Replicas: 1,
+			Replicas:        1,
 			Resources: &corev1.ResourceRequirements{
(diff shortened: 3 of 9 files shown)


dmage:
Looks good, but in this case all other tests that use EmptyDir shouldn't define storage configuration and use default storage configuration.

abhinavdahiya:
This also will make testing UPI substantialy difficult as they specifically set storage to emptyDir for testing.
Why does the operator degrade on EmptyDir, if it was specifically told to do so???

bparees:
> This also will make testing UPI substantialy difficult as they specifically set storage to emptyDir for testing.  Why does the operator degrade on EmptyDir, if it was specifically told to do so???

A registry using emptydir storage should always be considered "degraded" as you are at risk of losing your content if the registry pod is evicted/rescheduled.  An admin needs to be aware of this.

Why does it make UPI testing difficult?

abhinavdahiya:
> > This also will make testing UPI substantialy difficult as they specifically set storage to emptyDir for testing.  Why does the operator degrade on EmptyDir, if it was specifically told to do so???
> 
> A registry using emptydir storage should always be considered "degraded" as you are at risk of losing your content if the registry pod is evicted/rescheduled. An admin needs to be aware of this.

Warnings are great! but Degraded has meaning important to CVO and cluster creation complete. So if the admin chose that as storage, operator shouldn't be Degraded. 

> Why does it make UPI testing difficult?

Because it requires ci setup storage backend for metal and vsphere.

bparees:
> Degraded has meaning important to CVO and cluster creation complete.

what meaning does it have?  available and published version are all the CVO should care about that i am aware of.  Since when does it care about Degraded and what does it do w/ that information?  Regardless, we do not want you running a cluster upgrade while using emptydir storage on your registry, you are going to make a mess of it.

> Because it requires ci setup storage backend for metal and vsphere.

that is true regardless...  on baremetal and vsphere UPI you're actually going to explicitly get *no* storage configuration at all, and the registry will not be available, and CI will need to configure storage..this requirement (post-install configuration before CI can run - essentially "day 2 admin operations") has been discussed elsewhere.

This is being done to avoid blocking IPI installs in which we cannot configure a meaningful/real storage solution(yet).  We do not want those users to fail to realize the situation they are in.

Should we distinguish between "the registry operator chose this for you because we didn't know what else to do" and "the admin configured the registry operator this way"?  Maybe.  personally i think even if you chose emptydir, you should be informed the registry is degraded because it cannot provide the level of service that is really needed.

And let me be clear on the behavior the registry is providing here:

1) IPI installs on platforms we support: storage should be configured to something "real" by the time the install is done.

2) UPI installs on cloud platforms - if usable storage credentials were supplied, storage will be configured.  if not, storage will not be configured, registry will not be available at all when install completes.

3) UPI baremetal - storage will not be configured (we have no idea what you want), registry will not be available when install completes.

4) IPI install on "unknown/unsupported" providers(ie we can't configure storage on that provider) - emptydir storage will be configured, registry is degraded.  This is being done for OpenStack for now because we don't yet support it but we do not want to block teams trying to do IPI install work on that platform.  By the end of 4.2, of course, we will, and you'll get "real" storage on OpenStack.  It is being done in general so that when 4.3 work starts and a team wants to try out an IPI install on cloud platform $FOO, the registry does not block the install.


@derekwaynecarr @smarterclayton concerns?

dmage:
@bparees the operator can't distinguish between UPI and IPI, so we will have EmptyDir on baremetal. Essentially it's almost the same as "the storage is not configured".

bparees:
> @bparees the operator can't distinguish between UPI and IPI, so we will have EmptyDir on baremetal. Essentially it's almost the same as "the storage is not configured".

@dmage the operator can distinguish between "bare metal" and "unknown cloud provider" though...  so on bare metal we can reasonably come up as "storage not configured"  (bare metal will never be IPI, only UPI).

adambkaplan:
/close

Summarizing discussion with @abhinavdahiya , @bparees , @crawford , et. al.

We don't want to set `Degraded=true` or `Upgradeable=false` if the the cluster admin or the installer explicitly set `emptyDir` as storage. Setting either of these will block upgrade for the component, which we don't want to do if someone/thing explicitly set it as such.

We could do this if we use `emptyDir` as a last-resort default. However, we won't protect against data loss when the underlying nodes are upgraded, so the primary utility of setting `Degraded=true` or `Upgradeable=false` is to ensure we fail our CI tests. Unclear if this is better than having `StorageNotConfigured`.

Teams developing IPI for new cloud providers can modify the install process to set `emptyDir` as the storage for the registry to unblock the install from completing. @wking @abhinavdahiya is this process documented?

wking:
Docs for adding install-time manifests are [here][1].  They're currently pretty sparse.

[1]: https://github.com/openshift/installer/blame/0e4119dd152586a6cfed0f8bed512ceccbdfdac3/docs/user/customization.md#L26

PR #291, opened 2019-06-06
OpenAPI schema

dmage:


adambkaplan:
@dmage @coreydaley note that we'll want to backport this to 4.1.z. Given there is a dep bump this may have to be done manually. See https://bugzilla.redhat.com/show_bug.cgi?id=1714775

PR #292, opened 2019-06-07
Revert "Temporarily switch OpenStack to EmptyDir"

Fedosin:
Now cloud-credential-operator has OpenStack support, so we can enable Swift again.

This reverts #285


Fedosin:
Just tested it locally and seems like everything works great out of the box.

PR #294, opened 2019-06-10
Cluster proxy support for image registry operator and operand

coreydaley:
https://jira.coreos.com/browse/DEVEXP-351


manifests/02-rbac.yaml
```diff
@@ -95,6 +95,13 @@ rules:
   - create
   - get
   - update
+- apiGroups:
+    - config.openshift.io
+  resources:
+    - proxies
```
bparees: this is in anticipation of the global proxy resource existing?

coreydaley: Yes

bparees:
i'd like to see this include a mechanism for configuring the proxy directly on the registry operator resource as well (w/ that value taking precedence over the global proxy configuration setting).

the registry operator should use the proxy configuration for:

1) operator talking to cloud apis (configuring storage)
2) setting the proxy values on the registry operand so that the registry itself uses the proxies for talking to storage apis + pulling content from upstream registries

coreydaley:
@bparees is that not what this does? https://github.com/openshift/cluster-image-registry-operator/pull/294/files#diff-e3147933b179a3ade1d900049f1fa55bL132

coreydaley:
@bparees nevermind, i see what you are saying,

bparees:
That covers part of what i was asking (I didn't remember we already had that config, so that's great).

still need to make sure it's used w/ the right precedence and also that the operator itself starts using it when it needs to talk externally to cloud apis.

pkg/operator/controller.go
```diff
@@ -122,6 +124,26 @@ func (c *Controller) createOrUpdateResources(cr *imageregistryv1.Config) error {
...
+	if len(proxyConfig.Spec.HTTPProxy) != 0 {
+		if err := os.Setenv("HTTP_PROXY", proxyConfig.Spec.HTTPProxy); err != nil {
+			return fmt.Errorf("unable to set HTTP_PROXY env var: %v", err)
+		}
+	}
+	if len(proxyConfig.Spec.HTTPSProxy) != 0 {
+		if err := os.Setenv("HTTPS_PROXY", proxyConfig.Spec.HTTPSProxy); err != nil {
+			return fmt.Errorf("unable to set HTTPS_PROXY env var: %v", err)
+		}
+	}
```
dmage: It won't reflect changes of the proxy config made after the operator is started.

bparees: the operator will need to watch the global proxy config resource and terminate itself (so it gets restarted) if the config changes.

dmage: Can we just reset these environment variables at the beginning of the reconciliation cycle? I'd like to avoid self-termination.

bparees: yeah i guess there's technically no reason you have to actually restart the process, just updating the env vars should be sufficient as long as the aws sdk (or other client sdks you use) don't cache/store the proxy config somewhere on initialization.

coreydaley: Couldn't I just remove the `len(proxyConfig.Spec.<proxyinfo>) != 0` and if the proxyConfig variables are empty they will be set as empty?  I believe that to be the same as resetting them, and then setting them again if the` len() != 0` is true, it just saves some code and is possibly more clear.

bparees: At a minimum you need to ensure that this sync logic is being invoked any time the global proxy config resource (or the operator's own proxy config resource) changes.  I'm not sure when this sync() function is called currently.

Always setting the env var to the current value (whether it's emptystring or a real value) is probably fine, but you should test that behavior in go to ensure Setenv doesn't no-op if you pass an emptystring or something else weird.

and lastly, per my other comment, you need to ensure that any client tools you're using are re-reading the env vars every time you invoke them.  If they are caching this value, you'll need to find a way to reset the cache, or terminate the process entirely to force a re-init.

dmage: Yes, removing `len() != 0` should be enough.

coreydaley: afaik the sync process runs whenever an event is encountered by the listers.  So as long as the lister is watching the cluster proxy config, it will run.  @dmage, can you weight in on that?

dmage: yes, right, we call `sync` on every change event from informers (exception is events for ClusterOperators).

coreydaley: @bparees does not seem to throw an error on setting an empty string: https://play.golang.org/p/PUKIQ7NRLKj

coreydaley: > Can we just reset these environment variables at the beginning of the reconciliation cycle? I'd like to avoid self-termination.

I don't like the idea of just setting the env vars inside of the pod, it seems very black box.  I think the deployment should get updated with the new proxy configuration (which shouldn't happen very often) and then the operator redeploys with the new settings.  This would allow an admin to query the deployment for the operator and see that a proxy is set and being used.

@dmage thoughts?

dmage: @bparees can we ask CVO to set these environment variables for all Deployments?

bparees: that was discussed briefly during some of the proxy design discussions but i don't think it's got a lot of support...  that's a pretty big hammer.  (And @coreydaley unless the CVO sets them, we can't do what you want)

/cc @danehans

coreydaley: @bparees are there any other operators that you know of that have already implemented this proxy behavior that I can use as a reference? Or are we the guinea pigs yet again?

bparees: i'm not familiar w/ any.

pkg/storage/s3/s3.go
```diff
@@ -2,6 +2,9 @@ package s3
 
 import (
 	"fmt"
+
+	"net/http"
```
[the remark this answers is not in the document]

pkg/storage/s3/s3.go
```diff
@@ -12,6 +15,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 
+	"golang.org/x/net/http/httpproxy"
```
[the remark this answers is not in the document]

manifests/02-rbac.yaml
```diff
@@ -96,6 +96,14 @@ rules:
   - create
   - get
   - update
+- apiGroups:
+    - config.openshift.io
```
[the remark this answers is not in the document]
coreydaley: updated in current push

coreydaley:
@dmage ptal, updated per your comments.

PR #297, opened 2019-06-11
Bump deps to kube 1.14

coreydaley:
bump deps(*)
updates due to make generate after dep bump


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.10 AS builder
+FROM registry.svc.ci.openshift.org/openshift/release:golang-1.11 AS builder
```
adambkaplan: Should be golang 1.12.

Also be sure to have a follow-up PR to update the prow config in openshift/release

coreydaley:
Created https://github.com/openshift/release/pull/4033

Dockerfile
coreydaley: Updated & created

PR #302, opened 2019-06-14
OpenStack: allow to explicitly specify keystone auth version

Fedosin:
Docker registry uses ncw/swift library to interact with Swift, which determines Auth Version from the Auth URL:
https://github.com/ncw/swift/blob/24e3012fc8a71f004a6455bce2088031d50bf2b6/auth.go#L56-L65.

But since Keystone supports autodetection, often this parameter is not present in the URL, and because of this the code above fails with 500.

This patch allows to specify auth version in the config, and defaults it to 3, which is the only supported version.

It is also important from the viewpoint of stability because that code in ncw/swift simply checks for the presence of a substring in the Auth URL. It means that with AuthURL https://zxcv1.com/ v1 authentication will be used, which is wrong.

### DIFF
--- a/pkg/storage/swift/swift.go
+++ b/pkg/storage/swift/swift.go
@@ -1,8 +1,11 @@
 package swift
 
 import (
+	"fmt"
 	"math/rand"
+	"net/url"
 	"reflect"
+	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack"
@@ -45,6 +48,8 @@ func (d *driver) getSwiftClient(cr *imageregistryv1.Config) (*gophercloud.Servic
 	d.Config.Domain = replaceEmpty(d.Config.Domain, cfg.Storage.Swift.Domain)
 	d.Config.DomainID = replaceEmpty(d.Config.DomainID, cfg.Storage.Swift.DomainID)
 	d.Config.RegionName = replaceEmpty(d.Config.RegionName, cfg.Storage.Swift.RegionName)
+	d.Config.AuthVersion = replaceEmpty(d.Config.AuthVersion, cfg.Storage.Swift.IdentityAPIVersion)
+	d.Config.AuthVersion = replaceEmpty(d.Config.AuthVersion, "3")
 
 	opts := &gophercloud.AuthOptions{
 		IdentityEndpoint: d.Config.AuthURL,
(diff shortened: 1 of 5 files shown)


pkg/storage/swift/swift.go
```diff
@@ -116,6 +116,7 @@ func (d *driver) ConfigEnv() (envs []corev1.EnvVar, err error) {
 	d.Config.Domain = replaceEmpty(d.Config.Domain, cfg.Storage.Swift.Domain)
 	d.Config.DomainID = replaceEmpty(d.Config.DomainID, cfg.Storage.Swift.DomainID)
 	d.Config.RegionName = replaceEmpty(d.Config.RegionName, cfg.Storage.Swift.RegionName)
+	d.Config.RegionName = replaceEmpty(d.Config.AuthVersion, "3")
```
trown: small typo. should be "d.Config.AuthVersion = ..."

pkg/storage/swift/swift.go
```diff
@@ -116,6 +116,7 @@ func (d *driver) ConfigEnv() (envs []corev1.EnvVar, err error) {
 	d.Config.Domain = replaceEmpty(d.Config.Domain, cfg.Storage.Swift.Domain)
 	d.Config.DomainID = replaceEmpty(d.Config.DomainID, cfg.Storage.Swift.DomainID)
 	d.Config.RegionName = replaceEmpty(d.Config.RegionName, cfg.Storage.Swift.RegionName)
+	d.Config.AuthVersion = replaceEmpty(d.Config.AuthVersion, "3")
```
dmage: Do we only support v3 for auto-configuration?

Fedosin: Technically speaking for a long time v3 is the only version available: v1 never existed publicly, and v2 was deprecated about 6 years ago, and completely removed from OpenStack at the end of 2017.
https://blueprints.launchpad.net/keystone/+spec/deprecate-v2-api
https://docs.openstack.org/keystone/stein/contributor/http-api.html
Therefore all existing OSP releases use v3 API and don't allow to switch to v2.
For this reason, I believe that making v3 the default API version is absolutely justified.

tomassedovic:
Thanks @Fedosin! I've tried this, but I couldn't get it to work. When my `clouds.yaml` says `auth_url: http://<IP>:5000/` the registry pod keeps crashlooping with:

```
panic: Swift authentication failed: HTTP Error: 404: 404 NOT FOUND
```

Running `oc describe pod` does have the AUTHVERSION environment variable set to `3`:

```
    Environment:
      REGISTRY_STORAGE:                       swift
      REGISTRY_STORAGE_SWIFT_CONTAINER:       image_registry_GLYGCKWYCROCKUPI
      REGISTRY_STORAGE_SWIFT_AUTHURL:         http://<IP>:5000/
      REGISTRY_STORAGE_SWIFT_DOMAIN:          default
      REGISTRY_STORAGE_SWIFT_DOMAINID:        
      REGISTRY_STORAGE_SWIFT_TENANT:          openshift
      REGISTRY_STORAGE_SWIFT_TENANTID:        c67337afc37b4173a591b38d129657f2
      REGISTRY_STORAGE_SWIFT_USERNAME:        <set to the key 'REGISTRY_STORAGE_SWIFT_USERNAME' in secret 'image-registry-private-configuration'>  Optional: false
      REGISTRY_STORAGE_SWIFT_PASSWORD:        <set to the key 'REGISTRY_STORAGE_SWIFT_PASSWORD' in secret 'image-registry-private-configuration'>  Optional: false
      REGISTRY_STORAGE_SWIFT_REGION:          regionOne
      REGISTRY_STORAGE_SWIFT_AUTHVERSION:     3
```

When I edit my `clouds.yaml` and change `auth_url` to: `auth_url: http://<IP>:5000/v3` and deploy with that, the registry comes up (but this patch should make it come up even without the `/v3` suffix).

tomassedovic:
This latest version did make it work without an explicit `/v3` suffix in my auth URL, thanks!

+1 from me.

PR #305, opened 2019-06-26
Block public access to S3 bucket and its objects by default

coreydaley:
Blocks public access to the S3 bucket that we create and the objects that it contains along with tests to verify
https://docs.aws.amazon.com/AmazonS3/latest/dev/access-control-block-public-access.html

### DIFF
--- a/pkg/apis/imageregistry/v1/types.go
+++ b/pkg/apis/imageregistry/v1/types.go
@@ -59,7 +59,11 @@ const (
 	// that we created has encryption enabled
 	StorageEncrypted = "StorageEncrypted"
 
-	// StorageIncompleteUploadCleanupEnabled denotes whethere or not the registry storage
+	// StoragePublicAccessBlocked denotes whether or not the registry storage medium
+	// that we created has had public access to itself and its objects blocked
+	StoragePublicAccessBlocked = "StoragePublicAccessBlocked"
+
+	// StorageIncompleteUploadCleanupEnabled denotes whether or not the registry storage
 	// medium is configured to automatically cleanup incomplete uploads
 	StorageIncompleteUploadCleanupEnabled = "StorageIncompleteUploadCleanupEnabled"
 
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -328,6 +328,31 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 		return err
 	}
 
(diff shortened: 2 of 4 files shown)


dmage:
To be consistent, we should do the same actions during upgrade for managed buckets that were created by 4.1.

pkg/apis/imageregistry/v1/types.go
```diff
@@ -59,6 +59,10 @@ const (
 	// that we created has encryption enabled
 	StorageEncrypted = "StorageEncrypted"
 
+	// StoragePublicAccessBlocked denotes where or not the registry storage medium
```
bparees: s/where/whether/

bparees:
From the doc you linked:
```
By default, new buckets and objects don't allow public access, but users can modify bucket policies or object permissions to allow public access. 
```

so why is this needed?

coreydaley:
I believe that it keeps the bucket/objects from being exposed publicly by accident due to any other access controls being set/modified.

> From the doc you linked:
> 
> ```
> By default, new buckets and objects don't allow public access, but users can modify bucket policies or object permissions to allow public access. 
> ```
> 
> so why is this needed?

pkg/apis/imageregistry/v1/types.go
coreydaley: updated in next push

coreydaley:
> To be consistent, we should do the same actions during upgrade for managed buckets that were created by 4.1.

Are you ok with doing that in a follow up pr?  I think that it requires a little bit more re-working of what should and should not be run when the sync occurs and checks that should be made first vs what can just be re-applied without causing an error.  Along with thoughts on whether or not we should change settings back to what we want them to be if the admin has changed x setting on the s3 storage themselves.

coreydaley:
@dmage pushed updates, ptal

Fixed grammar issue identified by @bparees and nil pointer issue.

pkg/storage/s3/s3.go
```diff
@@ -328,6 +328,31 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 		return err
 	}
 
+	// Block public access to the s3 bucket and it's objects by default
```
dmage: `it's`?

coreydaley: Fixed

coreydaley:
@dmage fixed grammar, ptal

PR #308, opened 2019-07-03
Refactor to get rid of clusterconfig/clusterconfig.go

coreydaley:


coreydaley:
@dmage ptal

pkg/storage/util/util.go
```diff
@@ -50,18 +53,46 @@ func UpdateCondition(cr *imageregistryv1.Config, conditionType string, status op
...
+	if err := yaml.NewYAMLOrJSONDecoder(strings.NewReader(string(cm.Data["install-config"])), 100).Decode(installConfig); err != nil {
+		return nil, fmt.Errorf("unable to decode cluster install configuration: %v", err)
+	}
+
+	return installConfig, nil
 }
 
-func GetInfrastructure(kubeconfig *rest.Config) (*configv1.Infrastructure, error) {
-	client, err := configv1client.NewForConfig(kubeconfig)
+func GetInfrastructure(listers *regopclient.Listers) (*configv1.Infrastructure, error) {
```
coreydaley: @bparees I think that the refactor of this method will help negate the issue with upgrades and PlatformStatus not being there, but ptal and let me know what you think.

coreydaley:
Pushed a fix for the test, was missing some permissions in the rbac for infrastructures

pkg/storage/util/util.go
bparees: seems reasonable to me.  I didn't study the invocations and logic super closely, but in general it seems like a reasonable abstraction to protect yourself from the vagaries of 4.1 vs 4.2 install config apis.

coreydaley: Thanks! That's all I was looking for.

pkg/storage/s3/s3.go
```diff
@@ -47,26 +57,82 @@ func NewDriver(c *imageregistryv1.ImageRegistryConfigStorageS3, kubeconfig *rest
...
+// GetConfig reads configuration for the S3 cloud platform services.
+func GetConfig(kubeconfig *rest.Config, listers *regopclient.Listers) (*S3, error) {
+	cfg := &S3{}
+
+	infra, err := util.GetInfrastructure(listers)
+	if err != nil {
+		return nil, err
+	}
+
+	if infra.Status.PlatformStatus.Type == configapiv1.AWSPlatformType {
```
coreydaley: We should already know by this point if we are using AWS/S3, should this just be
`if infra.Status.PlatformStatus != nil`

adambkaplan: I don't think there is any harm doing both:
`if infra.Status.PlatformStatus != nil && infra.Status.PlaformStatus.Type == configapiv1.AWSPlatformType {`

pkg/storage/s3/s3.go
```diff
@@ -290,7 +351,7 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 		for i := 0; i < 5000; i++ {
```
adambkaplan: Side note - probably worth adding a comment that we retry if our generated name happens to match an existing s3 bucket.

pkg/storage/util/util.go
```diff
@@ -50,18 +53,47 @@ func UpdateCondition(cr *imageregistryv1.Config, conditionType string, status op
...
-func GetClusterVersionConfig(kubeconfig *rest.Config) (*configv1.ClusterVersion, error) {
-	client, err := configv1client.NewForConfig(kubeconfig)
+// GetInfrastructure gets information about the cloud platform that the cluster is
+// installed on including the Type, Region, and other platform specific information.
+// Currently the install config is used as a backup to be compatible with upgrades
+// from 4.1 -> 4.2 when platformStatus did not exist, but should be able to be removed
+// in the future.
+func GetInfrastructure(listers *regopclient.Listers) (*configv1.Infrastructure, error) {
+	infra, err := listers.Infrastructures.Get("cluster")
 	if err != nil {
```
adambkaplan: 1) Shouldn't we check for not found here?
2) In a 4.1 -> 4.2 scenario, will we ever get to this point? Will the lister cache even populate because we need the CVO to create the Infrastructures CRD?

pkg/storage/util/util.go
```diff
@@ -50,18 +53,47 @@ func UpdateCondition(cr *imageregistryv1.Config, conditionType string, status op
...
+		if err != nil {
+			return nil, err
+		}
+		installConfig := &installer.InstallConfig{}
+		if err := yaml.NewYAMLOrJSONDecoder(strings.NewReader(string(ic.Data["install-config"])), 100).Decode(installConfig); err != nil {
+			return nil, fmt.Errorf("unable to decode cluster install configuration: %v", err)
+		}
+
+		if installConfig.Platform.AWS != nil {
+			infra.Status.PlatformStatus.AWS.Region = installConfig.Platform.AWS.Region
```
adambkaplan: Panic risk - `PlatformStatus.AWS` is a pointer

test/e2e/baremetal_test.go
```diff
@@ -4,54 +4,56 @@ import (
...
-		Data: map[string]string{
-			"install-config": `apiVersion: v1beta4`,
+		Status: configapiv1.InfrastructureStatus{
+			InfrastructureName: infrastructureConfig.Status.InfrastructureName,
+			PlatformStatus: &configapiv1.PlatformStatus{
+				Type: configapiv1.BareMetalPlatformType,
+			},
+			EtcdDiscoveryDomain:  infrastructureConfig.Status.EtcdDiscoveryDomain,
+			APIServerURL:         infrastructureConfig.Status.APIServerURL,
+			APIServerInternalURL: infrastructureConfig.Status.APIServerInternalURL,
```
adambkaplan: What purpose does this serve? Doesn't this information already exist for a baremetal cluster?

test/e2e/baremetal_test.go
```diff
@@ -4,54 +4,56 @@ import (
...
+		previousConfig, err := client.Infrastructures().Get("cluster", metav1.GetOptions{})
+		if err != nil {
+			t.Fatal(err)
+		}
+		infrastructureConfig.ResourceVersion = previousConfig.ResourceVersion
+		_, err = client.Infrastructures().Update(infrastructureConfig)
+		if err != nil {
+			panic(fmt.Errorf("unable to restore infrastructure cluster config: %s; %v", err, infrastructureConfig))
+		}
+	}()
```
adambkaplan: Ditto - does it make sense to restore information that really we shouldn't change?

pkg/storage/s3/s3.go

pkg/storage/s3/s3.go

pkg/storage/util/util.go
coreydaley: I believe that infrastructures was already there in 4.1, it's not new to 4.2, what is new is PlatformStatus, which was Platform in 4.1, which we are setting up here if it's not found.  Does that help?

pkg/storage/util/util.go
coreydaley: Fixed in next push

test/e2e/baremetal_test.go
coreydaley: Originally, this test was deleting the installConfig to replicate a baremetal install, but I was updating it to use infrastructures, which you can't delete.  And if you set the infrastructure ti baremetal, it seems to change a lot of stuff with the cluster and makes the test fail.  So this test is currently being skipped until we have an actual baremetal install to test on.

test/e2e/baremetal_test.go
coreydaley: same as above

coreydaley:
@adambkaplan updates pushed and comments have been responded to

test/e2e/baremetal_test.go
adambkaplan: +1 to skipping unless we are on baremetal - need to check if we have a baremetal e2e suite (we should)

My inclination is to follow what happens with AWS - we don't mess with the infrastructure config, just the options that are available for baremetal.

coreydaley: @adambkaplan not sure I am following your train of thought here.  Can you elaborate?

pkg/storage/s3/s3.go
```diff
@@ -47,26 +57,82 @@ func NewDriver(c *imageregistryv1.ImageRegistryConfigStorageS3, kubeconfig *rest
...
+	client, err := coreset.NewForConfig(kubeconfig)
+	if err != nil {
+		return nil, err
+	}
+
+	// Look for a user defined secret to get the AWS credentials from first
+	sec, err := listers.Secrets.Get(imageregistryv1.ImageRegistryPrivateConfigurationUser)
+	if err != nil && errors.IsNotFound(err) {
+		// Fall back to those provided by the credential minter if nothing is provided by the user
+		sec, err = client.Secrets(imageregistryv1.ImageRegistryOperatorNamespace).Get(imageregistryv1.CloudCredentialsName, metav1.GetOptions{})
```
dmage: As it's in the operator's namespace, you can use `listers.Secrets` here.

coreydaley: If i remember correctly, we specifically did not use the listers cache here because we ran into an issue with it not being up-to-date maybe? let me see if i can hunt it down.

coreydaley: @dmage Couldn't find it.  Updated to use the lister, if it passes tests it should be fine.

test/e2e/baremetal_test.go
adambkaplan: @dmage can weigh in - my understanding is that the current baremetal tests run on AWS, but we fake things out so the registry operator thinks it is on baremetal.

I confirmed with @stevekuznetsov that we now have CI cluster profiles for baremetal and vSphere, so we should update our tests so that they run on the appropriate platform, and only verify settings that are specific to the storage provider for the given platform. Separately, we should add new e2e suites (`e2e-baremetal`, `e2e-baremetal-operator`, etc.)

coreydaley: Got it.  Updated the baremetal test to only run on baremetal, and assume that it is running on baremetal.

coreydaley:
/retest
level=fatal msg="failed to initialize the cluster: Some cluster operators are still updating: authentication, console, monitoring"

coreydaley:
@adambkaplan @dmage Any other comments for this pr? Looks like all tests are passing.

dmage:
looks good to me, maybe @adambkaplan has something.

pkg/storage/util/util.go
```diff
@@ -50,18 +53,47 @@ func UpdateCondition(cr *imageregistryv1.Config, conditionType string, status op
...
+
+	if infra.Status.PlatformStatus == nil {
+		infra.Status.PlatformStatus = &configv1.PlatformStatus{
+			Type: infra.Status.Platform,
+		}
+
+		// TODO: Eventually we should be able to remove our dependency on the install config
+		// but it is needed for now since platformStatus doesn't get set on upgrade
+		// from 4.1 -> 4.2
+		ic, err := listers.InstallerSecrets.Get("cluster-config-v1")
```
adambkaplan: I can see this being not guaranteed for during a 4.1 -> 4.2 upgrade, but we should be prepared for this secret to disappear for 4.2 -> 4.3 upgrades

coreydaley: Once PlatformStatus is guaranteed to be populated, that should be fine.

PR #310, opened 2019-07-04
[WIP] rename aws credentials request

coreydaley:


dmage:
/hold

How will upgrades handle this rename? What will remove the old resource and create the new one?

coreydaley:
@dmage Not sure yet, threw this PR up to see if tests would even pass (which isn't a 100% guaranteed that it will work, but it's a start)

PR #314, opened 2019-07-07
Bug 1727595: test operator when proxy config does not exist

dmage:


dmage:
> deleting required proxies.config.openshift.io resource, named cluster, is not allowed

@bparees so, if it is required, then why we face problems with upgrades?

bparees:
> @bparees so, if it is required, then why we face problems with upgrades?

not sure what is throwing that error (admission control preventing delete?) but it sounds like it might be an incomplete check..it it ensures it can't be deleted but isn't ensuring it's created in the first place.  

I'm not sure who added that check/requirement, it would be worth getting them to complete the other half of this (ensuring it gets created on an upgrade to 4.2), but it's still probably best if the registry operator can tolerate it not being there, if only because there may be ordering issues in which the new registry code starts running before the proxy resource is created.

@danehans might know more about the state of the proxy resource being required and how/when it's supposed to get created during an upgrade from 4.1 (since it definitely didn't exist in 4.1).

PR #315, opened 2019-07-07
Bug 1727595: Warn on missing proxy config instead of error

coreydaley:
https://bugzilla.redhat.com/show_bug.cgi?id=1727595

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -131,7 +131,7 @@ func (c *Controller) sync() error {
 		proxyConfig = &configapiv1.Proxy{}
 	} else if err != nil {
 		// TODO: should we report Degraded?
-		return fmt.Errorf("unable to get cluster proxy configuration: %v", err)
+		glog.Warningf("unable to get cluster proxy configuration: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/operator/controller.go
```diff
@@ -131,7 +131,7 @@ func (c *Controller) sync() error {
 		proxyConfig = &configapiv1.Proxy{}
 	} else if err != nil {
 		// TODO: should we report Degraded?
-		return fmt.Errorf("unable to get cluster proxy configuration: %v", err)
+		glog.Warningf("unable to get cluster proxy configuration: %v", err)
```
bparees: we need to sort out why the existing "IsNotFound" check didn't catch this scenario and revisit what the right thing to do is here, but for now this hopefully gets us back to green on the release jobs.

dmage: Because it's a wrong place... The new error happens there:

https://github.com/openshift/cluster-image-registry-operator/blob/f10bd546a46c3a15db8c760be9c80295808f78d4/pkg/resource/podtemplatespec.go#L116-L119

coreydaley: Yep, just found that, i have a pr up in a second

PR #319, opened 2019-07-12
Bug 1705752: update crd

dmage:


dmage:
/hold

waiting until GCS is landed

manifests/00-crd.yaml
```diff
@@ -209,7 +620,27 @@ spec:
...
                           description: PrivateKey points to secret containing the
                             private key, provided by AWS.
+                          properties:
+                            key:
+                              description: The key of the secret to select from.  Must
+                                be a valid secret key.
+                              type: string
+                            name:
+                              description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                TODO: Add other useful fields. apiVersion, kind, uid?'
```
adambkaplan: Need to clean up our doc comments - this will show up in `oc explain`

manifests/00-crd.yaml
```diff
@@ -209,7 +620,27 @@ spec:
...
+                                must be defined
+                              type: boolean
+                          required:
+                          - key
                           type: object
+                      required:
+                      - baseURL
+                      - duration
+                      - keypairID
+                      - privateKey
```
adambkaplan: Looks like our CRD generator is interpreting everything as required. API needs to be updated to have `// +optional` in the doc comment and `omitempty` in the json tag.

Refer to upstream api doc, types may also need to be switched to pointers - https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

manifests/00-crd.yaml
dmage: This is a part of `k8s.io/api/core/v1`.

manifests/00-crd.yaml
dmage: `// +optional` is enough. I added this comment tag to the optional fields.

pkg/apis/imageregistry/v1/types.go
```diff
@@ -99,52 +99,57 @@ type ConfigList struct {
...
-	// HTTPSecret is the value needed by the registry to secure uploads, generated by default
+	// HTTPSecret is the value needed by the registry to secure uploads, generated by default.
+	// +optional
 	HTTPSecret string `json:"httpSecret"`
-
-	// Proxy defines the proxy to be used when calling master api, upstream registries, etc
+	// Proxy defines the proxy to be used when calling master api, upstream
+	// registries, etc.
+	// +optional
 	Proxy ImageRegistryConfigProxy `json:"proxy"`
```
adambkaplan: Need a pointer here so we can distinguish between "no proxy was set" vs. "don't use a proxy for the registry".

```suggestion
	Proxy *ImageRegistryConfigProxy `json:"proxy,omitempty"`
```

pkg/apis/imageregistry/v1/types.go
```diff
@@ -99,52 +99,57 @@ type ConfigList struct {
...
 	// ReadOnly indicates whether the registry instance should reject attempts
 	// to push new images or delete existing ones.
+	// +optional
 	ReadOnly bool `json:"readOnly"`
-
-	// Requests controls how many parallel requests a given registry instance will handle before queuing additional requests
+	// Requests controls how many parallel requests a given registry instance
+	// will handle before queuing additional requests.
+	// +optional
 	Requests ImageRegistryConfigRequests `json:"requests"`
```
adambkaplan: Pointer+omitempty - otherwise you'll see `requests: {}` in the YAML output.

```suggestion
	Requests *ImageRegistryConfigRequests `json:"requests,omitempty"`
```

pkg/apis/imageregistry/v1/types.go
```diff
@@ -153,16 +158,19 @@ type ImageRegistryStatus struct {
...
 }
 
 type ImageRegistryConfigProxy struct {
-	HTTP    string `json:"http"`
-	HTTPS   string `json:"https"`
+	// +optional
+	HTTP string `json:"http"`
+	// +optional
+	HTTPS string `json:"https"`
+	// +optional
```
adambkaplan: If we make `Proxy` a pointer in `spec`, then we can can keep these as "required".

I will also be adding a PR soon to have a generic Proxy with these fields (and appropriate doc for oc explain)

pkg/apis/imageregistry/v1/types.go
```diff
@@ -176,7 +184,8 @@ type ImageRegistryConfigStorageS3CloudFront struct {
 	PrivateKey corev1.SecretKeySelector `json:"privateKey"`
 	// KeypairID is key pair ID provided by AWS.
 	KeypairID string `json:"keypairID"`
-	// Duration is the duration of the Cloudfront session (optional).
+	// Duration is the duration of the Cloudfront session.
+	// +optional
 	Duration metav1.Duration `json:"duration"`
```
adambkaplan: Do we want `omitempty` - otherwise we'll get `duration: 0s` in the YAML output

pkg/apis/imageregistry/v1/types.go
```diff
@@ -187,64 +196,89 @@ type ImageRegistryConfigStorageEmptyDir struct {
 // the registry to use the AWS S3 service for backend storage
 // https://docs.docker.com/registry/storage-drivers/s3/
 type ImageRegistryConfigStorageS3 struct {
-	// Bucket is the bucket name in which you want to store the registry's data
-	// Optional, will be generated if not provided
+	// Bucket is the bucket name in which you want to store the registry's
+	// data.
+	// Optional, will be generated if not provided.
+	// +optional
 	Bucket string `json:"bucket"`
```
adambkaplan: Convert to pointer + omitempty?

pkg/apis/imageregistry/v1/types.go
```diff
@@ -187,64 +196,89 @@ type ImageRegistryConfigStorageEmptyDir struct {
...
-	// Region is the AWS region in which your bucket exists
-	// Optional, will be set based on the installed AWS Region
+	// Region is the AWS region in which your bucket exists.
+	// Optional, will be set based on the installed AWS Region.
+	// +optional
 	Region string `json:"region"`
```
adambkaplan: @openshift/api-reviewers does it make sense to have an optional field that is not `omitempty`?

pkg/apis/imageregistry/v1/types.go
```diff
@@ -187,64 +196,89 @@ type ImageRegistryConfigStorageEmptyDir struct {
...
+	// +optional
+	Domain string `json:"domain"`
+	// +optional
+	DomainID string `json:"domainID"`
+	// +optional
+	Tenant string `json:"tenant"`
+	// +optional
+	TenantID string `json:"tenantID"`
+	// +optional
+	RegionName string `json:"regionName"`
```
adambkaplan: Needs doc for `oc explain`

pkg/apis/imageregistry/v1/types.go
```diff
@@ -187,64 +196,89 @@ type ImageRegistryConfigStorageEmptyDir struct {
...
 }
 
 type ImageRegistryConfigStoragePVC struct {
+	// +optional
```
adambkaplan: Needs doc for `oc explain`

pkg/apis/imageregistry/v1/types.go
```diff
@@ -255,47 +289,54 @@ type ImageRegistryConfigStorage struct {
 	// for production use. When the pod is removed from a node for any reason,
 	// the data in the emptyDir is deleted forever.
 	// This configuration is EXPERIMENTAL and is subject to change without notice.
+	// +optional
```
adambkaplan: We need to add a discriminator field and type, i.e.

```
type ImageRegistryConfigStorageType string

const (
    AzureImageRegistryStorageType ImageRegistryConfigStorageType = "azure"
    GCSImageRegistryStorageType ImageRegistryConfigStorageType = "gcs"
    S3ImageRegistryStorageType ImageRegistryConfigStorageType = "s3"
...
)
```

And in the storage struct:
```
    Type ImageRegistryConfigStorageType `json:type`
```
See https://github.com/openshift/api/pull/300

manifests/00-crd.yaml
adambkaplan: Something that we can do upstream!

adambkaplan: (obviously won't address in this PR)

pkg/apis/imageregistry/v1/types.go
dmage: Do you want this PR to change the logic for proxies? Or should this PR depend on your PR about a generic Proxy?

pkg/apis/imageregistry/v1/types.go
dmage: Apparently we also need to change our logic to use this field. And we should do it backward compatible. Should it be a separate PR?

pkg/apis/imageregistry/v1/types.go
dmage: @adambkaplan ?

pkg/apis/imageregistry/v1/types.go
dmage: @adambkaplan ?

adambkaplan:
/close

Replaced by #331

PR #320, opened 2019-07-14
Bug 1729714: cluster-config-v1 is a config map

dmage:


coreydaley:
@dmage good catch, not sure how I messed that up ... :(

PR #322, opened 2019-07-17
Bug 1701422: Use port 443 for registry service

adambkaplan:
* Adds port 443 as the primary serving port on the registry service
* Maintain port 5000 for backwards-compatibility
* Drops port number on internal registry hostname

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -76,6 +76,7 @@ func NewController(kubeconfig *restclient.Config) (*Controller, error) {
 	p.Healthz.TimeoutSeconds = 5
 
 	p.Service.Name = imageregistryv1.ImageRegistryName
+	p.Service.Ports = []int{443, 5000}
 	p.ImageConfig.Name = "cluster"
 	p.CAConfig.Name = imageregistryv1.ImageRegistryCertificatesName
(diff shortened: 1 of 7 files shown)


adambkaplan:
/assign @dmage 

/hold
Likely blocked by openshift/origin#23408

adambkaplan:
/retest

Flakes (to be investigated).
Upgrades passing means we are keeping backwards compatibility?

dmage:
No, it doesn't mean that we haven't broken backward compatibility. #267+#312 is a good example.

adambkaplan:
/test e2e-aws-builds

/hold

Builds do a lot of pull/push, so this can capture a few things.
What I really would like is a test that does the following:

1. Starts on a pre-cluster with imagestreams imported and a new-app template deployed.
2. Upgrades to the new internal registry hostname.
3. Runs an action that utilizes an "old" pull spec (start-build, maybe?)

I'll run this manually to sketch out a test case.

adambkaplan:
New PR updating the service account controller so that each port creates an auth token: https://github.com/openshift/openshift-controller-manager/pull/7

adambkaplan:
/close

Needs rebase and will likely need significant work/testing to ensure upgrades don't break things.

PR #323, opened 2019-07-18
[openstack] Allow custom cloud names

### LINKED ISSUE #303: [openstack] should not use hard code 'openstack' cloud
/bug

https://github.com/openshift/cluster-image-registry-operator/blob/52dfda7/pkg/clusterconfig/clusterconfig.go#L194

made assumption that the 'openstack' cloud is in cloud defintion but that's not always true and lead to https://github.com/openshift/installer/pull/1858 having issue in pass CI test

adambkaplan:
Fixes #303


pkg/storage/swift/swift.go
```diff
@@ -71,7 +71,23 @@ func GetConfig(listers *regopclient.Listers) (*Swift, error) {
...
+				if !errors.IsNotFound(err) {
+					return nil, fmt.Errorf("failed to get cluster infrastructure info: %v", err)
+				}
+			}
+			if cloudInfra != nil &&
+				cloudInfra.Status.PlatformStatus != nil &&
+				cloudInfra.Status.PlatformStatus.OpenStack != nil {
+				cloudName = cloudInfra.Status.PlatformStatus.OpenStack.CloudName
+			}
+			if len(cloudName) == 0 {
```
coreydaley: Are you also trying to capture if cloudInfra.Status.PlatformStatus.OpenStack.CloudName is blank here?
Just curious.

pkg/storage/swift/swift_test.go
```diff
@@ -144,6 +147,25 @@ func handleAuthentication(t *testing.T, endpointType string) {
 	})
 }
 
+func fakeInfrastructureLister(cloudName string) configlisters.InfrastructureLister {
```
coreydaley: Can this go here?
https://github.com/openshift/cluster-image-registry-operator/tree/master/test/framework/mock/listers

Also, should some of these tests in this file really be in e2e?

pkg/storage/swift/swift.go
adambkaplan: Yes, so we can unblock the install team. If we don't have the infra object, or the platform status doesn't have the cloud name filled yet, fall back to `openstack`.

adambkaplan:
/hold

Until issues related to openshift controller manager are resolved

PR #324, opened 2019-07-23
[WIP] Add option containerManaged for Azure

dmage:


pkg/resource/generator.go
```diff
@@ -175,7 +175,7 @@ func (g *Generator) removeObsoleteRoutes(cr *imageregistryv1.Config) error {
 
 func (g *Generator) Apply(cr *imageregistryv1.Config) error {
 	err := g.syncStorage(cr)
-	if err == storage.ErrStorageNotConfigured {
+	if err == storage.ErrStorageNotConfigured || err == storage.ErrStorageMisconfigured {
```
dmage: There will be check for util.ConfigurationError

dmage:
/assign @adambkaplan 

It's still in progress, but it still worth reviewing, especially API changes.

coreydaley:
I thought that we were using https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/apis/imageregistry/v1/types.go#L157 for this purpose?

dmage:
@coreydaley on Azure we have storage accounts and storage containers. One boolean flag is not enough.

dmage:
It's a bit scary that this flag is shared between storage providers (S3, Swift, Azure). And as I can see, Azure is the first driver that allows you to reset this flag. So on S3, if the operator creates you a storage, you can't opt-out of storage management, even if you change the bucket name. @adambkaplan should we fix it?

pkg/storage/azure/azure.go
```diff
@@ -457,9 +452,9 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 					return err
 				}
 				d.Config.AccountName = accountName
+				cr.Spec.Storage.Azure = d.Config.DeepCopy()
```
adambkaplan: I think the best way for us to detect if we created the storage account is to not overwrite `spec` when we create storage. This should probably be done in a separate PR since s3 overwrites `spec` when we create storage.

dmage: This is just writing AccountName back to Spec. d.Config is a copy of cr.Spec.

dmage: And if you are saying we shouldn't touch Spec at all, it will be backward incompatible change. Can we do it?

dmage:
Let's suppose the operator created the storage account and failed to create the storage container because of a networking error. On the next iteration it sees that the storage account exists. What should be in the StorageManged field?

bparees:
> In the latter, the operator can create storage if the admin specifies a bucket that doesn't exist (or is empty), but we won't declare it "managed" and thus delete it. If the storage account or the bucket previously exist, then we shouldn't consider that storage managed by the operator. Otherwise I feel like we can get in a weird place where an admin can modify the "managed" storage outside of the operator.

I think we can always be in a situation where the admin can modify the managed storage outside of the operator.

I think the distinction should be:  if we created the storage, we delete the storage.  Whether the admin gave us the creds, or the installer gave us the creds shouldn't matter.

dmage:
We may say that we support only one container in the storage account. And if doesn't exist, always create it. On the cleanup stage we either delete the storage account with all its containers (if the account was created by the operator), or do nothing.

dmage:
@bparees the problem that on Azure we have two entities: we have storage accounts (something that must be unique within Azure) and inside the storage account there might be several containers. Containers may not exist without a storage account.

When we create a storage account, it won't have any containers, and we should always manage both: the storage account and the container. Later the administrator can change the name of the container and this new container may or may not exist. But it doesn't matter, because if someone decides to remove the image registry, the operator will delete the storage account, and as a result its containers will be gone.

On the other hand, when the storage account was already there when the operator started its work, there are two cases: when the container already exists and when it doesn't exist. Does it worth to distinguish these two cases and remember that the created container should be deleted, but the storage account should be kept? Or can we always keep the container?

bparees:
@dmage ah, that explanation was super helpful, thanks.

i think ideally we should be symmetric.  If we created it(it being an account or a container), we delete it.  If not, we don't.  Which would mean keeping track of the storage account creation and the container creation independently, unfortunately. 

Though it makes the code more complex, it makes the behavior easier for a user to reason about.

dmage:
I've moved containerManged to status.storage.azure. As a result we have different types for `status.storage` and `spec.storage`, and I need to refactor all drivers to get rid of `d.Config`.

dmage:
This PR cannot be applied anymore, there are too many changes in CreateStorage of the Azure storage driver.

PR #325, opened 2019-07-24
OpenStack: tolerate incorrect urls with double / in the path

Fedosin:
In OSP13 there is a bug, that generates the url path with double /, i.e. http://10.1.52.2:5000//v3 in clouds.yaml file.

Keystone accepts this URL, but CIRO fails to parse it. Despite the fact this bug was fixed, some customers still have it. So, we have to be tolerant for this and support such a format as well.


pkg/storage/swift/swift.go
```diff
@@ -267,6 +267,13 @@ func (d *driver) ensureAuthURLHasAPIVersion() error {
 
 	path := parsedURL.Path
 
+	// There is a bug in OSP13 that generates the url path with double /
+	// i.e. http://10.1.52.2:5000//v3
+	// We have to be tolerant for this and support such a format as well.
+	if strings.HasPrefix(path, "//") {
```
dmage: What if someone generates the URL path with triple ///, `http://10.1.52.2:5000///v3`? Or uses `http://10.1.52.2:5000/./v3`?

Fedosin: as you wish :)

PR #331, opened 2019-07-26
Bug 1705752: Update crd

adambkaplan:
Replaces #319, rebased and with a newer bump (via `dep ensure -update`).


adambkaplan:
Per conversation with @sttts - we don't _need_ a discriminator field `Type` on the storage as there are still changes needed upstream to get value out of the field.

coreydaley:
@adambkaplan Should there be a bump commit from you? I don't see one.

adambkaplan:
@coreydaley I squashed mine and Oleg's together.

adambkaplan:
(the bump got a bit messy anyway on the rebase)

PR #333, opened 2019-07-27
Refactor client creation/usage

coreydaley:
Refactor client creation/usage between the controller and the generator
to reuse already created clients instead of creating so many new ones

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -12,6 +12,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeinformers "k8s.io/client-go/informers"
 	kubeset "k8s.io/client-go/kubernetes"
+	appsset "k8s.io/client-go/kubernetes/typed/apps/v1"
+	coreset "k8s.io/client-go/kubernetes/typed/core/v1"
+	rbacset "k8s.io/client-go/kubernetes/typed/rbac/v1"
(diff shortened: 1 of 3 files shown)


coreydaley:
/retest
unrelated failures/flakes

coreydaley:
@adambkaplan bump

pkg/operator/controller.go
```diff
@@ -12,15 +12,20 @@ import (
...
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
 	operatorapi "github.com/openshift/api/operator/v1"
 	configset "github.com/openshift/client-go/config/clientset/versioned"
+	configsetv1 "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	routeset "github.com/openshift/client-go/route/clientset/versioned"
+	routesetv1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
```
adambkaplan: Why do we have unversioned and versioned ClientSets?

coreydaley: The informers needed the versioned ones, and the generators needed the unversioned ones

adambkaplan:
/lgtm

As a follow-up the generators should use versioned clientsets (if possible)

PR #334, opened 2019-07-29
Bug 1733902: Retry container creation on correct ServiceCode

coreydaley:
Looks like we were not watching for the correct service code so
that we could retry the container creation with a different name

https://bugzilla.redhat.com/show_bug.cgi?id=1733902


coreydaley:
> /approve
> 
> Can we add testing for this (unit or an e2e?)

Unfortunately there is really no way to do that.  It can't be covered by a unit test, and we need Azure credentials to test creating a container.

adambkaplan:
/lgtm

Once `e2e-azure` is working we can set up an e2e for this.

coreydaley:
/retest
unrelated failures

PR #336, opened 2019-07-29
Don't use a global variable for the s3 service

coreydaley:


pkg/storage/s3/s3.go
```diff
@@ -144,9 +136,7 @@ func (d *driver) getS3Service() (*s3.S3, error) {
 		Fn:   request.MakeAddToUserAgentHandler("openshift.io cluster-image-registry-operator", version.Version),
 	})
 
-	s3Service := s3.New(sess)
-
-	return s3Service, nil
+	return s3.New(sess), nil
```
adambkaplan: /lgtm

Why did we cache this in the first place?

PR #338, opened 2019-07-30
Add scope to gcs credentials

coreydaley:


csrwng:
Adding the scope is the right thing to do, just not sure if it's the right scope.
/lgtm

coreydaley:
If this works I can always test it with ReadWrite instead of Full.

PR #340, opened 2019-07-30
DEVEXP-403: Mount cluster trusted CA to image registry operator

adambkaplan:
Injects the cluster proxy CA into `/etc/pki/ca-trust/source/ca-bundle.crt`

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -56,6 +56,9 @@ spec:
               value: "cluster-image-registry-operator"
             - name: IMAGE
               value: docker.io/openshift/origin-docker-registry:latest
+          volumeMounts:
+            - name: trusted-ca
+              mountPath: /etc/pki/ca-trust/extracted/pem/
         - name: cluster-image-registry-operator-watch
           image: docker.io/openshift/origin-cluster-image-registry-operator:latest
           command:
@@ -81,3 +84,14 @@ spec:
               fieldRef:
                 apiVersion: v1
                 fieldPath: metadata.namespace
+          volumeMounts:
+            - name: trusted-ca
+              mountPath: /etc/pki/ca-trust/extracted/pem/
+      volumes:
+        - name: trusted-ca
+          configMap:
+            name: trusted-ca
+            optional: true
+            items:
+            - key: ca-bundle.crt
+              path: tls-ca-bundle.pem
--- a/manifests/04-ca-trusted.yaml
+++ b/manifests/04-ca-trusted.yaml
@@ -1,0 +1,10 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    release.openshift.io/create-only: "true"
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
+  name: trusted-ca
+  namespace: openshift-image-registry


manifests/04-ca-proxy.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    config.openshift.io/inject-proxy-cabundle: "true"
```
adambkaplan: I noticed in https://github.com/openshift/cluster-image-registry-operator/blob/master/manifests/04-ca-config.yaml#L5-L6 that we are not marking the other CA ConfigMap as "create only" via  the annotation `release.openshift.io/create-only: "true"`. Is this right?

manifests/07-operator.yaml
```diff
@@ -53,3 +53,10 @@ spec:
               value: "cluster-image-registry-operator"
             - name: IMAGE
               value: docker.io/openshift/origin-docker-registry:latest
+          volumeMounts:
+            - name: proxyca
+              mountPath: /etc/pki/ca-trust/source/
+      volumes:
+        - name: proxyca
+          configMap:
+            name: proxyca
```
adambkaplan: Don't block mounting the specific `ca-bundle.crt` path.

manifests/04-ca-proxy.yaml
bparees: this also needs the create-only annotation so the CVO doesn't keep stomping it:
https://github.com/openshift/cluster-config-operator/blob/master/empty-resources/0000_05_config-operator_02_apiserver.cr.yaml#L6

manifests/07-operator.yaml
```diff
@@ -53,3 +53,10 @@ spec:
               value: "cluster-image-registry-operator"
             - name: IMAGE
               value: docker.io/openshift/origin-docker-registry:latest
+          volumeMounts:
+            - name: proxyca
+              mountPath: /etc/pki/ca-trust/source/
```
bparees: pretty sure the bundle you're trying to overwrite is this one:
/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem

but i guess i'm not positive which one golang is actually reading since there is also:
/usr/share/pki/ca-trust-legacy/ca-bundle.legacy.default.crt

but it should be one of those two (there are various symlinks that point to those 2 files)

bparees:
the other bit of this should be some logic in the operator itself to watch the mounted file.  If it changes, the operator should die (so it gets restarted).

(The file on disk will change if the configmap content changes).

manifests/04-ca-proxy.yaml
bparees: It seems wrong to me, yet i don't see that CM getting stomped by the CVO.  Perhaps the CVO does not stomp configmaps back to empty?

@deads2k @abhinavdahiya ?  (tldr: we have a CM being created in the manifest for service CAs.  It's not marked create-only, but it seems to be working ok, the service-ca gets injected and the value doesn't seem to be getting stomped back out).

bparees: (after looking at a cluster i think it is getting stomped and then re-injected.  though i'm a little surprised that isn't causing rollouts of the registry operator so we should probably confirm the registry operator is actually doing the right thing when the service ca changes.  In this case the cert doesn't actually change, but it's presumably flapping back and forth between not existing and existing, in the CM)

manifests/07-operator.yaml
adambkaplan: I've called this out in the proxy design doc. Will the operator container need to run `update-ca-trust extract` before it begins?

bparees: i would not expect so, no.  pretty sure update-ca-trust extract is what produces those bundles.

manifests/04-ca-proxy.yaml
adambkaplan: I was able to dig deeper on this - there are _three_ separate control loops that are modifying the serviceca ConfigMap:

1. The CVO
2. The ca-configmap-injector
3. The image registry operator itself - https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/serviceca.go#L52-L62

Will fix this in a separate PR

adambkaplan:
> the other bit of this should be some logic in the operator itself to watch the mounted file. If it changes, the operator should die (so it gets restarted).

@ricardomaraschini you and I will work on a follow-up PR to do this.

@bparees this PR will also stage the proxy CA that will be mounted into the registry itself - I assume we can use the same ConfigMap for both as long as the volume mounts are read-only.

manifests/04-ca-proxy.yaml
bparees: good catch on (3).  seems like we should not need that generator at all?

for (1) we should just set the create-only annotation on the CM.

bparees:
> @bparees this PR will also stage the proxy CA that will be mounted into the registry itself - I assume we can use the same ConfigMap for both as long as the volume mounts are read-only.

i'm not entirely sure what you mean, but the CM that's being annotated for proxy-ca injection needs to be used only for the proxy-ca, in the same way that CM's annotated for service-ca injection are used only for that.  They can't have multiple writers.

adambkaplan:
> i'm not entirely sure what you mean, but the CM that's being annotated for proxy-ca injection needs to be used only for the proxy-ca,

My intent is that we only have one CM for the proxy ca (rename this `trusted-ca`?). This ConfigMap will be mounted into the registry operator as well as the registry pod - hence read-only mounting.

I assume we don't need to mount this into the nodeca daemonset, and that another operator will be responsible for this.

bparees:
> My intent is that we only have one CM for the proxy ca (rename this trusted-ca?). This ConfigMap will be mounted into the registry operator as well as the registry pod - hence read-only mounting.

yes mounting it in two different pods should be fine.

> I assume we don't need to mount this into the nodeca daemonset, and that another operator will be responsible for this.

would be good to confirm with @mrunalp how they are handling it(not on this thread please) but i believe the intent is for the MCO to get the trusted CAs onto the nodes and CRIO should just pick it up from there.

mrunalp:
> would be good to confirm with @mrunalp how they are handling it(not on this thread please) but i believe the intent is for the MCO to get the trusted CAs onto the nodes and CRIO should just pick it up from there.
Yes, that's how it will work for CRI-O :+1:

adambkaplan:
Huzzah! No trust = no AWS storage (though we separately trust the apiserver?):

```
 {
            "apiVersion": "config.openshift.io/v1",
            "kind": "ClusterOperator",
            "metadata": {
                "creationTimestamp": "2019-08-01T15:58:23Z",
                "generation": 1,
                "name": "image-registry",
                "resourceVersion": "12484",
                "selfLink": "/apis/config.openshift.io/v1/clusteroperators/image-registry",
                "uid": "30c468a5-b475-11e9-a066-0afad4972984"
            },
            "spec": {},
            "status": {
                "conditions": [
                    {
                        "lastTransitionTime": "2019-08-01T15:58:23Z",
                        "message": "The deployment does not exist",
                        "reason": "DeploymentNotFound",
                        "status": "False",
                        "type": "Available"
                    },
                    {
                        "lastTransitionTime": "2019-08-01T15:58:23Z",
                        "message": "Unable to apply resources: unable to sync storage configuration: RequestError: send request failed\ncaused by: Head https://image-registry-us-east-1-ciop5nrpdzv2f0b2b5wzf6-30d741eeb47511.s3.amazonaws.com/: x509: certificate signed by unknown authority",
                        "reason": "Error",
                        "status": "True",
                        "type": "Progressing"
                    },
                    {
                        "lastTransitionTime": "2019-08-01T15:58:23Z",
                        "status": "False",
                        "type": "Degraded"
                    }
                ],
                "extension": null
            }
        },
```

danehans:
Should `s3.amazonaws.com` be added to noProxy for aws installs?

adambkaplan:
> Should `s3.amazonaws.com` be added to noProxy for aws installs?

If we use [VPC endpoints for s3](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-endpoints-s3.html), then yes. Otherwise no - iirc traffic to s3 by default routes over the public internet.

Is our proxy controller smart enough to detect this?

fyi @openshift/openshift-team-network-edge

adambkaplan:
/hold

Blocked by https://github.com/openshift/cluster-network-operator/pull/274

manifests/07-operator.yaml
```diff
@@ -53,3 +53,14 @@ spec:
...
+          volumeMounts:
+            - name: trusted-ca
+              mountPath: /etc/pki/ca-trust/extracted/pem/
+      volumes:
+        - name: trusted-ca
+          configMap:
+            name: trusted-ca
+            optional: true
+            items:
+            - key: ca-bundle.crt
```
adambkaplan: @bparees this uses the optional mount and overwrites the filename.

bparees: does "optional" apply to the keys as well as the configmap itself?

adambkaplan: @bparees I can confirm it does. If the key is not present, the pod is still scheduled (and in my testing had lots of unknown CA errors).

adambkaplan:
/hold cancel

Once https://github.com/openshift/cluster-network-operator/pull/274 this should work!

danehans:
@adambkaplan https://github.com/openshift/cluster-network-operator/pull/271 is also needed, but should merge tomorrow.

danehans:
@adambkaplan https://github.com/openshift/cluster-network-operator/pull/271 has merged.

danehans:
`fail [k8s.io/kubernetes/test/e2e/framework/framework.go:338]: Aug 16 19:37:17.100: Couldn't delete ns: "e2e-test-build-webhooks-9brtm": namespace e2e-test-build-webhooks-9brtm was not deleted with limit: timed out waiting for the condition, namespace is empty but is not yet removed (&errors.errorString{s:"namespace e2e-test-build-webhooks-9brtm was not deleted with limit: timed out waiting for the condition, namespace is empty but is not yet removed"})`

/test e2e-aws

danehans:
`level=error msg="Error: Error creating S3 bucket: TooManyBuckets: You have attempted to create more buckets than allowed"
`
/test e2e-aws
/test e2e-aws-image-registry
/test e2e-aws-upgrade

PR #341, opened 2019-07-31
Bug 1711173: Handle emptying the s3 bucket before deleting it

coreydaley:


pkg/storage/s3/s3.go
```diff
@@ -140,6 +141,28 @@ func (d *driver) getS3Service() (*s3.S3, error) {
...
+			return isBucketNotFound(origErr)
+		}
+	case s3manager.Error:
+		if s3Err.OrigErr != nil {
+			return isBucketNotFound(s3Err.OrigErr)
+		}
+	case s3manager.Errors:
+		if len(s3Err) == 1 {
+			return isBucketNotFound(s3Err[0])
+		}
```
adambkaplan: Why not loop over the errors, breaking if one of the errors is `NoSuchBucket`?

pkg/storage/s3/s3.go
```diff
@@ -523,6 +547,16 @@ func (d *driver) RemoveStorage(cr *imageregistryv1.Config) (bool, error) {
...
 	}
+
+	iter := s3manager.NewDeleteListIterator(svc, &s3.ListObjectsInput{
+		Bucket: aws.String(d.Config.Bucket),
+	})
+
+	err = s3manager.NewBatchDeleteWithClient(svc).Delete(aws.BackgroundContext(), iter)
+	if err != nil && !isBucketNotFound(err) {
+		return false, err
+	}
```
adambkaplan: For below - what if we try to delete a bucket that doesn't exist? Feels like we shouldn't report an error in that case.

coreydaley: We don't report an error, we update the condition to "S3 Bucket Deleted" and return.

pkg/storage/s3/s3.go
coreydaley: They are different kinds of errors that can be passed in, so only some can be cast to awserr.Error

coreydaley: Also, this was copied from the installer.

coreydaley:
@adambkaplan responded to your comments.

PR #342, opened 2019-07-31
Implementing Proxy CA Bundle validation.

ricardomaraschini:
This patch adds a sidecar command line parameter to the registry operator. This sidecar will be deployed with the operator and will constantly monitor for changes on the mounted config map, restarting the operator every time the certificate content changes.

It is a big PR as it includes needed updates for the following packages:

1. github.com/prometheus/client_golang
2. k8s.io/apiserver
3. k8s.io/kube-openapi
4. sigs.k8s.io/structured-merge-diff


pkg/operator/controller.go
```diff
@@ -40,6 +42,9 @@ const (
 	kubeSystemNamespace      = "kube-system"
 	workqueueKey             = "changes"
 	defaultResyncDuration    = 10 * time.Minute
+	caBundlePath             = "/etc/pki/ca-trust/source"
+	caBundleFileName         = "ca-bundle.crt"
```
adambkaplan: Some churn here - the CA will now be mounted into `/etc/pki/ca-trust/extracted/tls-ca-bundle.pem`.

pkg/operator/controller.go
```diff
@@ -121,6 +128,40 @@ func (c *Controller) createOrUpdateResources(cr *imageregistryv1.Config) error {
...
 }
 
+// localCABundleHash returns the hash for the current CA bundle file.
+//
+// Opens CA bundle file and calculate the SHA256 of it, returning it as a
+// string.
+func (c *Controller) localCABundleHash() (string, error) {
+	bundleContent, err := ioutil.ReadFile(
+		fmt.Sprintf("%s/%s", caBundlePath, caBundleFileName),
+	)
```
adambkaplan: You should `os.Stat` the file first - at the moment the ca-bundle will not be a required file, and could be missing. If missing return an empty hash.

pkg/operator/controller.go
```diff
@@ -133,6 +174,17 @@ func (c *Controller) sync() error {
...
 	}
 
+	// calculate current ca bundle hash.
+	if c.caBundleHash == "" {
+		klog.Info("calculating local ca bundle hash")
+		bundleHash, err := c.localCABundleHash()
+		if err != nil {
+			return fmt.Errorf("unable to calculate local ca bundle hash: %v", err)
+		}
+		klog.Infof("local ca bundle hash: %s", bundleHash)
```
adambkaplan: Use `klog.V(6).Infof` - with klog we tend to have the following conventions
`V(0)`, `Warningf`, `Errorf` - Sparse (warnings, errors, very high level info messages)
`V(2)` - Normal info messages
`V(5)` - Debug messages
`V(6)` and higher - Trace messages

pkg/operator/controller.go
```diff
@@ -155,6 +207,13 @@ func (c *Controller) sync() error {
 	cr = cr.DeepCopy() // we don't want to change the cached version
 	prevCR := cr.DeepCopy()
 
+	klog.Info("calculating remote ca bundle hash")
```
adambkaplan: Ditto - I'd use `V(6)` here

pkg/operator/controller.go
```diff
@@ -133,6 +174,17 @@ func (c *Controller) sync() error {
 		return fmt.Errorf("unable to get cluster proxy configuration: %v", err)
 	}
 
+	// calculate current ca bundle hash.
+	if c.caBundleHash == "" {
+		klog.Info("calculating local ca bundle hash")
+		bundleHash, err := c.localCABundleHash()
```
adambkaplan: We don't want to compute this during `sync` (which is called every time the operator gets an event from the cluster) - the caBundleHash should be set once when the controller is initialized. The empty string is an acceptable value if the trust bundle hasn't been injected yet.

pkg/operator/controller.go
```diff
@@ -179,6 +238,13 @@ func (c *Controller) sync() error {
 		klog.Warningf("unknown custom resource state: %s", cr.Spec.ManagementState)
 	}
 
+	// XXX
+	// if hashes are different we need to die.
+	if c.caBundleHash != remoteBundleHash {
+		close(c.stopCh)
+		return nil
+	}
```
adambkaplan: This should be the first thing done in the sync loop (and put in its own method) so we escape early and leave the prior state intact:

```
func (c *Controller) sync() error {
   caChanged, err := c.syncTrustedCABundle()
   if err != nil {
     return err
   }
  if caChanged {
    c.Stop()
    return nil
  }
  ...
}
```

pkg/operator/controller.go
```diff
@@ -344,8 +412,9 @@ func (c *Controller) handler() cache.ResourceEventHandlerFuncs {
 	}
 }
 
-func (c *Controller) Run(stopCh <-chan struct{}) error {
+func (c *Controller) Run(stopCh chan struct{}) error {
```
adambkaplan: Why remove `<-`? I thought you could define a member like so:

```
type MyChannels struct {
  stopCh <-chan struct{}
}
```

pkg/operator/controller.go
adambkaplan: We should also log here that we've detected the CA bundle has changed (and will be shutting down)

pkg/operator/controller.go
bparees: need to discuss/confirm with @JacobTanenbaum but i think what his controller will be injecting into the configmap will be a .pem encoded file.  

What I had in my head is something that we could just mount over top of `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`

As that file is pointed to by the `/etc/pki/tls/certs/ca-bundle.crt` symlink anyway, and it's better to change the actual file in case anything is reading it from there, rather than just stomping the symlink.

we may *also* need to mount it to `/usr/share/pki/ca-trust-legacy/ca-bundle.legacy.default.crt` since that file is symlinked from `/etc/pki/ca-trust/source/ca-bundle.legacy.crt`  (ie another location some libraries may read from)

pkg/operator/controller.go
```diff
@@ -179,6 +238,13 @@ func (c *Controller) sync() error {
 		klog.Warningf("unknown custom resource state: %s", cr.Spec.ManagementState)
 	}
 
+	// XXX
+	// if hashes are different we need to die.
+	if c.caBundleHash != remoteBundleHash {
```
bparees: rather than calculating local/remote hashes you should be able to just watch the local file for changes.  if it changes, die (the operator will be restarted).  i believe some of the other operators already implement this pattern.

bparees: (you do need to ensure changes to the configmap drive a rollout of the registry operand, but there is already a framework/logic for telling this operator to "watch all these configmaps and compute a hash from them.  put that hash on the operand deployment, so that if the hash ever changes, a deployment will be triggered").  you just need to make sure this new CA configmap is part of that list.

pkg/operator/controller.go
ricardomaraschini: This was my first attempt on ending the Operator(I see that if i close that channel it stops). The fact is that i can't close a `<-chan`, only if it is `chan`, go did not allow me to close a read only channel.

pkg/operator/controller.go
ricardomaraschini: One would assume that we will have an "index" called `tls-ca-bundle.pem` containing the file content inside the config map then. Right?

pkg/operator/controller.go
ricardomaraschini: I will resolve this one, my initial approach was just a sketch and the final solution will not involve keeping track of `stopCh`.

coreydaley: I believe that you can still use close(chan) to close the channe (notice not using the <-)l, the <-just signifies that it is a read only channel

bparees:
fyi there's a canonical example of doing the "watch this file for changes and then restart the operator" here: 
https://github.com/openshift/cluster-kube-controller-manager-operator/pull/254/commits

pkg/operator/controller.go
ricardomaraschini: Well, the problem is that i was willing to keep a reference to the channel within the struct so i could close it later on. If you look at this sample below you gonna see what i mean, i needed to keep the reference as `chan` instead of `<-chan`.

https://goplay.space/#qj9_WWu4izg

ricardomaraschini: Not that a reference to a `<-chan` is not valid but if i use it i can't use `close()`. Go is so clever :-D

pkg/operator/controller.go
ricardomaraschini: I am going to solve this one as the current implementation does not contain this snippet anymore. Other than that we are not hard coding any paths on the new implementation.

pkg/operator/controller.go
ricardomaraschini: @bparees I used the canonical example you provided below to re-implement this PR. Now we have this other container running as a sidecar that monitors for file changes. I had to update a few dependencies in order to import the `watchdog` package(hence the huge PR) and I also had to use a different binary name for the `watch` process as it was, sometimes, killing itself.

pkg/operator/controller.go
ricardomaraschini: @bparees We currently have the following content on ` /etc/pki/ca-trust/extracted/pem/` inside the `cluster-image-registry-operator` container:

```
sh-4.2$ ls -1 /etc/pki/ca-trust/extracted/pem/
README
email-ca-bundle.pem
objsign-ca-bundle.pem
tls-ca-bundle.pem
sh-4.2$
```

If we mount the "trusted ca config map" on `/etc/pki/ca-trust/extracted/pem/` we will loose some of these files, right? I remember @adambkaplan mentioned something about this but I can't remember what was it.

ricardomaraschini: I guess i can resolve this comment as the original PR was completely refactored and we don't have these paths hardcoded anymore.

adambkaplan: @ricardomaraschini yes, if we mount here with no content, the files get wiped out. That's a feature per @smarterclayton - if the cluster hasn't determined who to trust, trust no one.

ricardomaraschini: I am resolving this one then :-)

adambkaplan:
@ricardomaraschini two items before this is ready for review/merge:

1. Squash your commits with the exception of the "Update dependencies" commit.
2. Rename the update dependencies commit "bump(*):", with a short note explaining why the deps need to be updated (ex: "pulling in watchdog from library-go").  Keep this commit separate from "real" code changes.

ricardomaraschini:
@adambkaplan Commits squashed/renamed as requested.

ricardomaraschini:
As for reference, the deployment of the sidecar will be made as:

```
      - image: cluster-image-registry-operator
        imagePullPolicy: IfNotPresent
        name: cluster-image-registry-operator-watch
        volumeMounts:
        - mountPath: /path/to/certificates
          name: trusted-ca
        args:
        - --namespace=$(POD_NAMESPACE)
        - --process-name=cluster-image-registry-operator
        - --termination-grace-period=30s
        - --files=/path/to/certificate.pem
        command:
        - cluster-image-registry-operator-watch
        - file-watcher-watchdog
        env:
        - name: POD_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
        - name: POD_NAMESPACE
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
```

We also need to make sure both containers share the same process namespace by specifying ` shareProcessNamespace: true` on the Deployment. This way watcher can see the operator and kill it when necessary.

Dockerfile
```diff
@@ -5,6 +5,7 @@ RUN make build
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /go/src/github.com/openshift/cluster-image-registry-operator/tmp/_output/bin/cluster-image-registry-operator /usr/bin/
+RUN ln /usr/bin/cluster-image-registry-operator /usr/bin/cluster-image-registry-operator-watch
```
adambkaplan: Use symlink: 

```suggestion
RUN ln -s /usr/bin/cluster-image-registry-operator /usr/bin/cluster-image-registry-operator-watch
```

pkg/resource/podtemplatespec.go
```diff
@@ -231,6 +231,9 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 	mounts = append(mounts, corev1.VolumeMount{Name: vol.Name, MountPath: "/etc/pki/ca-trust/source/anchors"})
 	deps.AddConfigMap(vol.VolumeSource.ConfigMap.LocalObjectReference.Name)
 
+	// Adds the user provided certificates config map as a dependency.
+	deps.AddConfigMap(v1.TrustedCAConfigMapName)
+
```
adambkaplan: You can drop this - I'm working on a PR that adds this and mounts it to the right location.

pkg/apis/imageregistry/v1/types.go
```diff
@@ -80,6 +80,11 @@ const (
 	// VersionAnnotation reflects the version of the registry that this deployment
 	// is running.
 	VersionAnnotation = "release.openshift.io/version"
+
+	// TrustedCAConfigMapName stores the name of the config where trusted certificates
+	// are stored. This allows users to provide image registry with third party
+	// certificates.
+	TrustedCAConfigMapName = "trusted-ca"
```
adambkaplan: Likewise you can drop this.

pkg/operator/controller.go
adambkaplan: @bparees I'm taking care of the "bounce the registry" logic in #349.

manifests/07-operator.yaml
```diff
@@ -55,3 +56,28 @@ spec:
...
+        - name: cluster-image-registry-operator-watch
+          image: docker.io/openshift/origin-cluster-image-registry-operator:latest
+          command:
+          - cluster-image-registry-operator-watch
+          - file-watcher-watchdog
+          args:
+          - --namespace=$(POD_NAMESPACE)
+          - --process-name=cluster-image-registry-operator
+          - --termination-grace-period=30s
+          - --files=/usr/share/pki/ca-trust-source/ca-bundle.pem
```
adambkaplan: /hold

Mounted file name may change.

adambkaplan: @ricardomaraschini change the file to `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem`, and we can merge this.

ricardomaraschini: @adambkaplan done.

wking:
Upgrade [passed][1] in CI here, but this change seems to have [broken ART upgrade tests in 4.2][2].  I've hijacked [rhbz#1727080][3] to cover this (the CVO doesn't do a good job bubbling the cause up, so this is almost certainly different from what that Bugzilla was originally about).  You may want to revert until you have time to figure out and fix whatever's wrong with the ART nightlies.

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-image-registry-operator/342/pull-ci-openshift-cluster-image-registry-operator-master-e2e-aws-upgrade/425
[2]: https://openshift-release.svc.ci.openshift.org/#4.2.0-0.nightly
[3]: https://bugzilla.redhat.com/show_bug.cgi?id=1727080

bparees:
It looks like the dockerfile change (which added the symlink) did not get sync'd to distgit.  Will need to discuss w/ ART but i'll revert this for now to get us back to green.

bparees:
@ricardomaraschini @adambkaplan this PR needed to update Dockerfile.rhel7 in addition to Dockerfile, that is why the ART release jobs (which use the .rhel7 file) were broken by this PR even though it passed CI (which, I guess, use the plain Dockerfile).

danehans:
fyi @ricardomaraschini @adambkaplan I took a slightly different approach for adding trusted ca support to ingress operator: https://github.com/openshift/cluster-ingress-operator/pull/334

/cc @bparees

bparees:
> fyi @ricardomaraschini @adambkaplan I took a slightly different approach for adding trusted ca support to ingress operator: openshift/cluster-ingress-operator#334

@danehans by which you mean you built a filewatcher into the operator process and have the operator terminate itself when the file changes (which will cause the pod to be restarted)?  That seems fine to me.

ricardomaraschini:
@danehans I like it!

PR #343, opened 2019-07-31
Bug 1734564: Remove Service CA ConfigMap from CVO management

adambkaplan:
The serviceca ConfigMap is used by operand components.
CVO should not be responsible for creating it or managing its content.

### DIFF
--- a/manifests/04-ca-config.yaml
+++ b/manifests/04-ca-config.yaml
@@ -1,8 +1,0 @@
----
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  annotations:
-    service.alpha.openshift.io/inject-cabundle: "true"
-  name: serviceca
-  namespace: openshift-image-registry
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -57,12 +57,15 @@ func TestAWSDefaults(t *testing.T) {
 
 	client := framework.MustNewClientset(t, nil)
 
+	// TODO: Move these checks to a conformance test run on all providers
 	defer framework.MustRemoveImageRegistry(t, client)
 	framework.MustDeployImageRegistry(t, client, nil)
 	framework.MustEnsureImageRegistryIsAvailable(t, client)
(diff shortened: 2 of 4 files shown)


manifests/04-ca-config.yaml
```diff
@@ -1,8 +0,0 @@
----
-apiVersion: v1
-kind: ConfigMap
```
bparees: you should check w/ @abhinavdahiya what's going to happen here when this resource vanishes from the manifest in an upgrade scenario.

seems like it would be simpler to just mark it as a create-only resource.

bparees: it's not obvious to me what we buy by having the serviceca generator handle this resource.

adambkaplan: > you should check w/ @abhinavdahiya what's going to happen here when this resource vanishes from the manifest in an upgrade scenario.

Per Abhinav the CVO will leave the object as-is, so on upgrade the existing generator code will take over managing state.

> it's not obvious to me what we buy by having the serviceca generator handle this resource.

Maybe a minor benefit - if the operator state is set to `Removed`, we can delete the ConfigMap as the registry operator does not consume the service CA.

bparees: > Per Abhinav the CVO will leave the object as-is, so on upgrade the existing generator code will take over managing state.

keep in mind you don't really want the generator doing it either.  you want the injector doing it.

now that said, the generator isn't touching the keys inside the configmap, so it's *probably* ok.

adambkaplan: We have precedent in ocm-o, where the operator creates and manages the service CA ConfigMap and object meta: https://github.com/openshift/cluster-openshift-controller-manager-operator/blob/701ad56928ab79a26b696bf3fe66510216e172dd/pkg/operator/sync_openshiftcontrollermanager_v311_00.go#L198-L228

adambkaplan:
ping @bparees

adambkaplan:
/cherry-pick release-4.1

This fix improves the stability of the internal registry.

PR #346, opened 2019-07-31
Do not hardcode GOOS/GOARCH

yselkowitz:
This change is necessary to allow building on other architectures.


bparees:
i've brought the lease failure issue to the attention of DPTP: 
https://coreos.slack.com/archives/CBN38N3MW/p1564667197431600

Issue #317, opened 2019-07-08
[openstack] include cluster ID in swift container name


ppitonak:
Installer prefixes all OpenStack resources with unique cluster ID (such as `clustername-qbmkc`), including the Swift container with ignition files.

As an OpenStack admin, it is difficult to quickly identify which cluster uses container such as `image_registry_PMJWOBLNPFHJVTGJ`, I would prefer something like `clustername-qbmkc_image_registry_PMJWOBLNPFHJVTGJ` instead.

The function generating names is here:
https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/swift/swift.go#L226

PR #347, opened 2019-08-01
OpenStack: set container metadata

Fedosin:
This commit adds metadata to the generated Swift image registry container to allow Terraform to delete it at the time the cluster is destroyed.

Also it changes the format of container naming in Swift:
<clusterID>-image-registry-<8_random_letters>, which is common for all resources created by Terraform.

Fixes: #317

### DIFF
--- a/pkg/storage/swift/swift.go
+++ b/pkg/storage/swift/swift.go
@@ -49,7 +49,7 @@ func replaceEmpty(a string, b string) string {
 	return a
 }
 
-// GetSwiftConfig reads credentials
+// GetConfig reads credentials
 func GetConfig(listers *regopclient.Listers) (*Swift, error) {
 	cfg := &Swift{}
 
@@ -328,12 +328,12 @@ func (d *driver) StorageChanged(cr *imageregistryv1.Config) bool {
 	return false
 }
 
-func generateContainerName(prefix string) string {
-	bytes := make([]byte, 16)
-	for i := 0; i < 16; i++ {
+func generateContainerName(clusterID string) string {
+	bytes := make([]byte, 8)
+	for i := 0; i < 8; i++ {
 		bytes[i] = byte(65 + rand.Intn(25)) // A=65 and Z=65+25
 	}
-	return prefix + string(bytes)
+	return clusterID + "-image-registry-" + string(bytes)
 }
 
 func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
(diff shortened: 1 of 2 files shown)


pkg/storage/swift/swift.go
```diff
@@ -328,12 +327,13 @@ func (d *driver) StorageChanged(cr *imageregistryv1.Config) bool {
 	return false
 }
 
-func generateContainerName(prefix string) string {
-	bytes := make([]byte, 16)
-	for i := 0; i < 16; i++ {
-		bytes[i] = byte(65 + rand.Intn(25)) // A=65 and Z=65+25
+func (d *driver) getClusterID() (string, error) {
```
coreydaley: There is a util.GetInfrastructure(d.Listers) that you can use for this.

coreydaley: https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/s3/s3.go#L322

Fedosin: Done!

coreydaley: You should be able to dispense with the getClusterID function altogether, and use the util.GetInfrastructure where you are using the getClusterID.  You are basically wrapping a function that does what you want inside of another function that just returns it's value?

pkg/storage/swift/swift_test.go
```diff
@@ -581,6 +582,13 @@ func TestSwiftEndpointTypeObjectStore(t *testing.T) {
 	th.AssertEquals(t, container, installConfig.Status.Storage.Swift.Container)
 }
 
-func TestSwiftGenerateContainerName(t *testing.T) {
-	th.AssertEquals(t, false, generateContainerName("image_registry_") == generateContainerName("image_registry_"))
+func TestSwiftGetClusterID(t *testing.T) {
```
coreydaley: Not sure you really need this test? The clusterID should be provided by the infrastructures config.

pkg/storage/swift/swift.go
Fedosin: you're right, this function doesn't make any sense

pkg/storage/swift/swift_test.go
Fedosin: Yep, I removed this test

pkg/storage/swift/swift.go
```diff
@@ -343,14 +343,27 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 		return err
 	}
 
+	infra, err := util.GetInfrastructure(d.Listers)
+	if err != nil {
+		return fmt.Errorf("failed to get cluster infrastructure info: %v", err)
+	}
+
+	clusterID := infra.Status.InfrastructureName
```
coreydaley: You don't need to assign to an intermediary variable, you can just use infra.Status.InfrastructureName below instead.

pkg/storage/swift/swift.go
coreydaley: You should be able to remove this method now

pkg/storage/swift/swift.go
Fedosin: done :)

pkg/storage/swift/swift.go
```diff
@@ -343,14 +343,27 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
...
+
+	createOps := containers.CreateOpts{
+		Metadata: map[string]string{
+			"Openshiftclusterid": infra.Status.InfrastructureName,
+			"Name":               cr.Spec.Storage.Swift.Container,
+		},
 	}
 
-	_, err = containers.Create(client, cr.Spec.Storage.Swift.Container, containers.CreateOpts{}).Extract()
+	_, err = containers.Create(client, cr.Spec.Storage.Swift.Container, createOps).Extract()
```
coreydaley: Can you wrap this create in a loop that will try again with a newly generated name if the name is already taken? (set a limit of some kind to keep it from looping forever)

coreydaley:
On second thought, why don't we go ahead and get this merged, then I am going to refactor the storage bucket/container name creation to all use a common function so we can be consistent, and I will add the retry at that time for swift.

PR #349, opened 2019-08-02
WIP - Mount trusted CA to the registry operand

adambkaplan:
Building on top of #340 - mounting the cluster trusted CA to the "low-priority" trust source at `/usr/share/pki/ca-trust-source/` within the registry.

The registry calls `update-ca-trust extract` to create the combined CA bundle, which includes the CA for the internal registry and other externally trusted registries.

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -55,7 +55,7 @@ spec:
               value: docker.io/openshift/origin-docker-registry:latest
           volumeMounts:
             - name: proxyca
-              mountPath: /etc/pki/ca-trust/source/
+              mountPath: /etc/pki/ca-trust/extracted/pem/
       volumes:
         - name: proxyca
           configMap:
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -53,3 +53,10 @@ spec:
               value: "cluster-image-registry-operator"
             - name: IMAGE
               value: docker.io/openshift/origin-docker-registry:latest
+          volumeMounts:
+            - name: proxyca
+              mountPath: /etc/pki/ca-trust/source/
+      volumes:
+        - name: proxyca
+          configMap:
+            name: proxyca
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -53,10 +53,10 @@ spec:
               value: "cluster-image-registry-operator"
             - name: IMAGE
               value: docker.io/openshift/origin-docker-registry:latest
-          volumeMounts:
-            - name: proxyca
-              mountPath: /etc/pki/ca-trust/extracted/pem/
-      volumes:
-        - name: proxyca
-          configMap:
-            name: proxyca
+          # volumeMounts:
+          #   - name: trusted-ca
+          #     mountPath: /etc/pki/ca-trust/extracted/pem/
+      # volumes:
+      #   - name: trusted-ca
+      #     configMap:
+      #       name: trusted-ca
--- a/pkg/resource/podtemplatespec.go
+++ b/pkg/resource/podtemplatespec.go
@@ -15,7 +15,7 @@ import (
 
 	configapiv1 "github.com/openshift/api/config/v1"
 	configlisters "github.com/openshift/client-go/config/listers/config/v1"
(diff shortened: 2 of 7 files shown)


pkg/resource/podtemplatespec.go
```diff
@@ -15,7 +15,7 @@ import (
 
 	configapiv1 "github.com/openshift/api/config/v1"
 	configlisters "github.com/openshift/client-go/config/listers/config/v1"
-	"github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1"
+	v1 "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1"
```
coreydaley: this seems redundant, your ide probably did it autonomously.

pkg/resource/podtemplatespec.go
```diff
@@ -229,7 +229,24 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
...
 	volumes = append(volumes, vol)
 	mounts = append(mounts, corev1.VolumeMount{Name: vol.Name, MountPath: "/etc/pki/ca-trust/source/anchors"})
-	deps.AddConfigMap(vol.VolumeSource.ConfigMap.LocalObjectReference.Name)
+	deps.AddConfigMap(v1.ImageRegistryCertificatesName)
+
+	// Cluster trusted certificate authorities - mount to /usr/share/pki/ca-trust-source
+	// Registry runs update-ca-trust extract on startup, which merges the registry CAs with the cluster's trusted CAs
+	// into a single CA bundle.
+	vol = corev1.Volume{
+		Name: "trusted-ca",
```
coreydaley: nit: I think you could use params.TrustedCA.Name here also, but this might be cleaner.

manifests/07-operator.yaml
```diff
@@ -53,3 +53,10 @@ spec:
               value: "cluster-image-registry-operator"
             - name: IMAGE
               value: docker.io/openshift/origin-docker-registry:latest
+          # volumeMounts:
```
coreydaley: I assume you are going to remove these and just have them commented out for testing?

ricardomaraschini:
Just as a reminder, I need this on to finish https://github.com/openshift/cluster-image-registry-operator/pull/342

pkg/resource/podtemplatespec.go
```diff
@@ -229,7 +229,24 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 	}
 	volumes = append(volumes, vol)
 	mounts = append(mounts, corev1.VolumeMount{Name: vol.Name, MountPath: "/etc/pki/ca-trust/source/anchors"})
-	deps.AddConfigMap(vol.VolumeSource.ConfigMap.LocalObjectReference.Name)
+	deps.AddConfigMap(v1.ImageRegistryCertificatesName)
```
ricardomaraschini: I guess I crossed the boundaries here, I have also implemented this on https://github.com/openshift/cluster-image-registry-operator/pull/342

manifests/07-operator.yaml
ricardomaraschini: @coreydaley Question on this one: the work being made on this PR is to mount the `trusted-ca` config map on the operand, not on the operator, right? If that is indeed the case then we gonna need these lines to be active in order to mount the config map also on operator during the deployment, right?

pkg/resource/podtemplatespec.go
adambkaplan: yep. vscode's Golang extension seems to do this if multiple packages have the same name.

pkg/resource/podtemplatespec.go
adambkaplan: will do once the ca controller lands

manifests/07-operator.yaml
adambkaplan: @coreydaley yes once the CA controller lands

@ricardomaraschini this PR is built off of #340, which will mount `trusted-ca` to the operator. I did that because I also need to mount `trusted-ca` to the registry itself (operand)

adambkaplan:
/close

Moved to #360

PR #350, opened 2019-08-02
Bug 1737053: Fixing error 409 container name exists

coreydaley:
CreateStorage was always using a static container name of "image-registry", so if you removed the container name, it would try to recreate the same container name again.  Added some randomization to the name

### DIFF
--- a/pkg/storage/azure/azure.go
+++ b/pkg/storage/azure/azure.go
@@ -18,6 +18,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/rand"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/client-go/rest"
 	"k8s.io/klog"
 
@@ -62,15 +63,6 @@ type Azure struct {
 
 	// UPI
 	AccountKey string
-}
-
-type errNameNotAvailable struct {
-	AccountName string
-	Message     string
-}
-
-func (e *errNameNotAvailable) Error() string {
-	return fmt.Sprintf("storage account name %s is not available: %s", e.AccountName, e.Message)
 }
 
 type errDoesNotExist struct {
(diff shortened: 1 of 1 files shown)


pkg/storage/azure/azure.go
```diff
@@ -430,53 +418,42 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
...
 		} else {
 			// TODO: do we need to create a storage account if we are provided with its name?
-			err = d.createStorageAccount(storageAccountsClient, cfg.ResourceGroup, d.Config.AccountName, cfg.Region)
+			result, err := d.accountExists(storageAccountsClient, d.Config.AccountName)
 			if err != nil {
-				if _, ok := err.(*errNameNotAvailable); !ok {
+				return err
+			}
+			if *result.NameAvailable {
+				if err = d.createStorageAccount(storageAccountsClient, cfg.ResourceGroup, d.Config.AccountName, cfg.Region); err != nil {
```
adambkaplan: We also need to mark that storage is managed in this situation. My understanding is that if we are creating the storage account, we are creating the underlying storage, too.

pkg/storage/azure/azure.go
```diff
@@ -486,23 +463,37 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 				return err
 			}
 
-			containerName := "image-registry"
+			for i := 0; i < 5000; i++ {
+				// If the bucket name is blank, let's generate one
+				if len(d.Config.Container) == 0 {
+					d.Config.Container = fmt.Sprintf("%s-%s-%s", infra.Status.InfrastructureName, imageregistryv1.ImageRegistryName, strings.Replace(string(uuid.NewUUID()), "-", "", -1))[0:62]
```
adambkaplan: nit - comment that the container name limit is 63 characters.

coreydaley: updated in current push

pkg/storage/azure/azure.go
coreydaley: updated in current push

coreydaley:
/retest
unrelated failures

PR #352, opened 2019-08-06
Add Name tag to S3 bucket

Fedosin:
Without this tag the resource won't be destroyed by Terraform

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -446,6 +446,10 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 						Key:   aws.String("kubernetes.io/cluster/" + infra.Status.InfrastructureName),
 						Value: aws.String("owned"),
 					},
+					{
+						Key:   aws.String("Name"),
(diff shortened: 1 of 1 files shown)


wking:
> Without this tag the resource won't be destroyed by Terraform

Who's using Terraform for tear-down (the installer uses Terraform for cluster-creation, but calls the AWS Go SDK directly for `destroy cluster`)?  I don't think we need a Name tag; although it doesn't hurt if you want to add one.  I think the bucket leak we've seen in CI recently is a race between the installer's instance and bucket teardown: openshift/installer#2169.

pkg/storage/s3/s3.go
```diff
@@ -446,6 +446,10 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 						Key:   aws.String("kubernetes.io/cluster/" + infra.Status.InfrastructureName),
 						Value: aws.String("owned"),
 					},
+					{
```
abhinavdahiya: We don't need this tag in s3 bucket for deletion.. also terraform is NOT uses for deleltion

PR #357, opened 2019-08-08
[release-4.1] Bug 1714168: Properly setting "maxWaitTime" environment…

ricardomaraschini:
Image registry expects REGISTRY_OPENSHIFT_REQUESTS_READ_MAXWAITINQUEUE
and REGISTRY_OPENSHIFT_REQUESTS_WRITE_MAXWAITINQUEUE to be of type
time.Duration. This patch renders only the Duration property of a metav1.Duration
struct.


derekwaynecarr:
Was there a corresponding change made in 4.2?
If so, where is the associated BZ that tracked that change?  Has it been verified?
The linked BZ targets 4.1, but I would have expected a dependent bug in 4.2 that was VERIFIED.  Can we create a 4.2 clone of the bug that has the fix verified first?

/hold

ricardomaraschini:
@derekwaynecarr bz for 4.2 seems to be in VERIFIED state:

https://bugzilla.redhat.com/show_bug.cgi?id=1726894

bparees:
Looks like the dependency ordering is backwards in the BZs, though the bot didn't complain(probably because it only cares *if* there is a blocking BZ against the 4.1.z BZ, then it checks that the blocking BZ is verified.  If there's no blocking BZ then it probably doesn't care.

(the 4.1.z BZ should depend on the 4.2BZ and the 4.2BZ should block the 4.1.z BZ.  The dep ordering gets set up based on which bug was cloned from which, normally.  You can edit both BZs to fix it, though again since the bot doesn't seem to be complaining i dunno how much we care)

PR #362, opened 2019-08-09
Updating name of container to inject proxy information into

coreydaley:


manifests/07-operator.yaml
```diff
@@ -4,7 +4,7 @@ kind: Deployment
 metadata:
   name: cluster-image-registry-operator
   namespace: openshift-image-registry
-  annotations:
+  labels:
```
bparees: this isn't what changed.  the configmap changed from annotation to label.  The env vars are still injected via annotation on the deployment.

manifests/07-operator.yaml
```diff
@@ -4,7 +4,7 @@ kind: Deployment
...
     config.openshift.io/inject-proxy: operator
```
bparees: but this value is wrong.  it's supposed to be the name of the container you want injected.  your container is named cluster-image-registry-operator

coreydaley: Updated.  Interesting ... Can you point me to some documentation on that?  I looked all over when I set this up and only found another operator that was doing it and theirs just said operator like this one did.

coreydaley:
@bparees ptal, updated per your comment

manifests/07-operator.yaml
bparees: @coreydaley https://github.com/openshift/cluster-version-operator/pull/224#issue-298946446

i dunno that it's documented anywhere unfortunately. 

@danehans we need to make sure teams that have added this label to their operator deployment did so correctly.  the value of the label needs to be the name(s) of the container(s) that are to get the env var injection.

PR #365, opened 2019-08-12
Removed Dockerfile.rhel7.

ricardomaraschini:
This patch migrates the content of Dockerfile.rhel7 to Dockerfile and removes the old file. As this removes a file that is used during ART we need to wait for the OK from them before we merge this.

There is a conversation on https://bugzilla.redhat.com/show_bug.cgi?id=1727080 to certify we can get rid of this file.


Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
+FROM registry.svc.ci.openshift.org/ocp/builder:golang-1.12 AS builder
```
bparees: you're going to need to talk to DPTP too, pretty sure this change is going to have unintended consequences in terms of what image is used in our CI builds if you do this.  Same for the FROM change in stage 2 below.

ricardomaraschini: After some discussions we ended up creating the following PR on ART side: https://gitlab.cee.redhat.com/openshift-art/ocp-build-data/merge_requests/207

With this patch on they wont use `Dockerfile.rhel7` anymore so i restored the original `Dockerfile` on this PR. I will check with DPTP if they were also using `Dockerfile.rhel7` before i unhold this one.

gabemontero: hey @ricardomaraschini  - unless you explicitly set the Dockefile path in this repo's config file in the openshift/release repo, the api.ci cluster will use `Dockerfile`.  Some references:
- the ci tooling that constructs the Docker Build Strategy Build in our CI has this method:  https://github.com/openshift/ci-tools/blob/master/pkg/steps/source.go#L134 .... if you look for calls of that method in the repo, you'll see it typically defaults to empty string for the dockerfilePath unless you explicitly configure it 
- the corresponding config for the image registry operator repo is at https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-image-registry-operator/openshift-cluster-image-registry-operator-master.yaml#L12-L14 ... you'll see no overrides for the dockerfile path
- the configuration reference for the ci-tools is at https://github.com/openshift/ci-tools/blob/master/CONFIGURATION.md ... here a full field example ref for `images` from that doc that corresponds to the entry for the image registry operator in my second bullet:

```
images:
- context_dir: ''
  dockerfile_path: ''
  from: ''
  inputs:
    <name>:
      as: ''
      paths:
      - destination_dir: ''
        source_path: ''
  optional: false
  to: ''

```

ricardomaraschini: Thanks @gabemontero , that really helped! I checked the openshift/release repo and i also could not find any reference to a Dockerfile so I imagined that it would use the "default" one during the process(Dockerfile).

I went ahead and asked on forum-testplatform and they told me there that we are not, as far as i understand, enforcing any specific file so I would say we are good with this one.

The changes on ART side to use Dockerfile instead of Dockerfile.rhel7 were already merged so I will remove the WIP from this one.

If you have any other concerns please let me know.

bparees:
/lgtm

thanks for cleaning this up @ricardomaraschini

PR #368, opened 2019-08-15
DEVEXP-404: Added implementation for operator metrics/alerts.

ricardomaraschini:
This commits adds a Prometheus endpoint, exposing three metrics
indicating the following:

- Wether we have storage backend configured for the registry.
- How many times we had to change the storage backend(buckets and such).
- If we are in Degraded mode.

Metric for Degraded does not consider the lack of storage backend
configuration.

Rules for alerting are also created by this commit.

### DIFF
--- a/manifests/0000_90_cluster-image-registry-operator_02_operator-servicemonitor.yaml
+++ b/manifests/0000_90_cluster-image-registry-operator_02_operator-servicemonitor.yaml
@@ -1,0 +1,17 @@
+---
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: image-registry-operator
+  namespace: openshift-image-registry
+spec:
+  endpoints:
+  - targetPort: 60000
+    interval: 60s
+    scheme: https
+    path: /metrics
+    tlsConfig:
+      caFile: /etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt
+      serverName: image-registry-operator.openshift-image-registry.svc
+  selector:
+    name: image-registry-operator
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -57,6 +57,8 @@ spec:
           volumeMounts:
             - name: trusted-ca
               mountPath: /var/run/configmaps/trusted-ca/
+            - name: image-registry-operator-tls
+              mountPath: /etc/secrets
         - name: cluster-image-registry-operator-watch
           image: docker.io/openshift/origin-cluster-image-registry-operator:latest
           command:
@@ -67,6 +69,8 @@ spec:
           - --process-name=cluster-image-registry-operator
           - --termination-grace-period=30s
           - --files=/var/run/configmaps/trusted-ca/tls-ca-bundle.pem
+          - --files=/etc/secrets/tls.crt
+          - --files=/etc/secrets/tls.key
           imagePullPolicy: IfNotPresent
           resources:
             requests:
@@ -83,6 +87,9 @@ spec:
                 apiVersion: v1
                 fieldPath: metadata.namespace
       volumes:
+        - name: image-registry-operator-tls
+          secret:
+            secretName: image-registry-operator-tls
         - name: trusted-ca
           configMap:
             name: trusted-ca
--- a/manifests/09-prometheus-rules.yaml
+++ b/manifests/09-prometheus-rules.yaml
@@ -1,0 +1,37 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    name: image-registry-operator-alerts
+  name: image-registry-operator-alerts
+  namespace: openshift-image-registry
+spec:
+  groups:
+  - name: ImageRegistryOperator
+    rules:
+    - alert: ImageRegistryMissingStorageConfiguration
+      expr: increase(image_registry_operator_status_degraded_total{reason="StorageNotConfigured"}[5m]) > 0
+      for: 48h
+      labels:
+        severity: warning
+      annotations:
+        message: |
+          Image Registry is missing a backend storage configuration. This alert
+          will auto resolve once Image Registry is provided with a valid
+          backend configuration.
+    - alert: ImageRegistryDegraded
+      expr: increase(image_registry_operator_status_degraded_total{reason!="StorageNotConfigured"}[5m]) > 0
+      for: 5m
+      labels:
+        severity: critical
+      annotations:
+        message: |
+          Image Registry could not be deployed and is in Degraded status.
+    - alert: ImageRegistryStorageReconfigured
+      expr: increase(image_registry_operator_storage_reconfigured_total[30m]) > 0
+      labels:
+        severity: info
+      annotations:
+        message: |
+          Image Registry Storage configuration has changed in the last 30
+          minutes. This change may have caused data loss.
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -1,0 +1,26 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	registry            = prometheus.NewRegistry()
+	storageReconfigured = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "storage",
+		Name:      "reconfigured_total",
+		Help:      "Total times the image registry's storage was reconfigured.",
+	})
+	degraded = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "status",
+		Name:      "degraded_total",
+		Help:      "Total times the image registry operator reported itself `Degraded`, per reason.",
+	}, []string{"reason"})
+)
+
+func init() {
+	registry.MustRegister(
+		storageReconfigured,
+		degraded,
+	)
+}
--- a/pkg/metrics/server.go
+++ b/pkg/metrics/server.go
@@ -1,0 +1,69 @@
+package metrics
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"k8s.io/klog"
+)
+
+var (
+	tlsCRT = "/etc/secrets/tls.crt"
+	tlsKey = "/etc/secrets/tls.key"
+)
+
+// RunServer starts the metrics server.
+func RunServer(port int, stopCh <-chan struct{}) {
+	if port <= 0 {
+		klog.Error("invalid port for metric server")
+		return
+	}
+
+	handler := promhttp.HandlerFor(
+		registry,
+		promhttp.HandlerOpts{
+			ErrorHandling: promhttp.HTTPErrorOnError,
+		},
+	)
+
+	bindAddr := fmt.Sprintf(":%d", port)
+	router := http.NewServeMux()
+	router.Handle("/metrics", handler)
+	srv := &http.Server{
+		Addr:    bindAddr,
+		Handler: router,
+	}
+
+	go func() {
+		err := srv.ListenAndServeTLS(tlsCRT, tlsKey)
+		if err != nil && err != http.ErrServerClosed {
+			klog.Errorf("error starting metrics server: %v", err)
+		}
+	}()
+
+	<-stopCh
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	if err := srv.Shutdown(ctx); err != nil {
+		klog.Errorf("error closing metrics server: %v", err)
+	}
+}
+
+// Degraded increases the degraded counter for provided reason.
+func Degraded(reason string) {
+	if reason == "" {
+		return
+	}
+	degraded.WithLabelValues(reason).Inc()
+}
+
+// StorageReconfigured keeps track of the number of times the operator got its
+// underlying storage reconfigured.
+func StorageReconfigured() {
+	storageReconfigured.Inc()
+}
--- a/pkg/metrics/server_test.go
+++ b/pkg/metrics/server_test.go
@@ -1,0 +1,234 @@
+package metrics
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
+	"io"
+	"io/ioutil"
+	"math/big"
+	"net/http"
+	"os"
+	"testing"
+	"time"
+
+	io_prometheus_client "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+func TestMain(m *testing.M) {
+	var err error
+
+	tlsKey, tlsCRT, err = generateTempCertificates()
+	if err != nil {
+		panic(err)
+	}
+
+	// sets the default http client to skip certificate check.
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = &tls.Config{
+		InsecureSkipVerify: true,
+	}
+
+	ch := make(chan struct{})
+	go RunServer(5000, ch)
+
+	// give http handlers/server some time to process certificates and
+	// get online before running tests.
+	time.Sleep(time.Second)
+
+	code := m.Run()
+	os.Remove(tlsKey)
+	os.Remove(tlsCRT)
+	close(ch)
+	os.Exit(code)
+}
+
+func generateTempCertificates() (string, string, error) {
+	key, err := rsa.GenerateKey(rand.Reader, 1024)
+	if err != nil {
+		return "", "", err
+	}
+
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+	}
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, key.Public(), key)
+	if err != nil {
+		return "", "", err
+	}
+
+	cert, err := ioutil.TempFile("", "testcert-")
+	if err != nil {
+		return "", "", err
+	}
+	defer cert.Close()
+	pem.Encode(cert, &pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: derBytes,
+	})
+
+	keyPath, err := ioutil.TempFile("", "testkey-")
+	if err != nil {
+		return "", "", err
+	}
+	defer keyPath.Close()
+	pem.Encode(keyPath, &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(key),
+	})
+
+	return keyPath.Name(), cert.Name(), nil
+}
+
+func TestRun(t *testing.T) {
+	resp, err := http.Get("https://localhost:5000/metrics")
+	if err != nil {
+		t.Fatalf("error requesting metrics server: %v", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		t.Errorf("expected 200, received %d instead.", resp.StatusCode)
+	}
+}
+
+func TestDegraded(t *testing.T) {
+	metricName := "image_registry_operator_status_degraded_total"
+	for _, tt := range []struct {
+		name     string
+		reason   string
+		expected float64
+		exists   bool
+	}{
+		{
+			name:     "storage not configured",
+			reason:   "StorageNotConfigured",
+			expected: 1,
+			exists:   true,
+		},
+		{
+			name:     "other reason",
+			reason:   "OtherReason",
+			expected: 1,
+			exists:   true,
+		},
+		{
+			name:     "storage not configured again",
+			reason:   "StorageNotConfigured",
+			expected: 2,
+			exists:   true,
+		},
+		{
+			name:     "other reason again",
+			reason:   "OtherReason",
+			expected: 2,
+			exists:   true,
+		},
+		{
+			name:   "empty reason",
+			reason: "",
+			exists: false,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			Degraded(tt.reason)
+
+			resp, err := http.Get("https://localhost:5000/metrics")
+			if err != nil {
+				t.Fatalf("error requesting metrics server: %v", err)
+			}
+
+			metrics := findMetricsByCounter(resp.Body, metricName)
+			if len(metrics) == 0 {
+				t.Fatal("unable to locate metric", metricName)
+			}
+
+			var counter *io_prometheus_client.Counter
+			for _, m := range metrics {
+				label := *m.Label[0].Value
+				if label != tt.reason {
+					continue
+				}
+				counter = m.Counter
+				break
+			}
+
+			if counter == nil {
+				if tt.exists {
+					t.Errorf("unable to locate counter for %q", tt.reason)
+				}
+				return
+			} else if !tt.exists {
+				t.Errorf("counter for %q should not exist", tt.reason)
+				return
+			}
+
+			val := *counter.Value
+			if val != tt.expected {
+				t.Errorf(
+					"expected %.0f, found %.0f, reason %s",
+					tt.expected, val, tt.reason,
+				)
+			}
+		})
+	}
+}
+
+func TestStorageReconfigured(t *testing.T) {
+	metricName := "image_registry_operator_storage_reconfigured_total"
+	for _, tt := range []struct {
+		name string
+		iter int
+		expt float64
+	}{
+		{
+			name: "zeroed",
+			iter: 0,
+			expt: 0,
+		},
+		{
+			name: "increase to five",
+			iter: 5,
+			expt: 5,
+		},
+		{
+			name: "increase to ten",
+			iter: 5,
+			expt: 10,
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			for i := 0; i < tt.iter; i++ {
+				StorageReconfigured()
+			}
+
+			resp, err := http.Get("https://localhost:5000/metrics")
+			if err != nil {
+				t.Fatalf("error requesting metrics server: %v", err)
+			}
+
+			metrics := findMetricsByCounter(resp.Body, metricName)
+			if len(metrics) == 0 {
+				t.Fatal("unable to locate metric", metricName)
+			}
+
+			val := *metrics[0].Counter.Value
+			if val != tt.expt {
+				t.Errorf("expected %.0f, found %.0f", tt.expt, val)
+			}
+		})
+	}
+}
+
+func findMetricsByCounter(buf io.ReadCloser, name string) []*io_prometheus_client.Metric {
+	defer buf.Close()
+	mf := io_prometheus_client.MetricFamily{}
+	decoder := expfmt.NewDecoder(buf, "text/plain")
+	for err := decoder.Decode(&mf); err == nil; err = decoder.Decode(&mf) {
+		if *mf.Name == name {
+			return mf.Metric
+		}
+	}
+	return nil
+}
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -8,6 +8,7 @@ import (
 
 	operatorapiv1 "github.com/openshift/api/operator/v1"
 	imageregistryv1 "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1"
+	"github.com/openshift/cluster-image-registry-operator/pkg/metrics"
 )
 
 func updateCondition(cr *imageregistryv1.Config, condtype string, condstate operatorapiv1.OperatorCondition) {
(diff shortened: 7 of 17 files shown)


bparees:
I assume the intent of most of these is to then create an alert that we can see in telemeter?

> How many times we had to change the storage backend(buckets and such).

what's the intent of this metric?

manifests/0000_90_cluster-image-registry-operator_02_operator-servicemonitor.yaml
```diff
@@ -0,0 +1,16 @@
...
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: image-registry-operator
+  namespace: openshift-image-registry
+spec:
+  endpoints:
+  - targetPort: 60000
+    path: /metrics
+    scheme: http
```
bparees: i realize this is a WIP so it might be on your radar already but just a heads up that this will need to be switched to https before merging.

ricardomaraschini: Thanks for the heads up @bparees . I am cutting corners to have everything in place, later on I have to revisit everything.

ricardomaraschini:
> I assume the intent of most of these is to then create an alert that we can see in telemeter?
> 
> > How many times we had to change the storage backend(buckets and such).
> 
> what's the intent of this metric?

The idea here is to have something in telemeter showing up that the storage back end got "recreated". The goal is to alert as this may, potentially, mean data loss. What do you reckon?

bparees:
> The idea here is to have something in telemeter showing up that the storage back end got "recreated". The goal is to alert as this may, potentially, mean data loss. What do you reckon?

I'm not 100% certain on our alerting philosophy, but i'm wary of firing alerts for things that the user can't take action on (would there be any way to reset this back to "it's fine"?)

pkg/operator/controller.go
```diff
@@ -120,6 +121,7 @@ func (c *Controller) createOrUpdateResources(cr *imageregistryv1.Config) error {
 
 	err := verifyResource(cr)
 	if err != nil {
+		metrics.Degraded(true)
```
dmage: It's better to calculate this value based on the degraded condition.

ricardomaraschini: I guess that then, if the storage is not configured, we will alert it as Degraded. I guess the goal was to have two distinct alerts: if it is Degrade by any other reason but not Storage Not Configured and it does not have Storage configured for more than X days.

ricardomaraschini:
> > The idea here is to have something in telemeter showing up that the storage back end got "recreated". The goal is to alert as this may, potentially, mean data loss. What do you reckon?
> 
> I'm not 100% certain on our alerting philosophy, but i'm wary of firing alerts for things that the user can't take action on (would there be any way to reset this back to "it's fine"?)

The way I am reasoning about this is:

We gonna have a time window(let's say 60 minutes), if there is any change on the metric during this time window, we create an alert. Once the change we detected is older than the time window, the alert would solve by itself.

bparees:
> We gonna have a time window(let's say 60 minutes), if there is any change on the metric during this time window, we create an alert. Once the change we detected is older than the time window, the alert would solve by itself.

ok...  not sure how you're intending to track "when" the change was made (adding a field to the registry status?), but assuming you do that I guess i can see this working...  though i'm not totally sure how useful it will be (by the time anyone notices that image pulls are failing because the storage got changed, the alert may have cleared).

ricardomaraschini:
> > We gonna have a time window(let's say 60 minutes), if there is any change on the metric during this time window, we create an alert. Once the change we detected is older than the time window, the alert would solve by itself.
> 
> ok... not sure how you're intending to track "when" the change was made (adding a field to the registry status?), but assuming you do that I guess i can see this working... though i'm not totally sure how useful it will be (by the time anyone notices that image pulls are failing because the storage got changed, the alert may have cleared).

Well, as Prometheus will be collecting constantly every time we see a "bump" on the collected values we know that there is a change with the back end. As the Alert is just a promql I am assuming i can detect these "bumps" by querying the metric where `timestamp >= now() - 60 minutes`.

bparees:
> As the Alert is just a promql I am assuming i can detect these "bumps" by querying the metric where timestamp >= now() - 60 minutes.

yeah i didn't realize you were going to report the configured storage as a metric. that approach sounds like it could work.

pkg/operator/controller.go
dmage: Conditions have Reasons, and we can use them to distinguish StorageNotConfigured from other problems.

ricardomaraschini: Ah, I see. You mean setting the Degraded counter after we know the reason for the degradation. Let me see if i can achieve this.

manifests/0000_90_cluster-image-registry-operator_02_operator-servicemonitor.yaml
ricardomaraschini: Done, metrics are now served over https.

pkg/metrics/server_test.go
```diff
@@ -0,0 +1,268 @@
...
+			}
+
+			val := *metrics[0].Gauge.Value
+			if val != tt.expt {
+				t.Errorf("expected %.0f, found %.0f", tt.expt, val)
+			}
+		})
+	}
+
+	close(ch)
```
dmage: It's better to use defer.

pkg/operator/controller.go
ricardomaraschini: Done, could you check now?

pkg/metrics/server_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	template := x509.Certificate{
+		SerialNumber: big.NewInt(1),
+	}
+	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, key.Public(), key)
+	if err != nil {
+		return "", "", err
+	}
+
+	cert, err := ioutil.TempFile("/tmp", "testcert-")
```
dmage: It's better to use `""` instead of `"/tmp"`. And as I can see, after `go test` there will be leftovers in /tmp, can we not pollute /tmp?

manifests/07-operator.yaml
```diff
@@ -21,6 +21,12 @@ spec:
       priorityClassName: system-cluster-critical
       nodeSelector:
         node-role.kubernetes.io/master: ""
+      volumes:
+      - name: image-registry-operator-tls
+        projected:
```
dmage: You don't need `projected` here.

pkg/metrics/server.go
```diff
@@ -0,0 +1,66 @@
...
+// mode or not.
+func Degraded(deg bool) {
+	if deg {
+		degraded.Set(1)
+		return
+	}
+	degraded.Set(0)
+}
+
+// IncStorageChanged increases the number of times the storage got changed.
```
dmage: The description doesn't match the current implementation.

Currently this function is called when the storage driver creates a new storage medium. Is it intentionally?

pkg/metrics/server_test.go
ricardomaraschini: Right. Migrated as proposed.

pkg/metrics/server_test.go
ricardomaraschini: I have added a fix for this.

manifests/07-operator.yaml
ricardomaraschini: I added a fix for this.

pkg/metrics/server.go
ricardomaraschini: The idea here is to take care of how many times we "create" a storage medium and report if we ended up doing it multiple times. I have renamed the functions to reflect this but I would like to hear what are your concerns.

manifests/07-operator.yaml
```diff
@@ -21,6 +21,11 @@ spec:
       priorityClassName: system-cluster-critical
       nodeSelector:
         node-role.kubernetes.io/master: ""
+      volumes:
+      - name: image-registry-operator-tls
+        secret:
+          optional: true
```
chancez: Don't forget to remove this, you'll have a bad time debugging with this if the pod starts before the secret is created and TLS isn't getting setup

ricardomaraschini:
> @ricardomaraschini - was there a reason you did not add the `openshift.io/cluster-monitoring=true` to the namesapce definition in the manifest dir.
> 
> From https://docs.google.com/document/d/1rCKAYTrYMESjJDyJ0KvNap05NNukmNXVVi6MShISnOw/edit#
> 
> ```
> To register your application’s metrics endpoint with the monitoring platform, create a ServiceMonitor object that points to your metrics endpoint. The ServiceMonitor must be in a namespace labelled with “openshift.io/cluster-monitoring=true”, due that labelling the cluster-monitoring stack will pick up the ServiceMonitor and configure Prometheus respectively. See this example.
> ```

It already has it:

```01-namespace.yaml:    openshift.io/cluster-monitoring: "true"```

**Comment by adambkaplan** [comment]
/hold

1. Waiting for 4.3 to open up
2. Needs review from telemetry folks

/cc @sichvoge

**Comment by ricardomaraschini** [comment]
> @sichvoge

Now that 4.3 season is open could we get this on? @sichvoge could you give us your OK?

**Comment by ricardomaraschini** [comment]
> /hold
> 
> 1. Waiting for 4.3 to open up
> 2. Needs review from telemetry folks
> 
> /cc @sichvoge

Now that 4.3 season is open could we get this on? @sichvoge could you give us your OK?

**Comment by bparees** [comment]
@s-urbaniak could we ask you or someone on your team (or maybe @brancz's team) to give this a review to see if we're doing metrics/alerting in a sane way?

(@sichvoge is out this week and probably not the ideal person to look at it anyway)

File: manifests/09-prometheus-rules.yaml [hunk 8]
```diff
@@ -0,0 +1,30 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  labels:
+    name: image-registry-operator-alerts
+  name: image-registry-operator-alerts
+  namespace: openshift-image-registry
+spec:
+  groups:
+  - name: ImageRegistryOperator
```
brancz: Same as here: https://github.com/openshift/cluster-samples-operator/pull/178/files#r328461466

manifests/09-prometheus-rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+    rules:
+    - alert: ImageRegistryMissingStorageConfiguration
+      expr: max_over_time(image_registry_operator_storage_configured[48h]) == 0
```
brancz: This will only fire 48h after a config is missing, is that intentional?

manifests/09-prometheus-rules.yaml
```diff
@@ -0,0 +1,30 @@
...
+          Image Registry misses a backend storage configuration. This alert
+          will auto resolve once Image Registry is provided with a valid
+          backend configuration.
+    - alert: ImageRegistryDegraded
+      expr: image_registry_operator_status_degraded == 1
+      annotations:
+        message: |
+          Image Registry could not be deployed and is in Degraded status.
+    - alert: ImageRegistryStorageBackendReconfigured
+      expr: increase(image_registry_operator_storage_backend_created[30m]) > 0
```
brancz: at some point a configuration will have to be made no? This will fire even on initial configuration

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+	degraded = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "status",
+		Name:      "degraded",
+		Help:      "Indicates if registry is degraded.",
+	})
+)
+
+func init() {
+	prometheus.MustRegister(
```
brancz: we highly recommend not to use the global metrics registry, as for example that's how we ended up having etcd and apiserver metrics in the kubernetes controller manager, which is very confusing

manifests/09-prometheus-rules.yaml
ricardomaraschini: Yes, that is the idea. As to configure a storage for image registry may be a "day 2" operation the plan is to give the administrator two days before starting to complain that there is no storage configured.

manifests/09-prometheus-rules.yaml
ricardomaraschini: I sent a new commit that set these labels but before we move on: could you clarify the `info` severity level? I read on [1] that an info alert "never fires on its own" and I could not understand exactly when to use this severity then. You mean that an "info" severity rule has no expression or has an expression that never fires?


/cc @adambkaplan

[1]  https://github.com/openshift/cluster-samples-operator/pull/178/files#r328461466

manifests/09-prometheus-rules.yaml
```diff
@@ -11,18 +11,24 @@ spec:
...
+      labels:
+        severity: warning
       annotations:
         message: |
           Image Registry misses a backend storage configuration. This alert
           will auto resolve once Image Registry is provided with a valid
           backend configuration.
     - alert: ImageRegistryDegraded
       expr: image_registry_operator_status_degraded == 1
+      labels:
```
brancz: this should have a for clause, a single sample with this value should not cause an alert

manifests/09-prometheus-rules.yaml
brancz: Sorry, yes I should have been more clear. Info alerts are no different in their definition other than the `severity: info` label, they just never cause a notification on their own (due to the way that Alertmanager is configured), only if in combination with a warning or critical alert in the same group.

brancz: The idea is that they are largely informational, and don't necessarily have meaning on their own, but could be a first hint to investigate if something is truly on fire.

pkg/metrics/metrics.go
ricardomaraschini: Could you point me to an already existing implementation to guide me here?

manifests/09-prometheus-rules.yaml
ricardomaraschini: OK, I get it. Thanks for the explanation.

manifests/09-prometheus-rules.yaml
ricardomaraschini: This is a very good point indeed. Thanks for catching this.

pkg/metrics/metrics.go
brancz: The cluster-monitoring-operator does this. It instantiates and registers the default collectors here: https://github.com/openshift/cluster-monitoring-operator/blob/9dbc613664ffc99f5a82d82a05cd145dedcb863c/cmd/operator/main.go#L114-L118

And then the registry is passed around wherever needed, for example the operator then registers its metrics and collectors here: https://github.com/openshift/cluster-monitoring-operator/blob/9dbc613664ffc99f5a82d82a05cd145dedcb863c/cmd/operator/main.go#L132

It's a good idea to pass the registry around in a similar way you would with a logger instance.

ricardomaraschini: This is an excellent idea. As the code is organised now(we are centralizing all metrics into one single package) it is not entirely necessary but as the code grows this most likely will come handy. I sent a commit that address this, ptal.

manifests/09-prometheus-rules.yaml
ricardomaraschini: @adambkaplan 

I have set the following severities:

1. ImageRegistryMissingStorageConfiguration as **WARNING**
2. ImageRegistryDegraded as **CRITICAL**
3. ImageRegistryStorageBackendReconfigured as **WARNING**

Do you think these "values" are adequate?

dmage: I'd say ImageRegistryStorageBackendReconfigured should have `severity: info`. As an administrator I don't want to receive notifications about actions that I've just made.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,33 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	registry          = prometheus.NewRegistry()
```
brancz: I recommend to not introduce yet another global registry, but rather pass it around to where it's needed

ricardomaraschini: What do you mean? This is not global nor exported, it is part of a package and it is impossible to use it anywhere else. It is declared where it is used and there is no reason to pass it around. Could you clarify?

brancz: it's still package global, which is a bit better but function scoped is better in my opinion :)

ricardomaraschini: I see. Correct me if I am wrong but you are advocating in favor of something closer to a dependency injection approach, where I would inject this "registry" in other entities(packages or structs) that then would register their own "counters" and "metrics" towards it. It has its charm i admit, and that was my initial approach. That being said, I advocate for a less intrusive approach: make metrics centralized and accessible through exported functions. :)

brancz: it ultimately comes down to preference, we've had bad experiences with anything other than dependency injection (it also allows you to use the [label/name-prefix currying](https://github.com/prometheus/client_golang/blob/20428fa0bffcb4ee7a7941545ae0df9acdc9d8a7/prometheus/wrap.go#L25-L47))

manifests/09-prometheus-rules.yaml
```diff
@@ -0,0 +1,37 @@
...
+  groups:
+  - name: ImageRegistryOperator
+    rules:
+    - alert: ImageRegistryMissingStorageConfiguration
+      expr: max_over_time(image_registry_operator_storage_configured[48h]) == 0
+      labels:
+        severity: warning
+      annotations:
+        message: |
+          Image Registry misses a backend storage configuration. This alert
```
bparees:
```suggestion
          Image Registry is missing a backend storage configuration. This alert
```

manifests/09-prometheus-rules.yaml
```diff
@@ -0,0 +1,37 @@
...
+      for: 5m
+      labels:
+        severity: critical
+      annotations:
+        message: |
+          Image Registry could not be deployed and is in Degraded status.
+    - alert: ImageRegistryStorageBackendReconfigured
+      expr: increase(image_registry_operator_storage_backend_created[30m]) > 0
+      labels:
+        severity: warning
```
bparees: agree with @dmage that this should be Info.

ricardomaraschini: Migrated it to `info`.

manifests/09-prometheus-rules.yaml
ricardomaraschini: @dmage Migrated it to `info`.

pkg/operator/controller.go
ricardomaraschini: Hey @dmage , i am resolving this comment as I have already implemented the requested changes and you haven't yet provided any other input.

pkg/metrics/server.go
ricardomaraschini: @dmage I've already implemented the requested changes and due to inactivity I am resolving this comment.

manifests/09-prometheus-rules.yaml
ricardomaraschini: Resolving this one as the changes were already implemented.

manifests/09-prometheus-rules.yaml
ricardomaraschini: I haven't seen this issue, I guess that is because Prometheus generally takes some time to actually query the metrics endpoint. I have refactored the code to support this scenario. Thanks for catching this, PTAL.

pkg/resource/generator.go
```diff
@@ -99,13 +100,54 @@ func (g *Generator) syncStorage(cr *imageregistryv1.Config) error {
...
+// configuration is extracted from Status while current comes from Spec. If
+// storage is not configured false is returned.
+func (g *Generator) storageReconfigured(cr *imageregistryv1.Config) bool {
+	prev := cr.Status.Storage
+	cur := cr.Spec.Storage
+
+	// If we cannot find storage configuration on our Status it means we
+	// are creating the storage for the first time. On this case we simply
+	// return false.
+	prevStorageID, err := storage.ID(prev)
```
dmage: I'm concerned about the switches inside `ID` and `Engine`, someone may forget to update it.

We can use newDriver here and add `ID() string` to our `Driver` interface. If types are different (`reflect.TypeOf(prevDriver) != reflect.TypeOf(curDriver)`) then we know for sure that the storage is reconfigured. If types are the same, we can use `prevDriver.ID() != curDriver.ID()`.

@ricardomaraschini what do you think?

ricardomaraschini: That would work if we first solve #394 Chicken and egg now.

ricardomaraschini: It is implemented. PTAL.

ricardomaraschini:
/hold

We need https://github.com/openshift/cluster-image-registry-operator/pull/394 to be merged first otherwise we may not report back end storage re-configurations due to the bug it fixes.

ricardomaraschini:
/hold cancel

Ready to go.

pkg/operator/status.go
```diff
@@ -44,6 +45,26 @@ func updateCondition(cr *imageregistryv1.Config, condtype string, condstate oper
...
+	if condstate.Status == operatorapiv1.ConditionFalse {
+		metrics.Degraded(false)
+		return
+	}
+	// StorageNotConfigured is not considered to be notified through
+	// Degraded metric, it has its own indication.
+	if condstate.Reason == "StorageNotConfigured" {
+		metrics.Degraded(false)
+		return
+	}
```
adambkaplan: This feels problematic, though I understand the rationale.

I think a better solution is to include a reason label for the degraded metric. We can then differentiate between `StorageNotConfigured` and other degraded reasons.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,33 @@
...
+	storageBackendReconfigured = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "storage_backend",
+		Name:      "reconfigured",
+		Help:      "Number of times registry storage backend got reconfigured.",
+	})
+	degraded = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "status",
+		Name:      "degraded",
```
adambkaplan: Perhaps we should include a `reason` label as a part of this metric.

The we can be more intelligent about the alerting:

1. Warning if degraded with `reason=StorageNotConfigured` for 48h
2. Critical otherwise

We may also be able to remove the gauge that merely indicates if storage has been configured.

pkg/operator/status.go
ricardomaraschini: The problem with the labels is that "Reasons" are mutually exclusive, in a given moment of time `T` only _one or none_ of the labels must be set to `true`. This premise tells that when the Operator is setting itself as degraded by reason `Y` it will have to clear reasons(labels) for `W` and `Z` as the operator could  previously be degraded by any of the latter. The same is valid when leaving the Degraded state, all Reasons should have their values cleared(set to `false`). That involves to know all reasons for Degraded and to clean all of others when setting one of them to `true`. Here `true` and `false` are represented by a `1` or a `0` on a Gauge metric.

Another possible solution considered was to use a Counter instead of a Gauge. That would make the alert dependent on the metric's `increase` instead of the metric's `current value`, on this case we could generate the alert using a query similar to:

```
increase(image_registry_operator_degraded{reason!="StorageNotConfigured"}[5m]) > 0
```

That would alert if there was an increase on any "Reason" different than `StorageNotConfigured` on the last 5 minutes. The problem with this approach is that it is tightly coupled with events coming in on the Operator, if there is no event there is no increase and eventually the alert could solve itself(even though still Degraded). This solution will also create a "constantly increasing" Degraded metric and that might seem counter-intuitive for someone looking at it: "Is it switching back and forth from and to Degraded at this ratio? What is going on?". Other than that such a metric AFAIK would not be feasible to properly measure the 48 hours of "StorageNotConfigured": the alert for this reason behaves differently from all the others as we should be able to create a promql that would alert only if the observed increase for Reason "StorageNotConfigured" is happening for 48 consecutive hours, if it stopped to increase on the last, lets say, 2 minutes the alert should be off. I don't know if such a query is possible.

To fill in the request on this comment I am going to implement the first alternative as for me it seems like the only real option here.

ricardomaraschini: After experimenting with different implementations I managed to find a way of making this work by using Counters instead of Gauges. With this approach we don't need the whole "clean up old Degraded reasons" magic and that, by itself, is a win in terms of complexity. PTAL.

pkg/storage/azure/azure.go
```diff
@@ -525,11 +525,6 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 			util.UpdateCondition(cr, imageregistryv1.StorageExists, operatorapiv1.ConditionTrue, storageExistsReasonContainerExists, "Storage container exists")
 
 			break
-
-			if len(d.Config.Container) == 0 {
-				util.UpdateCondition(cr, imageregistryv1.StorageExists, operatorapiv1.ConditionFalse, "CreateStorageFailed", "Unable to generate a unique container name")
-				return fmt.Errorf("unable to generate a unique azure container name")
-			}
```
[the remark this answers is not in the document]
ricardomaraschini: Answering to your question, it is not necessary. Do you prefer that I do this in another PR?

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,26 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	registry                   = prometheus.NewRegistry()
+	storageBackendReconfigured = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "storage_backend",
```
adambkaplan: nit - could just be `storage` (aligns with the object in the CRD).

pkg/storage/azure/azure.go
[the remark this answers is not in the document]
adambkaplan: Keep it in this PR - it's dead code.

pkg/operator/status.go
```diff
@@ -44,6 +45,19 @@ func updateCondition(cr *imageregistryv1.Config, condtype string, condstate oper
 	}
 
 	cr.Status.Conditions = conditions
+	updateDegradedMetric(condtype, condstate)
```
adambkaplan: Isn't `updateCondition` called multiple times? Can't this result in a lot of increments per reconciliation loop?

pkg/metrics/metrics.go
ricardomaraschini: Will fix it

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,26 @@
...
+	storageReconfigured = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "storage",
+		Name:      "reconfigured",
+		Help:      "Number of times registry storage got reconfigured.",
+	})
+	degraded = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "status",
+		Name:      "degraded",
```
adambkaplan:
```suggestion
		Name:      "degraded_total",
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,26 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	registry            = prometheus.NewRegistry()
+	storageReconfigured = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "storage",
+		Name:      "reconfigured",
```
adambkaplan:
```suggestion
		Name:      "reconfigured_total",
```

pkg/metrics/metrics.go
ricardomaraschini: Done.

pkg/operator/status.go
ricardomaraschini: My bad, `updateCondition` is called multiple times but only once using `Degraded` condition as parameter. This never increases more than once per loop.

pkg/metrics/metrics.go
ricardomaraschini: I can't accept this proposal here as it will break things. I need to update the tests and also the query done using this parameter inside manifests. I am going to send a commit with this update.

pkg/metrics/metrics.go
ricardomaraschini: I can't accept this proposal here as it will break things. I need to update the tests and also the query done using this parameter inside manifests. I am going to send a commit with this update.

pkg/metrics/metrics.go
ricardomaraschini: Done.

pkg/metrics/metrics.go
ricardomaraschini: Done.

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,26 @@
...
+
+import "github.com/prometheus/client_golang/prometheus"
+
+var (
+	registry            = prometheus.NewRegistry()
+	storageReconfigured = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "storage",
+		Name:      "reconfigured_total",
+		Help:      "Number of times registry storage got reconfigured.",
```
adambkaplan: grammar nit:
```suggestion
		Help:      "Total times the image registry's storage was reconfigured.",
```

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,26 @@
...
+	})
+	degraded = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "status",
+		Name:      "degraded_total",
+		Help:      "Number times the Operator reported Degraded, per reason.",
```
adambkaplan: nit:
```suggestion
		Help:      "Total times the image registry operator reported itself `Degraded`, per reason.",
```

PR #373, opened 2019-08-22
Report HasTrustedCA Status

adambkaplan:
Report the operator at level if the trusted CA bundle has not been injected yet.
This will allow upgrades to proceed while the networking operator rolls out.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -2,12 +2,14 @@ package operator
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 	kmeta "k8s.io/apimachinery/pkg/api/meta"
 	metaapi "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeinformers "k8s.io/client-go/informers"
 	kubeset "k8s.io/client-go/kubernetes"
@@ -44,6 +46,7 @@ const (
 	kubeSystemNamespace      = "kube-system"
 	workqueueKey             = "changes"
 	defaultResyncDuration    = 10 * time.Minute
+	defaultDegradedTimeout   = 1 * time.Hour
 )
 
 type permanentError struct {
@@ -91,13 +94,27 @@ func NewController(kubeconfig *restclient.Config) (*Controller, error) {
 
 	listers := &regopclient.Listers{}
 	clients := &regopclient.Clients{}
+
+	var degradedTimeout time.Duration
+	timeout, ok := os.LookupEnv("OPERATOR_DEGRADED_TIMEOUT")
+	if ok {
+		degradedTimeout, err = time.ParseDuration(timeout)
+		if err != nil {
+			klog.V(4).Infof("Could not parse degraded timeout %s, defaulting to 1h", timeout)
+			degradedTimeout = defaultDegradedTimeout
+		}
+	} else {
+		degradedTimeout = defaultDegradedTimeout
+	}
+
 	c := &Controller{
-		kubeconfig: kubeconfig,
-		params:     p,
-		generator:  resource.NewGenerator(kubeconfig, clients, listers, &p),
-		workqueue:  workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Changes"),
-		listers:    listers,
-		clients:    clients,
+		kubeconfig:      kubeconfig,
+		params:          p,
+		generator:       resource.NewGenerator(kubeconfig, clients, listers, &p),
+		workqueue:       workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "Changes"),
+		listers:         listers,
+		clients:         clients,
+		degradedTimeout: degradedTimeout,
 	}
 
 	// Initial event to bootstrap CR if it doesn't exist.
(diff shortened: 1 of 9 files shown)


adambkaplan:
/assign @dmage 
/cc @ricardomaraschini @coreydaley 

This is another late requirement in light of https://bugzilla.redhat.com/show_bug.cgi?id=1743699.
I could use some help figuring out how to test this - perhaps we need to use CVO overrides to wipe out the `trusted-ca` ConfigMap.

pkg/operator/controller.go
```diff
@@ -154,6 +156,29 @@ func (c *Controller) sync() error {
...
+			Reason:  "TrustedCANotPresent",
+			Message: "Inactive due to trusted CA bundle not present.",
+		})
+		updateCondition(cr, operatorapi.OperatorStatusTypeProgressing, operatorapi.OperatorCondition{
+			Status:  operatorapi.ConditionFalse,
+			Reason:  "TrustedCANotPresent",
+			Message: "Inactive due to trusted CA bundle not present.",
+		})
+		updateCondition(cr, operatorapi.OperatorStatusTypeDegraded, operatorapi.OperatorCondition{
+			Status:  operatorapi.ConditionFalse,
```
dmage: True?

adambkaplan: I need to refine this a bit more - we in fact should report `Degraded=true` if we don't have CA's after a specific period of time (an hour perhaps).

pkg/operator/controller.go
```diff
@@ -491,3 +499,33 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 
 	return nil
 }
+
+func (c *Controller) hasTrustedCA() bool {
+	_, err := os.Stat("/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem")
+	return err != nil
```
ricardomaraschini: if err == nil then we have the trusted ca, no? it seems like you are doing the opposite here.

pkg/operator/controller.go
```diff
@@ -491,3 +499,33 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
...
+}
+
+func (c *Controller) syncOperatorStatus(cr *imageregistryv1.Config, prevCR *imageregistryv1.Config) error {
+	cr.Status.ObservedGeneration = cr.Generation
+	statusChanged := !reflect.DeepEqual(prevCR.Status, cr.Status)
```
ricardomaraschini: Just a small improvement here, we could design this as:

```
if reflect.DeepEqual(prevCR.Status, cr.Status) {
        // No changes, just return.
        return nil
}
```
We have the whole function within an if statement otherwise.

pkg/operator/controller.go
```diff
@@ -205,26 +230,9 @@ func (c *Controller) sync() error {
 		cr.ResourceVersion = updatedCR.ResourceVersion
 	}
```
dmage: As you extracted `syncOperatorStatus`, you can also extract `syncOperatorResource`.

adambkaplan:
/retitle Report HasTrustedCA Status

We should discuss if any bits of this PR are still relevant in light of #379 merging.

The code delaying degraded status needs to go, as adding a timeout on reporting degraded does not give us anything in this particular case. There may be other use cases where going degraded makes sense after a delay, but not here.

I think there is still value in reporting `HasTrustedCA` as a status condition. We could use this to report degraded if the trusted CA bundle disappears on us, as such.

1. 4.1 -> 4.2 upgrade or new 4.2, registry creates the new status condition and comes up with `HasTrustedCA=False`.
2. Some time later we get `HasTrustedCA=True` once the network operator controllers are running.
3. If for any reason we lose the trusted CA, we'll have `HasTrustedCA=False` again, in which case we should report `Degraded=True`. With the current logic, we are reverting back to using the default trust bundle.

@dmage @danehans @bparees @ricardomaraschini thoughts?

dmage:
If we don't support 4.1->4.3 upgrades, we can remove fallback to the default trust bundle in 4.3.

adambkaplan:
@dmage I'm pretty certain that we will never have an update path directly through a minor release. 4.1 clusters will always need to hit 4.2 before they are ready to upgrade to 4.3.

@bparees @smarterclayton can you please confirm?

bparees:
> @dmage I'm pretty certain that we will never have an update path directly through a minor release. 4.1 clusters will always need to hit 4.2 before they are ready to upgrade to 4.3.
@bparees @smarterclayton can you please confirm?

that is correct.

pkg/operator/status.go
```diff
@@ -98,6 +108,19 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
+		// In this situation the operator should initially report itself at level, but report degraded after a reasonable timeout.
+		// Later, when the trusted CA bundle is present, we proceed as usual and upgrade the operands.
+		operatorAvailable.Message = fmt.Sprintf("Trusted CA bundle is not present after waiting %s.", c.degradedTimeout)
+		operatorAvailable.Reason = "NoTrustedCA"
+		if !degradedTimeoutReached {
+			operatorAvailable.Message = "TrustedCA bundle is not present."
+			operatorAvailable.Status = operatorapiv1.ConditionTrue
+		} else {
+			klog.Error(operatorAvailable.Message)
+		}
```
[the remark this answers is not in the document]
bparees: are we really unavailable in this case?  the registry should still work i think, just pullthrough+proxy flows may fail right?

smarterclayton:
Correct

smarterclayton:
And we will likely start restricting old upgrades (we won't let you go from
new 4.1 to very old 4.2 because of potential for regressing CVE fixes)

PR #374, opened 2019-08-27
DEVEXP-403: Using hard link when creating watcher binary.

ricardomaraschini:
Watcher, on library-go, reports the same binary name when using symlink
as it searchs for the executable name by looking at /proc/pid/exe link.

/proc/pid/exe points to the same destination as the symlink points to,
this causes both process to have the same "name".

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN make build
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
 COPY --from=builder /go/src/github.com/openshift/cluster-image-registry-operator/tmp/_output/bin/cluster-image-registry-operator /usr/bin/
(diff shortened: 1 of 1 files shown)


ricardomaraschini:
To see how watches locates what process to kill, please see https://github.com/openshift/library-go/blob/master/pkg/operator/watchdog/proc.go#L17

dmage:
@ricardomaraschini can this trick be covered by tests?

dmage:
If it's hard to automate this test, can we create a document with step-by-step description how to test it manually (run this command, check that output, wait until this happens)?

ricardomaraschini:
> If it's hard to automate this test, can we create a document with step-by-step description how to test it manually (run this command, check that output, wait until this happens)?

I think we may have an e2e test that checks if the operator is restarting when the trusted-ca config map changes. That would spot if there is an issue with the watchdog mechanics. What do you think?

PR #377, opened 2019-08-28
Bug 1738475: Update condition reason code for invalid keyID

coreydaley:
A future PR will address setting the operators available false, degraded true

### DIFF
--- a/pkg/storage/gcs/gcs.go
+++ b/pkg/storage/gcs/gcs.go
@@ -283,7 +283,8 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 
 			if err != nil {
 				if gerr, ok := err.(*gapi.Error); ok {
-					util.UpdateCondition(cr, imageregistryv1.StorageEncrypted, operatorapi.ConditionFalse, strconv.Itoa(gerr.Code), gerr.Error())
+					util.UpdateCondition(cr, imageregistryv1.StorageEncrypted, operatorapi.ConditionFalse, "InvalidStorageConfiguration", gerr.Error())
+					return err
 				} else {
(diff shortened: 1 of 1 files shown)


dmage:
@coreydaley I don't see how this change can affect the Degraded condition. Have you tested it?

coreydaley:
> @coreydaley I don't see how this change can affect the Degraded condition. Have you tested it?

I thought we had discussed on slack doing that in a separate pr that would addresss the issue of iterating over the conditions and setting the available and degraded, maybe I misunderstood.

dmage:
I was a bit confused by `degraded true` in the description.

@adambkaplan Your comment in BZ is about additional wishes, it's not about how to fix the bug, did I get it right?

@coreydaley The referenced bug is about `panic`, when this PR gets merged, robots will move this bug to MODIFIED and QE will test it again. So we can't merge it until it fixes the bug.

pkg/storage/gcs/gcs.go
```diff
@@ -283,7 +283,8 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 
 			if err != nil {
 				if gerr, ok := err.(*gapi.Error); ok {
-					util.UpdateCondition(cr, imageregistryv1.StorageEncrypted, operatorapi.ConditionFalse, strconv.Itoa(gerr.Code), gerr.Error())
+					util.UpdateCondition(cr, imageregistryv1.StorageEncrypted, operatorapi.ConditionFalse, "InvalidStorageConfiguration", gerr.Error())
+					return err
```
adambkaplan: We can probably be a bit more intelligent here about the reason - ex: a 5xx error does not imply that we sent bad configuration to GCP.

PR #380, opened 2019-08-29
node-ca: Handle graceful shutdown when receiving TERM

smarterclayton:
When we receive a sig term, we should exit with 0 instead of failing.


ricardomaraschini:
@smarterclayton we already have a PR that adds this feature:

https://github.com/openshift/cluster-image-registry-operator/pull/372

smarterclayton:
/close

Fixed by #372

PR #382, opened 2019-09-02
Bug 1745811: Making sure we don't have a sequence of dashes.

ricardomaraschini:
In some cases infrastructure contains multiple dashes in a sequence, Azure does not like it. This patch makes sure we always remove a sequence of dashes by a single one.

Tests were refactored to contains a variable infra name.

During the investigations I have seen on the must-gather for the problematic deployment that the image registry was deployed with the following environment variable:

```
    Environment:
      REGISTRY_STORAGE_AZURE_CONTAINER:       qewalid420825233east--rhgcq-image-registry-exmgmspubjtcamgowcr
```

Double dashes(`--`) are considered invalid for a container name: https://docs.microsoft.com/en-us/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata

These two extra dashes were coming in from the Infrastructure name configuration.

### DIFF
--- a/pkg/storage/util/util.go
+++ b/pkg/storage/util/util.go
@@ -3,6 +3,7 @@ package util
 import (
 	"fmt"
 	"math/rand"
+	"regexp"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
Link for the must-gather: http://file.rdu.redhat.com/wabouham/OCP_4.2_BZs/Azure/4.2_08-25-image_registry_crashlooping/

PR #387, opened 2019-09-20
Bug 1753778: report Available when registry is explicitly Removed

dmage:


dmage:
@bparees yes, I want to backport the CDR from master next week.

test/e2e/managementstate_test.go
```diff
@@ -0,0 +1,141 @@
...
+		if err != nil {
+			return err
+		}
+		return nil
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// TODO(dmage): wait for the resource to be observed
```
ricardomaraschini: Are we planning to wait for this one before merging? This could generate some flake events, no?

dmage: No, we have retries below.

pkg/operator/status.go
```diff
@@ -87,10 +87,18 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
 		Message: "",
 		Reason:  "",
 	}
-	if deploy == nil {
+	if cr.Spec.ManagementState == operatorapiv1.Unmanaged {
+		operatorAvailable.Status = operatorapiv1.ConditionTrue
```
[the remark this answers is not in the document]
dmage: Yes, but we don't want to [block upgrades](https://github.com/openshift/cluster-version-operator/blob/a17c59a1fd6c03f9af40082c380888523d60820b/pkg/cvo/internal/operatorstatus.go#L149-L189), so we report that everything is OK.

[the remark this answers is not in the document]
adambkaplan: @dmage Think of it this way - if not in the `Managed` state, `Available` means "Can the _operator_ do what you ask it to do?" Unless we are not running, the answer is yes/`True`.

Ex: if the management state is changed from `Unmanaged` to `Managed`, will the operator react as expected? Yes.

adambkaplan:
/hold

Known issue with 1.16 storage tests (waiting for kubelet/RHCOS skew to be reconciled).

pkg/operator/status.go
```diff
@@ -87,10 +87,18 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
+		operatorAvailable.Message = "The registry configuration is set to unmanaged mode"
+		operatorAvailable.Reason = "Unmanaged"
+	} else if deploy == nil {
 		if e, ok := applyError.(permanentError); ok {
 			operatorAvailable.Message = applyError.Error()
 			operatorAvailable.Reason = e.Reason
+		} else if removed {
+			operatorAvailable.Status = operatorapiv1.ConditionTrue
+			operatorAvailable.Message = "The deployment is removed"
```
ricardomaraschini: Could we make this Message something like down below when we read:

"The registry is removed"

test/e2e/managementstate_test.go
ricardomaraschini: So we don't need the TODO right? Or I misunderstand you?

test/e2e/managementstate_test.go
```diff
@@ -0,0 +1,141 @@
...
+
+	err = client.Deployments(imageregistryv1.ImageRegistryOperatorNamespace).Delete(imageregistryv1.ImageRegistryName, &metav1.DeleteOptions{})
```
ricardomaraschini: Isn't this Delete automatically happen slightly after we set the Management State to Removed?

adambkaplan:
/lgtm cancel

@dmage please address or remove the TODOs in your comments.

/retest

adambkaplan:
/hold cancel

RHCOS skew issues should be resolved now

PR #391, opened 2019-09-23
OpenStack: quote passwords

Fedosin:
OpenStack passwords can contain special characters and therefore must be quoted to prevent yaml parsing errors.

Fixes: https://github.com/openshift/installer/issues/2392

### DIFF
--- a/pkg/storage/swift/swift.go
+++ b/pkg/storage/swift/swift.go
@@ -3,6 +3,7 @@ package swift
 import (
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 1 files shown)


dmage:
The registry parses environment variables as YAML. Most of the time YAML treats a random sequence of symbols as a string. But sometimes it doesn't. So if we have a user-provided string and we want to use it as a value for the `REGISTRY_` environment, we should yaml-encode it.

It's better to encode values in `ConfigEnv` and `Secrets` (these values should be used only as EnvVarSource).

/cc @coreydaley @ricardomaraschini

adambkaplan:
> But sometimes it doesn't. So if we have a user-provided string and we want to use it as a value for the `REGISTRY_` environment, we should yaml-encode it.

Particularly for YAML that is all-numeric or a boolean equivalent (`true`, `false`)

dmage:
I'd suggest to change ConfigEnv to return map[string]interface{} and encode values in the PodTemplateSpec generator. @Fedosin do you want me to do it?

PR #392, opened 2019-09-24
DEVEXP-419: Bootstrapping operator as Removed on BareMetal.

ricardomaraschini:
If we detect that the platform we are bootstrapping on is bare metal we
set management state to "Removed". We still report ourselves as "Available"
but with "Reason" set to "Removed".

### DIFF
--- a/manifests/09-prometheus-rules.yaml
+++ b/manifests/09-prometheus-rules.yaml
@@ -17,3 +17,14 @@ spec:
         message: |
           Image Registry Storage configuration has changed in the last 30
           minutes. This change may have caused data loss.
+    - alert: ImageRegistryRemoved
+      expr: cluster_operator_conditions{name="image-registry",condition="Available",reason="Removed"} > 0
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        message: |
+          Image Registry has been removed. ImageStreamTags, BuildConfigs and
+          DeploymentConfigs which reference ImageStreamTags may not work as
+          expected. Please configure storage and update the config to Managed
+          state by editing configs.imageregistry.operator.openshift.io.
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -13,12 +13,16 @@ import (
 	imageregistryv1 "github.com/openshift/cluster-image-registry-operator/pkg/apis/imageregistry/v1"
 	regopset "github.com/openshift/cluster-image-registry-operator/pkg/generated/clientset/versioned/typed/imageregistry/v1"
 	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage"
 )
 
 // randomSecretSize is the number of random bytes to generate
 // for the http secret
 const randomSecretSize = 64
 
+// Bootstrap registers this operator with OpenShift by creating an appropriate
+// ClusterOperator custom resource. This function also creates the initial
+// configuration for the Image Registry.
 func (c *Controller) Bootstrap() error {
 	cr, err := c.listers.RegistryConfigs.Get(imageregistryv1.ImageRegistryResourceName)
 	if err != nil && !errors.IsNotFound(err) {
@@ -40,6 +44,22 @@ func (c *Controller) Bootstrap() error {
 		return fmt.Errorf("could not generate random bytes for HTTP secret: %s", err)
 	}
 
+	platformStorage, err := storage.GetPlatformStorage(c.listers)
+	if err != nil {
+		return err
+	}
+
+	noStorage := imageregistryv1.ImageRegistryConfigStorage{}
+
+	// We bootstrap as "Removed" if the platform is known and does not
+	// provide persistent storage out of the box. If the platform is
+	// unknown we will bootstrap as Managed but using EmptyDir storage
+	// engine(ephemeral).
+	mgmtState := operatorapi.Managed
+	if platformStorage == noStorage {
+		mgmtState = operatorapi.Removed
+	}
+
 	cr = &imageregistryv1.Config{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:       imageregistryv1.ImageRegistryResourceName,
@@ -47,7 +67,7 @@ func (c *Controller) Bootstrap() error {
 			Finalizers: []string{parameters.ImageRegistryOperatorResourceFinalizer},
 		},
 		Spec: imageregistryv1.ImageRegistrySpec{
-			ManagementState: operatorapi.Managed,
+			ManagementState: mgmtState,
 			LogLevel:        2,
 			Storage:         imageregistryv1.ImageRegistryConfigStorage{},
 			Replicas:        1,
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -105,7 +105,7 @@ func newDriver(cfg *imageregistryv1.ImageRegistryConfigStorage, kubeconfig *rest
 func NewDriver(cfg *imageregistryv1.ImageRegistryConfigStorage, kubeconfig *rest.Config, listers *regopclient.Listers) (Driver, error) {
 	drv, err := newDriver(cfg, kubeconfig, listers)
 	if err == ErrStorageNotConfigured {
-		*cfg, err = getPlatformStorage(listers)
+		*cfg, err = GetPlatformStorage(listers)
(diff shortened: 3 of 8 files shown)


pkg/operator/bootstrap.go
```diff
@@ -40,14 +45,25 @@ func (c *Controller) Bootstrap() error {
...
+	if err != nil {
+		return fmt.Errorf("could not get infra configuration")
+	}
+
+	// We bootstrap ourselves as Removed if we are running on BareMetal.
+	mgmtState := operatorapi.Managed
+	if infra.Status.PlatformStatus.Type == configapiv1.BareMetalPlatformType {
+		mgmtState = operatorapi.Removed
+	}
+
```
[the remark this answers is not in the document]
adambkaplan: @bparees agree this needs to be in a separate PR (and will be in a separate code location - this sets the initial operator config if there is none already on the cluster)

test/e2e/baremetal_test.go
```diff
@@ -25,27 +25,20 @@ func TestBaremetalDefaults(t *testing.T) {
...
-			if cond.Status != configapiv1.ConditionFalse {
-				t.Errorf("expected clusteroperator to report Available=%s, got %s", configapiv1.ConditionFalse, cond.Status)
-			}
-		case configapiv1.OperatorDegraded:
 			if cond.Status != configapiv1.ConditionTrue {
-				t.Errorf("expected clusteroperator to report Degraded=%s, got %s", configapiv1.ConditionTrue, cond.Status)
+				t.Errorf("expected clusteroperator to report Available=%s, got %s", configapiv1.ConditionTrue, cond.Status)
+			}
+			if cond.Reason != "Removed" {
+				t.Errorf("expected reason to be 'Removed', got %s", cond.Reason)
```
adambkaplan: @ricardomaraschini I believe that we have a utility method to ensure that the cluster operator is reporting itself "green" (available=true, progressing=false, degraded=false) - `MustEnsureClusterOperatorStatusIsNormal`

pkg/operator/bootstrap.go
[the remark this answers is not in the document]
bparees: yeah i put my comment in the wrong place, i meant to make it on the logic below that sets the operator status based on the management state.

regardless, separate PR is fine.

pkg/operator/status.go
```diff
@@ -87,7 +87,12 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
 		Message: "",
 		Reason:  "",
 	}
-	if deploy == nil {
+
+	if removed {
```
dmage: https://github.com/openshift/cluster-image-registry-operator/pull/387

test/e2e/baremetal_test.go
ricardomaraschini: Good one! I also think it is important to check if `Reason` is properly set to "Removed" in all `Conditions`. I am building another solution using another already provided function called `framework.ConditionExistsWithStatusAndReason`. PTAL.

pkg/operator/status.go
ricardomaraschini: Yes, we need to get #387 merged first.

ricardomaraschini: I have rebased this PR on top of #387 .

pkg/storage/storage.go
```diff
@@ -125,17 +134,18 @@ func getPlatformStorage(listers *regopclient.Listers) (imageregistryv1.ImageRegi
...
-		// EmptyDir will never be used in a production cluster.
-		cfg.EmptyDir = &imageregistryv1.ImageRegistryConfigStorageEmptyDir{}
+
+	// These are the platforms we don't configure any backend for, on these
+	// we should bootstrap the image registry as "Removed".
+	case configapiv1.BareMetalPlatformType,
+		configapiv1.OvirtPlatformType,
+		configapiv1.LibvirtPlatformType,
+		configapiv1.NonePlatformType,
+		configapiv1.VSpherePlatformType:
```
adambkaplan: `libvirt` may in fact support PVC with 10Gi of storage. We can revisit this in a separate PR.

test/e2e/managementstate_test.go
```diff
@@ -131,3 +131,56 @@ func TestManagementStateRemoved(t *testing.T) {
 		t.Fatalf("deployment is expected to be removed, got %v %v", d, err)
 	}
 }
+
+func TestRemovedToManagedTransition(t *testing.T) {
```
adambkaplan: @ricardomaraschini I suspect this test is pushing the e2e over the 30 minute timeout in the makefile target. Origin uses 2 hours as an acceptable timeout for e2e tests - I think we can do the same given that these are operator tests that cause rollouts to occur.

ricardomaraschini: Increased to 2h.

pkg/storage/storage.go
ricardomaraschini: Are they going to provide a PVC with a default name and we are going to bootstrap using it?

dmage: @ricardomaraschini they are going to reconfigure the bootstrapped registry config because by default we use EmptyDir for libvirt.

manifests/09-prometheus-rules.yaml
```diff
@@ -17,3 +17,14 @@ spec:
...
+      expr: cluster_operator_conditions{name="image-registry",condition="Available",reason="Removed"} > 0
+      for: 5m
+      labels:
+        severity: warning
+      annotations:
+        message: |
+          Image Registry has been removed. ImageStreamTags, BuildConfigs and
+          DeploymentConfigs which reference ImageStreamTags may not work as
+          expected. Please configure storage and update the config to Managed
+          state editing configs.imageregistry.operator.openshift.io.
```
adambkaplan: nit:
```suggestion
          state by editing configs.imageregistry.operator.openshift.io.
```

pkg/operator/bootstrap.go
```diff
@@ -13,12 +13,16 @@ import (
...
 	regopset "github.com/openshift/cluster-image-registry-operator/pkg/generated/clientset/versioned/typed/imageregistry/v1"
 	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage"
 )
 
 // randomSecretSize is the number of random bytes to generate
 // for the http secret
 const randomSecretSize = 64
 
+// Bootstrap registers this operator with OpenShift by creating appropriate
```
adambkaplan: nit:
```suggestion
// Bootstrap registers this operator with OpenShift by creating an appropriate
```

pkg/storage/storage.go
```diff
@@ -114,10 +114,20 @@ func NewDriver(cfg *imageregistryv1.ImageRegistryConfigStorage, kubeconfig *rest
...
-// getPlatformStorage returns the storage configuration that should be used
+// GetPlatformStorage returns the storage configuration that should be used
 // based on the cloudplatform we are running on, as determined from the
 // infrastructure configuration.
-func getPlatformStorage(listers *regopclient.Listers) (imageregistryv1.ImageRegistryConfigStorage, error) {
+//
+// Following rules apply:
+// - If it is a known platform for which we have a backend implementation (e.g.
+//   AWS) we return a storage configuration that uses that implementation.
+// - If it is a known platform and it doesnt provide any backend implementation
```
adambkaplan: nit:
```suggestion
// - If it is a known platform and it doesn't provide any backend implementation,
```

pkg/storage/storage.go
```diff
@@ -145,6 +155,14 @@ func getPlatformStorage(listers *regopclient.Listers) (imageregistryv1.ImageRegi
 		cfg.GCS = &imageregistryv1.ImageRegistryConfigStorageGCS{}
 	case configapiv1.OpenStackPlatformType:
 		cfg.Swift = &imageregistryv1.ImageRegistryConfigStorageSwift{}
+
+	// Unknown platforms or LibVirt: we configure image registry using
+	// EmptyDir storage. We could omit LibVirt on this switch but it is
```
adambkaplan: You can remove the second sentence

manifests/09-prometheus-rules.yaml
adambkaplan: /cc @bmcelvee 

If necessary we can revisit the alert language.

adambkaplan:
/hold cancel

https://github.com/openshift/release/pull/5649 is in the merge pool

PR #394, opened 2019-09-30
Bug 1722878: reseting configured storages before DeepCopy()

ricardomaraschini:
When migrating from one storage to another configs/cluster Status ended up
showing multiple storage configurations, for example:

```
status:
  storage:
    emptyDir: {}
    s3:
      bucket: rmarasch-cluster-66-575tx-image-registry-us-east-1-porcoxpfplu
      encrypt: true
      keyID: ""
      region: us-east-1
      regionEndpoint: ""
```

This patch resets Storage property before running any DeepCopy(), making
sure that if the storage is changed to another media we will update
correctly.

### DIFF
--- a/pkg/resource/generator.go
+++ b/pkg/resource/generator.go
@@ -103,6 +103,20 @@ func (g *Generator) syncStorage(cr *imageregistryv1.Config) error {
 			return err
 		}
 	}
+
+	// XXX BZ 1722878
+	// This is a workaround for a bug that affected OCP 4.2 and earlier. We
+	// may have multiple storage engines set on Status, that is not a valid
+	// situation. Here we check if this is the case and if yes we fix it by
+	// copying what we have on the Spec into the Status.
+	_, err = storage.NewDriver(
+		&cr.Status.Storage, g.kubeconfig, g.listers,
+	)
+	if _, ok := err.(*storage.MultiStoragesError); ok {
+		specCopy := cr.Spec.Storage.DeepCopy()
+		cr.Status.Storage = *specCopy
+	}
+
 	return nil
 }
 
--- a/pkg/storage/azure/azure.go
+++ b/pkg/storage/azure/azure.go
@@ -419,7 +419,9 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 		}
 
 		cr.Status.StorageManaged = false
-		cr.Status.Storage.Azure = d.Config.DeepCopy()
+		cr.Status.Storage = imageregistryv1.ImageRegistryConfigStorage{
+			Azure: d.Config.DeepCopy(),
+		}
 		util.UpdateCondition(cr, imageregistryv1.StorageExists, operatorapiv1.ConditionTrue, storageExistsReasonUserManaged, "Storage is managed by the user")
 	} else {
 		// IPI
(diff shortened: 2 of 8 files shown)


dmage:
We also need to provide a way to fix existing resources that already have two storage configurations in status. Ideally it should be fixed automatically by the operator.

coreydaley:
I think that there should be one place in the code that checks to see whether the storage configuration has changed from one type to another, and it would then clear the old storage configuration (should it run the finalizer?) and set up the new storage as if it had started from a clean slate, rather than checking/resetting each time that we do a copy of spec to status.

Thoughts?

coreydaley:
I also thought that https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/storage.go#L84 was supposed to catch that? Maybe we are not implementing it correctly or maybe I am mistaken about what it is supposed to do?

ricardomaraschini:
@dmage Are you sure you linked the right BZ? Is the problem related there solved by this patch? I have no idea I was solving it when I worked on this one :-)

ricardomaraschini:
> I think that there should be one place in the code that checks to see whether the storage configuration has changed from one type to another, and it would then clear the old storage configuration

I have added a function on #368 that checks if the storage has been changed, that function is used for metrics but could be adjusted for the purpose(check if there are more than one storage enabled on Status and reset it to be the Spec one).

> (should it run the finalizer?)

I am really not sure about this one. Is it desirable? By `finalizer` you mean `delete` the old storage right? I would prefer to let it there because a simply typo on the config should not make the user to loose its entire data(at least not without an extra confirmation step), well, I am assuming that finalizer is indeed delete.

ricardomaraschini:
> I also thought that https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/storage.go#L84 was supposed to catch that? Maybe we are not implementing it correctly or maybe I am mistaken about what it is supposed to do?

This avoids the Driver to be created with multiple storage at the same time, the problem here seems to be when we "swap" storages from, let's say, `s3` to `emptyDir`.

ricardomaraschini:
> We also need to provide a way to fix existing resources that already have two storage configurations in status. Ideally it should be fixed automatically by the operator.

I could do small adjustments on https://github.com/openshift/cluster-image-registry-operator/pull/368/files#diff-f201afca3f23e7411fe9cb0cfc8a2011R135 and get this working but I would need #368 merged first.

ricardomaraschini:
@dmage @coreydaley What do you guys think about something like this?

https://github.com/openshift/cluster-image-registry-operator/pull/394/commits/00ada4a83cf3cd7b0cb4b3804ff863084e9c23f2

ricardomaraschini:
@dmage @coreydaley What do you guys think about a workaround like https://github.com/openshift/cluster-image-registry-operator/pull/394/commits/dba9c37e74ecf316a246d0af139912b207914786 ?

pkg/resource/generator.go
```diff
@@ -103,6 +103,19 @@ func (g *Generator) syncStorage(cr *imageregistryv1.Config) error {
...
 	}
+
+	// XXX
+	// This is a workaround for a bug that affected old releases. We may
+	// have multiple storage engines set on Status, that is not a valid
+	// situation. Here we check if this is the case and if yes we fix it
+	// by copying what we have on the Spec into the Status.
+	if _, err := storage.NewDriver(
+		&cr.Status.Storage, g.kubeconfig, g.listers,
+	); err == storage.ErrMultipleStorageConfigured {
```
adambkaplan: Nits:

1. Please reference the bug ID here - always good to refer to if this is an issue in the future.
2. /s/"old releases"/"OCP 4.2 and earlier"/

ricardomaraschini: Excellent points. Done.

dmage:
@adambkaplan as the fix is relatively straightforward, I'd like to backport it to 4.2.

adambkaplan:
@dmage @ricardomaraschini I don't think we can backport without some level of unit or e2e tests. QE does not have the bandwidth to verify every bug in every weekly z-stream.

ricardomaraschini:
> @dmage @ricardomaraschini I don't think we can backport without some level of unit or e2e tests. QE does not have the bandwidth to verify every bug in every weekly z-stream.

I will write a e2e test case to cover this.

PR #398, opened 2019-10-04
Implementing test for storage swap.

ricardomaraschini:
When user swaps storages on image registry configuration only one should
still be enabled on Status. Added test to swap an IPI provided storage
by an EmptyDir one and checking if only EmptyDir is still set on Status.

This is a test to cover #394 , it is a nice to have even if we decide not to 
backport the fix.

PR #399, opened 2019-10-07
[4.2] Bug 1759159: Reset storage when swapping back end type

ricardomaraschini:
When migrating from one storage to another configs/cluster Status ended up showing multiple storage configurations, for example:

```
    status:
      storage:
        emptyDir: {}
        s3:
          bucket: rmarasch-cluster-66-575tx-image-registry-us-east-1-porcoxpfplu
          encrypt: true
          keyID: ""
          region: us-east-1
          regionEndpoint: ""
 ```
  
This patch resets Storage property before running any DeepCopy(), making sure that if the storage is changed to another media we will update correctly.

### DIFF
--- a/pkg/resource/generator.go
+++ b/pkg/resource/generator.go
@@ -103,6 +103,20 @@ func (g *Generator) syncStorage(cr *imageregistryv1.Config) error {
 			return err
 		}
 	}
+
+	// XXX BZ 1722878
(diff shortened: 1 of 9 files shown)


adambkaplan:
Patch manager note - this is a manual cherry-pick of two PRs:

1. https://github.com/openshift/cluster-image-registry-operator/pull/394
2. https://github.com/openshift/cluster-image-registry-operator/pull/398

PR #404, opened 2019-10-31
WIP - Bug 1726976: Using PVC ResourceVersion when creating the deploy.

ricardomaraschini:
If something has changed on the PVC we need to restart the deploy. This
patch makes use of PVC's resource version inside the deployment in such
a way that if the revision changes the deployment will be different.


ricardomaraschini:
Wrong patch. Addressed the problem in the wrong way.

PR #406, opened 2019-11-04
Bug 1753778: Set Version on Operator status even if Removed.

ricardomaraschini:
This patch uses the environment variable RELEASE_VERSION to set Operator
version if ManagementState is not "Managed". This is introduced now as we
may, in some situations, bootstrap the Image Registry Operator with Removed
ManagementState.

### DIFF
--- a/pkg/resource/clusteroperator.go
+++ b/pkg/resource/clusteroperator.go
@@ -2,6 +2,7 @@ package resource
 
 import (
 	"fmt"
+	"os"
 	"reflect"
 
 	appsapi "k8s.io/api/apps/v1"
@@ -217,33 +218,49 @@ func isDeploymentStatusAvailableAndUpdated(deploy *appsapi.Deployment) bool {
 		deploy.Status.UpdatedReplicas == deploy.Status.Replicas
 }
 
-func (gco *generatorClusterOperator) syncVersions(op *configapi.ClusterOperator) (modified bool, err error) {
-	deploy, err := gco.deployLister.Get(imageregistryv1.ImageRegistryName)
-	if err != nil {
-		if kerrors.IsNotFound(err) {
-			err = nil
-		}
-		return
+// syncVersions updates reported version.
+//
+// If in "Managed" state we use the version stored as a annotation on registry'
+// Deployment, if not we use RELEASE_VERSION environment variable.
+func (gco *generatorClusterOperator) syncVersions(op *configapi.ClusterOperator) (bool, error) {
+	if gco.cr == nil {
+		return false, fmt.Errorf("invalid nil configuration provided")
 	}
 
-	deploymentVersion := deploy.Annotations[imageregistryv1.VersionAnnotation]
-	if len(deploymentVersion) == 0 || !isDeploymentStatusAvailableAndUpdated(deploy) {
-		return
+	version := os.Getenv("RELEASE_VERSION")
+	if gco.cr.Spec.ManagementState == operatorapi.Managed {
+		deploy, err := gco.deployLister.Get(imageregistryv1.ImageRegistryName)
+		if err != nil {
+			if kerrors.IsNotFound(err) {
+				return false, nil
+			}
+			return false, err
+		}
+
+		if !isDeploymentStatusAvailableAndUpdated(deploy) {
+			return false, nil
+		}
+
+		version = deploy.Annotations[imageregistryv1.VersionAnnotation]
+	}
+
+	if len(version) == 0 {
+		return false, nil
 	}
 
 	newVersions := []configapi.OperandVersion{
 		{
 			Name:    "operator",
-			Version: deploymentVersion,
+			Version: version,
 		},
 	}
 
-	if !reflect.DeepEqual(op.Status.Versions, newVersions) {
-		op.Status.Versions = newVersions
-		modified = true
+	if reflect.DeepEqual(op.Status.Versions, newVersions) {
+		return false, nil
 	}
 
-	return
+	op.Status.Versions = newVersions
+	return true, nil
 }
 
 func (gco *generatorClusterOperator) syncRelatedObjects(op *configapi.ClusterOperator) (modified bool) {
--- a/pkg/resource/clusteroperator_test.go
+++ b/pkg/resource/clusteroperator_test.go
@@ -1,0 +1,278 @@
+package resource
+
+import (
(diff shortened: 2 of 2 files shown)


pkg/resource/clusteroperator.go
```diff
@@ -217,33 +218,44 @@ func isDeploymentStatusAvailableAndUpdated(deploy *appsapi.Deployment) bool {
...
-		return
+	if len(version) == 0 {
+		return false, fmt.Errorf("invalid empty operator version")
 	}
 
 	newVersions := []configapi.OperandVersion{
 		{
 			Name:    "operator",
-			Version: deploymentVersion,
+			Version: version,
```
adambkaplan: It's not explicit in the docs, but other operators can report versions for other things [1]. I propose we append the `image-registry` item to this slice if the registry has been deployed.

[1] https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md

dmage: @adambkaplan do we want to report the "operator" version immediately then? For some reason today we report our version only when we've successfully created a deployment.

ricardomaraschini: I changed the approach to report versions(operator and image-registry) separately. PTAL.

ricardomaraschini:
@adambkaplan I have implemented unit testing for the `syncVersions()` function.

pkg/resource/clusteroperator.go
adambkaplan: @dmage I seem to recall there was a lot of churn around when to report version in the 4.0 -> 4.1 timeframe.

Today I agree we should report the operator version immediately. In principle we should have `Progressing=true` during upgrade (and if not, that is a bug) [1]. Version for the image-registry shouldn't be updated until we confirm that the deployment has rolled out.

[1] https://github.com/openshift/cluster-version-operator/blob/master/docs/dev/clusteroperator.md#what-should-an-operator-report-with-clusteroperator-custom-resource

pkg/resource/clusteroperator.go
```diff
@@ -217,33 +218,44 @@ func isDeploymentStatusAvailableAndUpdated(deploy *appsapi.Deployment) bool {
...
+	} else if err == nil {
+		v := deploy.Annotations[imageregistryv1.VersionAnnotation]
+		if len(v) == 0 || !isDeploymentStatusAvailableAndUpdated(deploy) {
+			return false, nil
+		}
+
+		newVersions = append(newVersions, configapi.OperandVersion{
+			Name:    "image-registry",
+			Version: v,
+		})
```
adambkaplan: I think on upgrade we can have a state where we report `operator` version, but nothing for the `image-registry` even though we have an existing registry there.

I propose the following for this sync function:

1. Convert the current `op.Status.Versions` slice into a `map[string]string` (map[name]version)
2. Always update the `operator` key with the current version
3. Update the `image-registry` key if the deployment has the right annotation and has rolled out.
4. Convert the map back to a `Versions` slice.

pkg/resource/clusteroperator_test.go
```diff
@@ -0,0 +1,192 @@
...
+				{
+					Name:    "operator",
+					Version: "1",
+				},
+				{
+					Name:    "image-registry",
+					Version: "2",
+				},
+			},
+		},
```
adambkaplan: Add a realistic upgrade scenario

1. Operator is at version 2
2. image-registry is still at version 1

ricardomaraschini: Done.

pkg/resource/clusteroperator.go
ricardomaraschini: Done.

pkg/resource/clusteroperator_test.go
ricardomaraschini: Done.

pkg/resource/clusteroperator.go
ricardomaraschini: I got the idea. The approach of converting from `map[string]string` back into a `[]OperandVersion` is incompatible with `reflect.DeepEqual` though, the resulting slice will certainly contain all Versions but there is no guarantee in what order they are going to show up on the new slice: https://blog.golang.org/go-maps-in-action#TOC_7.

I implemented the same behavior in a different way, ptal.

pkg/resource/clusteroperator.go
```diff
@@ -217,33 +218,57 @@ func isDeploymentStatusAvailableAndUpdated(deploy *appsapi.Deployment) bool {
...
 
-	if !reflect.DeepEqual(op.Status.Versions, newVersions) {
-		op.Status.Versions = newVersions
-		modified = true
+	for i, v := range op.Status.Versions {
+		if v.Name != name {
+			continue
+		}
+		op.Status.Versions[i].Version = version
+		return v.Version != version
```
adambkaplan: Be glad golang doesn't have Java's `ConcurrentModificationException`!

pkg/resource/clusteroperator.go
```diff
@@ -217,33 +218,57 @@ func isDeploymentStatusAvailableAndUpdated(deploy *appsapi.Deployment) bool {
...
-		return
 	}
 
-	deploymentVersion := deploy.Annotations[imageregistryv1.VersionAnnotation]
-	if len(deploymentVersion) == 0 || !isDeploymentStatusAvailableAndUpdated(deploy) {
-		return
+	if v := os.Getenv("RELEASE_VERSION"); len(v) > 0 {
+		if gco.updateVersionInPlace(op, "operator", v) {
+			updated++
+		}
```
adambkaplan: @ricardomaraschini I think my own confusion around when to report the `operator` version led us astray here.

We need the following logic:

1. If the operator is `Managed`, we can't update the `operator` version until the image registry has rolled out (via `isDeploymentStatusAvailableAndUpdated`)
2. For all other management states, report the `operator` version (and truthfully we shouldn't report any other version number).

adambkaplan:
/retitle Bug 1753778,1768357: Set Version on Operator status even if Removed.

adambkaplan:
/retitle Bug 1753778, 1768357: Set Version on Operator status even if Removed.

adambkaplan:
/retitle Bug 1753778: Set Version on Operator status even if Removed.

PR #407, opened 2019-11-05
DEVEXP-191: CronJob to automate image registry pruning

coreydaley:


pkg/resource/prunercronjob.go
```diff
@@ -0,0 +1,208 @@
...
+			StartingDeadlineSeconds:    gcj.getStartingDeadlineSeconds(),
+			JobTemplate: batchapi.JobTemplateSpec{
+				Spec: batchv1.JobSpec{
+					Template: kcorev1.PodTemplateSpec{
+						Spec: kcorev1.PodSpec{
+							RestartPolicy:      kcorev1.RestartPolicyOnFailure,
+							ServiceAccountName: "pruner",
+							Containers: []kcorev1.Container{
+								{
+									Image:                    "quay.io/openshift/origin-cli:4.1",
```
adambkaplan: @coreydaley I believe there is an `ImageStreamTag` for `oc` that we should reference instead. @gabemontero I believe the samples operator (or CVO) does this along with the must-gather tool?

If we have such imagestreamtag, we'd have to dynamically inject the pull spec to the internal registry - I don't think `CronJob` supports the image-lookup [1], and even so will it resolve imagestreams in the openshift namespace?

[1] https://docs.openshift.com/container-platform/3.11/admin_guide/image_policy.html#image-policy-configuring-the-image-policy-admission-plug-in

bparees: +1, use the imagestream. this would also fail in disconnected envs and using a version tag that needs to be updated would be problematic.

adambkaplan:
@coreydaley this PR needs to be rebased for us to get any kind of test coverage.

At this point I'd like to get the working bits merged so other work - like publishing metrics - is unblocked.

coreydaley:
/retest
Looks like resource setup failed.

coreydaley:
/retest
level=error msg="Error: ResourceLimitExceeded: Task limit exceeded"

coreydaley:
/retest
setup errors

adambkaplan:
@coreydaley the image used for the image pruner (oc) needs to be added to the image references file:

https://github.com/openshift/cluster-image-registry-operator/blob/master/manifests/image-references

adambkaplan:
@bparees how do we know which image we should reference so that the image replacement works correctly?

coreydaley:
@adambkaplan Can i use the one(s) from here? https://github.com/openshift/cluster-samples-operator/blob/master/manifests/image-references

adambkaplan:
@coreydaley let's go with the cli image reference - don't see why that would not work.

pkg/resource/prunercronjob.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		return nil, err
+	}
+	gcj.cr = cr
+
+	cj := &batchapi.CronJob{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        gcj.GetName(),
+			Namespace:   gcj.GetNamespace(),
+			Annotations: map[string]string{cronJobOwnerAnnotation: "true"},
```
dmage: Why do you need this annotation?

adambkaplan:
@coreydaley another note - the image used for the pruner needs to be passed to the operator as a new env var - see https://github.com/openshift/cluster-image-registry-operator/blob/master/manifests/07-operator.yaml#L55-L56

pkg/resource/prunercronjob.go
```diff
@@ -0,0 +1,237 @@
...
+func (gcj *generatorPrunerCronJob) GetName() string {
+	return "image-pruner"
+}
+
+func (gcj *generatorPrunerCronJob) expected() (runtime.Object, error) {
+	cr, err := gcj.prunerLister.Get(defaults.ImageRegistryImagePrunerResourceName)
+	if err != nil {
+		return nil, err
+	}
+	gcj.cr = cr
```
dmage: Please don't. You never know when `gcj.cr` gets changed and it makes code more complicated.

If you want to have a bunch of helpers to get values from this cr, pass this cr as an argument. Don't hide dependencies between functions.

pkg/operator/controller.go
```diff
@@ -116,7 +120,7 @@ type Controller struct {
 	clients    *regopclient.Clients
 }
 
-func (c *Controller) createOrUpdateResources(cr *imageregistryv1.Config) error {
+func (c *Controller) createOrUpdateResources(cr *imageregistryv1.Config, pcr *imageregistryv1.ImagePruner) error {
```
dmage: There is no need in this change.

pkg/resource/prunercronjob.go
coreydaley: to show that the cronjob was created by the operator?

pkg/resource/prunercronjob.go
coreydaley: updated in next push

pkg/operator/controller.go
coreydaley: removed in next push, that was a leftover that i didn't catch from the previous way i was doing things.

pkg/resource/prunercronjob.go
dmage: but we don't have such annotations for other resources

coreydaley: Saw it in https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/resource/route.go , guess I should have checked the others also, i will remove it from mine.

manifests/07-operator.yaml
```diff
@@ -54,6 +54,8 @@ spec:
               value: "cluster-image-registry-operator"
             - name: IMAGE
               value: docker.io/openshift/origin-docker-registry:latest
+            - name: IMAGE_PRUNER
+              value: openshift/cli:latest
```
bparees: this value needs to be `quay.io/openshift/origin-cli:v4.0`

bparees: to match what you put in image-references

coreydaley: updated, thanks

coreydaley:
@adambkaplan @dmage ptal

pkg/resource/generator.go
```diff
@@ -69,6 +69,10 @@ func (g *Generator) list(cr *imageregistryv1.Config) ([]Mutator, error) {
 	mutators = append(mutators, newGeneratorService(g.listers.Services, g.clients.Core, g.params, cr))
 	mutators = append(mutators, newGeneratorDeployment(g.listers.Deployments, g.listers.ConfigMaps, g.listers.Secrets, g.listers.ProxyConfigs, g.clients.Core, g.clients.Apps, driver, g.params, cr))
 	mutators = append(mutators, g.listRoutes(cr)...)
+	mutators = append(mutators, newGeneratorPrunerResource(g.listers.ImagePrunerConfigs, g.clients.RegOp.ImageregistryV1(), g.params))
+	mutators = append(mutators, newGeneratorPrunerClusterRoleBinding(g.listers.ClusterRoleBindings, g.clients.RBAC, g.params))
+	mutators = append(mutators, newGeneratorPrunerServiceAccount(g.listers.ServiceAccounts, g.clients.Core, g.params))
```
adambkaplan: looks good - seems this is in keeping with how we create manifests for the CVO (create RBAC before creating the service account)

pkg/operator/controller.go
```diff
@@ -135,6 +139,81 @@ func (c *Controller) createOrUpdateResources(cr *imageregistryv1.Config) error {
...
+		prunerCronJob, err := c.listers.CronJobs.Get("image-pruner")
+		if errors.IsNotFound(err) {
+			prunerCronJob = nil
+		} else if err != nil {
+			return fmt.Errorf("failed to get image-pruner pruner job: %s", err)
+		} else {
+			prunerCronJob = prunerCronJob.DeepCopy()
+		}
+
+		prunerJobs, err := c.listers.Jobs.List(labels.Everything())
```
adambkaplan: List _all_ Jobs in the namespace? In general I don't admins will run their own one-off jobs in image registry namespace, but you never know. We should apply a common label to all Jobs created by the pruner CronJob and filter on that value.

pkg/operator/controller.go
```diff
@@ -135,6 +139,81 @@ func (c *Controller) createOrUpdateResources(cr *imageregistryv1.Config) error {
...
+		if err != nil {
+			return fmt.Errorf("failed to get pruner jobs: %s", err)
+		}
+		lastPrunerJobConditions := []batchv1.JobCondition{}
+		if len(prunerJobs) > 0 {
+			lastPrunerJobConditions = prunerJobs[len(prunerJobs)-1].Status.Conditions
```
adambkaplan: Are we guaranteed to get the list of `Jobs` in their order of creation, or some other value?

pkg/resource/prunercronjob.go
```diff
@@ -0,0 +1,226 @@
...
+)
+
+var (
+	defaultSuspend                          = false
+	defaultSchedule                         = "*/1 * * * *"
+	defaultStartingDeadlineSeconds    int64 = 60
+	defaultFailedJobsHistoryLimit     int32 = 3
+	defaultSuccessfulJobsHistoryLimit int32 = 3
+	defaultKeepTagRevisions                 = 3
+	defaultKeepYoungerThan                  = "60m"
```
adambkaplan: nit - `const` for things that are primitives? Or does this break the requirement for pointers?

pkg/resource/prunerresource.go
```diff
@@ -0,0 +1,96 @@
...
+
+	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
+	imageregistryv1clientset "github.com/openshift/client-go/imageregistry/clientset/versioned/typed/imageregistry/v1"
+	imageregistryv1listers "github.com/openshift/client-go/imageregistry/listers/imageregistry/v1"
+	"github.com/openshift/cluster-image-registry-operator/defaults"
+	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
+)
+
+var (
+	defaultPrunerSuspend                    = false
```
adambkaplan: ditto - does using `const` instead of `var` break the pointer requirement?

pkg/resource/prunercronjob.go
coreydaley: Correct, I was unable to use a constant here.

pkg/resource/prunerresource.go
coreydaley: Correct, I was unable to use a constant here.

pkg/operator/controller.go
coreydaley: Since the cronjob creates the job, it looks like we are not able to set a label on it when it is created.  I will check if the cronjob sets a label on it already.

adambkaplan: I would assume that's something that can be set in the `CronJob` spec, in a similar fashion to labels in a `Deployment`

coreydaley: updated in next push

manifests/02-rbac.yaml
```diff
@@ -37,6 +40,14 @@ rules:
   - pods
   - secrets
   - services
+  - cronjobs
```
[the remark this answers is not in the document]
coreydaley: fixed

adambkaplan:
@dmage I think we should address that in a follow-up PR. I have a PR that builds on this, and does a quite a bit of refactoring so that we can run multiple controllers within the same binary.

https://github.com/openshift/cluster-image-registry-operator/pull/447

pkg/resource/prunercronjob.go
```diff
@@ -0,0 +1,235 @@
...
+
+	imageregistryapiv1 "github.com/openshift/api/imageregistry/v1"
+	imageregistryv1listers "github.com/openshift/client-go/imageregistry/listers/imageregistry/v1"
+	"github.com/openshift/cluster-image-registry-operator/defaults"
+	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
+)
+
+var (
+	defaultSuspend                          = false
+	defaultSchedule                         = "*/1 * * * *"
```
adambkaplan: This is "run every minute" - needs to be changed to something more sensible like `0 0 * * *` (daily at midnight)

coreydaley: fixed in current push (i can read cronjob syntax :) )

pkg/operator/controller.go
coreydaley: I'm not sure, i don't see anything in the docs about what order they are in.  I thought we ran into this somewhere else in the code also, let me see if i can find it.

coreydaley: Ah, I think it was pruning builds.  Let me go look there.

dmage:
@adambkaplan I'm ok to have several iterations.

/approve

pkg/operator/controller.go
coreydaley: Added a sort by creation timestamp in current push.

coreydaley:
@adambkaplan @dmage ptal, I have addressed all feedback

dmage:
We need a separate controller for sure, it will make the operator more reliable (problems in one component shouldn't affect another component).

coreydaley:
/retest

level=error msg="Error: ResourceLimitExceeded: Task limit exceeded"

coreydaley:
You can do it!

PR #420, opened 2019-12-18
Encode environment variables as YAML

dmage:
Previous attempt #391.
Fixes openshift/installer#2392.

### DIFF
--- a/pkg/envvar/envvar.go
+++ b/pkg/envvar/envvar.go
@@ -1,0 +1,81 @@
+package envvar
+
+import (
+	"fmt"
+	"strings"
+
+	"gopkg.in/yaml.v2"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// EnvVar represents a value for a Distribution configuration parameter.
+type EnvVar struct {
+	// Name is the environment name for the parameter.
+	Name string
+	// Value is the value of the parameter.
+	Value interface{}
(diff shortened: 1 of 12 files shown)


pkg/envvar/envvar.go
```diff
@@ -0,0 +1,81 @@
...
+				return nil, err
+			}
+			envvar.Value = value
+		}
+		envvars = append(envvars, envvar)
+	}
+	return envvars, nil
+}
+
+// Secret returns a data for the secret that should be used with the EnvVars
```
ricardomaraschini: `comment on exported method List.SecretData should be of the form "SecretData ..."`

pkg/resource/secret.go
```diff
@@ -61,11 +61,25 @@ func (gs *generatorSecret) expected() (runtime.Object, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
+	data, err := configenv.SecretData()
+	if err != nil {
+		return nil, err
+	}
+
+	_, _, volumesData, err := gs.driver.Volumes()
```
ricardomaraschini: I would still prefer to have the old `Secrets()` function instead of adding a forth return(`volumeData`) to `Volumes()`. What do you reckon?

dmage:
@ricardomaraschini all comments have been addressed, ptal

PR #421, opened 2019-12-19
Bug 1785115: tolerate all NoSchedule taints

dmage:
/assign @adambkaplan

### DIFF
--- a/pkg/resource/nodecadaemon.go
+++ b/pkg/resource/nodecadaemon.go
@@ -36,7 +36,6 @@ spec:
       priorityClassName: system-cluster-critical
       tolerations:
       - effect: NoSchedule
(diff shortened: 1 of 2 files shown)


dmage:
@adambkaplan do you want a unit test that checks that daemonset has a pod template spec with the same tolerations as we have it in the template?

adambkaplan:
> do you want a unit test that checks that daemonset has a pod template spec with the same tolerations as we have it in the template?

@dmage works for me. Per the kube docs, removing the key means the CA daemonset should schedule on every node [1]

[1] https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/

dmage:
@adambkaplan ptal

PR #423, opened 2020-01-02
Remove Gopkg files

dmage:
We use go modules for this repository. These files are leftovers.


dmage:
This is a follow up to our scrum.

/assign @ricardomaraschini

bparees:
trivial cleanup, non-shipping, no BZ needed.

PR #427, opened 2020-01-07
DEVEXP-433: Report image pruner's install status in Prometheus

adambkaplan:
* Add a new Gauge metric which reports the installation status of the
automatic impage pruner.
* Alert if the image pruner is not installed and enabled
* TEMP: report that the image pruner has been installed for CI to pass

### DIFF
--- a/manifests/09-prometheus-rules.yaml
+++ b/manifests/09-prometheus-rules.yaml
@@ -28,3 +28,18 @@ spec:
           DeploymentConfigs which reference ImageStreamTags may not work as
           expected. Please configure storage and update the config to Managed
           state by editing configs.imageregistry.operator.openshift.io.
+  - name: ImagePruner
+    rules:
+    - alert: ImagePruningDisabled
+      expr: image_registry_operator_image_pruner_install_status < 2
+      labels:
+        severity: warning
+      annotations:
+        message: |
+          Automatic image pruning is not enabled. Regular pruning of images
+          no longer referenced by ImageStreams is strongly recommended to
+          ensure your cluster remains healthy.
+          
+          To remove this warning, install the image pruner by creating an
+          imagepruner.imageregistry.operator.openshift.io resource with the
+          name `cluster`. Ensure that the `suspend` field is set to `false`.
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -5,15 +5,18 @@ import "github.com/prometheus/client_golang/prometheus"
 var (
 	registry            = prometheus.NewRegistry()
 	storageReconfigured = prometheus.NewCounter(prometheus.CounterOpts{
-		Namespace: "image_registry_operator",
-		Subsystem: "storage",
-		Name:      "reconfigured_total",
-		Help:      "Total times the image registry's storage was reconfigured.",
+		Name: "image_registry_operator_storage_reconfigured_total",
+		Help: "Total times the image registry's storage was reconfigured.",
+	})
+	imagePrunerInstallStatus = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "image_registry_operator_image_pruner_install_status",
+		Help: "Installation status code related to the automatic image pruning feature. 0 = not installed, 1 = suspended, 2 = enabled",
 	})
 )
 
 func init() {
 	registry.MustRegister(
 		storageReconfigured,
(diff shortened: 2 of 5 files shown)


adambkaplan:
@openshift/openshift-team-monitoring PTAL. We want to send the `install_info` metric to Telemeter.

manifests/09-prometheus-rules.yaml
```diff
@@ -28,3 +28,18 @@ spec:
           DeploymentConfigs which reference ImageStreamTags may not work as
           expected. Please configure storage and update the config to Managed
           state by editing configs.imageregistry.operator.openshift.io.
+  - name: ImagePruner
+    rules:
+    - alert: ImagePruningDisabled
+      expr: image_registry_operator_image_pruner_install_info{} < 2
```
metalmatze: You don't need the empty `{}`. 
```suggestion
      expr: image_registry_operator_image_pruner_install_info < 2
```

pkg/metrics/metrics.go
```diff
@@ -10,10 +10,17 @@ var (
 		Name:      "reconfigured_total",
 		Help:      "Total times the image registry's storage was reconfigured.",
 	})
+	imagePrunerInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "image_registry_operator",
+		Subsystem: "image_pruner",
+		Name:      "install_info",
```
metalmatze: We tend to simply do the entire thing in name these days. I think the other fields might even be deprecated going forward.
`Name: image_registry_operator_image_pruner_install_info`.

pkg/metrics/metrics.go
```diff
@@ -10,10 +10,17 @@ var (
...
+		Help:      "Installation status code related to the automatic image pruning feature. 0 = not installed, 1 = suspended, 2 = enabled",
+	}, []string{})
```
brancz: There is no need for this to be a GaugeVec if there are no labels, and the `WithLabelValues` call is also not needed in that case.

brancz:
Judging by the code I see, this alert will always fire as we always set the metric to not enabled and not installed. I assume this is currently just done to ensure that this actually fails e2e tests when set that way?

manifests/09-prometheus-rules.yaml
```diff
@@ -28,3 +28,18 @@ spec:
...
+      labels:
+        severity: warning
+      annotations:
+        message: |
+          Automatic image pruning is not enabled. Regular pruning of images
+          no longer referenced by ImageStreams is strongly recommened to ensure
```
simonpasquier: `recommened` -> `recommended`

adambkaplan:
@brancz the current code that sets the metric & fires the alert is purely for test purposes. On new clusters the image pruner will be installed and we shouldn't get the alert.

Upgraded clusters are the more interesting case - they won't automatically get the pruner. In this case we _want_ the alert fired so admins are prompted to install it afterwards.

brancz:
Ack, that makes sense. In that case once this is ready from your side I'm happy with it.

manifests/09-prometheus-rules.yaml
```diff
@@ -28,3 +28,18 @@ spec:
           DeploymentConfigs which reference ImageStreamTags may not work as
           expected. Please configure storage and update the config to Managed
           state by editing configs.imageregistry.operator.openshift.io.
+  - name: ImagePruner
+    rules:
+    - alert: ImagePruningDisabled
+      expr: image_registry_operator_image_pruner_install_info < 2
```
adambkaplan: @brancz @metalmatze @simonpasquier perhaps `install_status` is a better metric name? This metric sits in between "metadata" info and a real value.

brancz: Yeah that works as well. It’s true that info is really meant for constant value time series.

adambkaplan:
/assign @coreydaley 

We can land this as is to stage our alert that pruning is not installed and/or enabled.

adambkaplan:
/retest

AWS rate limit exceeded

adambkaplan:
/assign @dmage 

/cc @coreydaley 

Tests are green

adambkaplan:
/hold

Upgrade tests are getting bit by https://bugzilla.redhat.com/show_bug.cgi?id=1793051

PR #428, opened 2020-01-09
Start using rolloutStrategy parameter

Fedosin:
This parameter allows to define rollout strategy for the image registry deployment.
Now support two possible rollout strategies: RollingUpdate (default) and Recreate.

The parameter was introduced by https://github.com/openshift/api/pull/560

### DIFF
--- a/pkg/resource/deployment.go
+++ b/pkg/resource/deployment.go
@@ -82,6 +82,12 @@ func (gd *generatorDeployment) expected() (runtime.Object, error) {
 	}
 	podTemplateSpec.Annotations[parameters.ChecksumOperatorDepsAnnotation] = depsChecksum
 
+	// Strategy defaults to RollingUpdate
+	strategy := gd.cr.Spec.RolloutStrategy
+	if strategy == "" {
+		strategy = string(appsapi.RollingUpdateDeploymentStrategyType)
+	}
+
 	deploy := &appsapi.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      gd.GetName(),
@@ -97,6 +103,9 @@ func (gd *generatorDeployment) expected() (runtime.Object, error) {
 				MatchLabels: gd.params.Deployment.Labels,
 			},
 			Template: podTemplateSpec,
+			Strategy: appsapi.DeploymentStrategy{
+				Type: appsapi.DeploymentStrategyType(strategy),
+			},
 		},
 	}
 
--- a/test/e2e/configuration_test.go
+++ b/test/e2e/configuration_test.go
@@ -19,6 +19,7 @@ import (
 	configapiv1 "github.com/openshift/api/config/v1"
 	imageregistryapiv1 "github.com/openshift/api/imageregistry/v1"
 	operatorapiv1 "github.com/openshift/api/operator/v1"
+	appsapi "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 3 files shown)


pkg/resource/deployment.go
```diff
@@ -82,6 +83,17 @@ func (gd *generatorDeployment) expected() (runtime.Object, error) {
...
 
+	strategy := gd.cr.Spec.RolloutStrategy
+	// We support only two possible rollout strategies: RollingUpdate and Recreate
+	if strategy != "RollingUpdate" && strategy != "Recreate" {
+		return nil, fmt.Errorf("Unsupported rollout strategy type: %v", strategy)
+	}
+
+	// if we use Recreate rollout strategy, then we must have only one image registry replica
+	if gd.cr.Spec.RolloutStrategy == "Recreate" && gd.cr.Spec.Replicas != 1 {
+		return nil, fmt.Errorf("Not possible to use more than 1 replica with Recreate rollout strategy")
```
bparees: isn't this going to start failing existing clusters?  tightening validation rules is very problematic.

also why can't you use recreate with 1 replica?  (yes you'll have an outage, but it's technically possible/allowed).

Fedosin: Agree, we need one replica when we want to upgrade, not here.
I'll update the patch in a minute.

adambkaplan:
/hold

@coreydaley also needs an api (and client-go) bump to get his auto-pruner feature to land. We need to determine which goes first, or if one of creates a separate api/client-go bump PR.

/cc @dmage

Fedosin:
I suggest to create a separate PR with bump of api and client-go, and just keep the code in this one

adambkaplan:
@Fedosin looks like the API validation is breaking all of our tests. This will without question break customers on upgrade.

Fedosin:
https://github.com/openshift/cluster-image-registry-operator/pull/432 should be merged first

test/e2e/configuration_test.go
```diff
@@ -89,6 +90,65 @@ func TestPodResourceConfiguration(t *testing.T) {
...
+					Effect:   "NoSchedule",
+				},
+			},
+		},
+	}
+	framework.MustDeployImageRegistry(t, client, cr)
+	framework.MustEnsureImageRegistryIsAvailable(t, client)
+	framework.MustEnsureClusterOperatorStatusIsNormal(t, client)
+
+	deployments, err := client.Deployments(defaults.ImageRegistryOperatorNamespace).List(metav1.ListOptions{})
```
dmage: There is no need to list all deployments. We know the exact name of the deployment `image-registry`.

You can use something like

```go
    registryDeployment, err := client.Deployments(defaults.ImageRegistryOperatorNamespace).Get(defaults.ImageRegistryName, metav1.GetOptions{})
    if err != nil {
        t.Fatal(err)
    }
```

manifests/registry-config.yaml
```diff
@@ -16,3 +16,4 @@ spec:
       volumeSource:
         emptyDir: {}
   replicas: 1
+  rolloutStrategy: RollingUpdate
```
dmage: We don't use this file and I'm going to remove it. It's better to keep this file as is to avoid possible conflicts between PRs.

test/e2e/configuration_test.go
```diff
@@ -89,6 +90,65 @@ func TestPodResourceConfiguration(t *testing.T) {
 	}
 }
 
+func TestRolloutStrategyConfiguration(t *testing.T) {
+	client := framework.MustNewClientset(t, nil)
+
+	defer framework.MustRemoveImageRegistry(t, client)
+
+	cr := &imageregistryapiv1.Config{
+		TypeMeta: metav1.TypeMeta{
```
dmage: It's better to keep TypeMeta unset.

test/e2e/configuration_test.go
Fedosin: done

manifests/registry-config.yaml
Fedosin: done

test/e2e/configuration_test.go
Fedosin: done

PR #434, opened 2020-01-13
DEVEXP-497: enable dualstack endpoints for S3

dmage:


pkg/storage/s3/s3.go
```diff
@@ -193,6 +193,7 @@ func (d *driver) ConfigEnv() (envs envvar.List, err error) {
 		envvar.EnvVar{Name: "REGISTRY_STORAGE_S3_BUCKET", Value: d.Config.Bucket},
 		envvar.EnvVar{Name: "REGISTRY_STORAGE_S3_REGION", Value: d.Config.Region},
 		envvar.EnvVar{Name: "REGISTRY_STORAGE_S3_ENCRYPT", Value: d.Config.Encrypt},
+		envvar.EnvVar{Name: "REGISTRY_STORAGE_S3_USEDUALSTACK", Value: true},
```
adambkaplan: @dmage since we need to backport this change, can you please add a unit test or e2e which verifies this env var is set?

dmage: @adambkaplan I don't see any reasons to write a test for this, but okay

adambkaplan: @dmage tbh I would have likely been OK leaving as-is if this was only needed for 4.4. However, since we need to back-port this feature the patch managers will likely raise concerns if we have feature code landing without some form of verification.

PR #439, opened 2020-01-15
DEVEXP-523: Allow to use RWO access mode for PVC storage type

Fedosin:
We allow using RWO PV backend, but it has some limitations:
1. Image registry rollout strategy must be set to Recreate (default is RollingUpdate).
2. It's not possible to use more than 1 replica of the image registry.

### DIFF
--- a/pkg/storage/pvc/pvc.go
+++ b/pkg/storage/pvc/pvc.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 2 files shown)


Fedosin:
/hold
we need https://github.com/openshift/cluster-image-registry-operator/pull/428 to be merged first

pkg/storage/pvc/pvc.go
```diff
@@ -116,13 +117,36 @@ func (d *driver) checkPVC(cr *imageregistryv1.Config, claim *corev1.PersistentVo
...
 
-	for _, claimMode := range claim.Spec.AccessModes {
+	// Check what access modes are available.
+
+	// We allow using RWO PV backend, but it has some limitations:
+	// 1. Image registry rollout strategy must be set to Recreate (default is RollingUpdate).
+	// 2. It's not possible to use more than 1 replica of the image registry.
+
+	// RWX backends are accepted with no additional conditions.
+	RWOModeEnabled := false
```
adambkaplan: nit - start lowercase. `rwoModeEnabled`

pkg/storage/pvc/pvc.go
```diff
@@ -116,13 +117,36 @@ func (d *driver) checkPVC(cr *imageregistryv1.Config, claim *corev1.PersistentVo
...
 			return nil
 		}
+		if claimMode == corev1.ReadWriteOnce {
+			RWOModeEnabled = true
+		}
+	}
+
+	if RWOModeEnabled {
+		if cr.Spec.Replicas > 1 {
+			return fmt.Errorf("Cannot use %s access mode with more than one replica of the image registry", corev1.ReadWriteOnce)
```
adambkaplan: nit - start error message with lowercase.
```suggestion
			return fmt.Errorf("cannot use %s access mode with more than one replica of the image registry", corev1.ReadWriteOnce)
```

pkg/storage/pvc/pvc.go
```diff
@@ -116,13 +117,36 @@ func (d *driver) checkPVC(cr *imageregistryv1.Config, claim *corev1.PersistentVo
...
+		}
+
+		if cr.Spec.RolloutStrategy != string(appsv1.RecreateDeploymentStrategyType) {
+			return fmt.Errorf("Cannot use %s access mode with %s rollout strategy", corev1.ReadWriteOnce, cr.Spec.RolloutStrategy)
```
adambkaplan: nit - start with lowercase
```suggestion
			return fmt.Errorf("cannot use %s access mode with %s rollout strategy", corev1.ReadWriteOnce, cr.Spec.RolloutStrategy)
```

Fedosin: done

pkg/storage/pvc/pvc.go
Fedosin: done

pkg/storage/pvc/pvc.go
Fedosin: done

pkg/storage/pvc/pvc.go
```diff
@@ -116,13 +117,36 @@ func (d *driver) checkPVC(cr *imageregistryv1.Config, claim *corev1.PersistentVo
...
 	}
 
+	// Check what access modes are available.
+
+	// We allow using RWO PV backend, but it has some limitations:
+	// 1. Image registry rollout strategy must be set to Recreate (default is RollingUpdate).
+	// 2. It's not possible to use more than 1 replica of the image registry.
+
+	// RWX backends are accepted with no additional conditions.
+	rwoModeEnabled := false
```
pweil-: this validation looks reasonable to me, thanks for the update.  

For other reviewers, if RWX is **anywhere** in the list the volume will be mounted so multiple pods can use it.  If RWO is the **only** thing in the list then it will be mounted so only a single pod can use it.  If both exist the volume is mounted to support both so it falls into the first use case.

Fedosin: thanks :) everything you said is correct

PR #441, opened 2020-01-17
Swiftless OpenStack

Fedosin:
This PR allows to use RWO PVC backend for OpenStack if Swift service is not available.

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -7,12 +7,18 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/defaults"
 	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
 	"github.com/openshift/cluster-image-registry-operator/pkg/storage"
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage/pvc"
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage/swift"
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage/util"
 
+	configapiv1 "github.com/openshift/api/config/v1"
 	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
 	operatorapi "github.com/openshift/api/operator/v1"
 	regopset "github.com/openshift/client-go/imageregistry/clientset/versioned/typed/imageregistry/v1"
 	appsapi "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 )
@@ -61,6 +67,32 @@ func (c *Controller) Bootstrap() error {
 		mgmtState = operatorapi.Removed
 	}
 
+	infra, err := util.GetInfrastructure(c.listers)
+	if err != nil {
+		return err
+	}
+
+	rolloutStrategy := appsapi.RollingUpdateDeploymentStrategyType
+
+	// If Swift service is not available for OpenStack, we have to start using
+	// Cinder with RWO PVC backend. It means that we need to create an RWO claim
+	// and set the rollout strategy to Recreate.
+	switch infra.Status.PlatformStatus.Type {
+	case configapiv1.OpenStackPlatformType:
+		isSwiftEnabled, err := swift.IsSwiftEnabled(c.listers)
+		if err != nil {
+			return err
+		}
+		if !isSwiftEnabled {
+			err = c.createPVC(corev1.ReadWriteOnce)
+			if err != nil {
+				return err
+			}
+
+			rolloutStrategy = appsapi.RecreateDeploymentStrategyType
+		}
+	}
+
 	cr = &imageregistryv1.Config{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:       defaults.ImageRegistryResourceName,
@@ -73,7 +105,7 @@ func (c *Controller) Bootstrap() error {
 			Storage:         imageregistryv1.ImageRegistryConfigStorage{},
 			Replicas:        1,
 			HTTPSecret:      fmt.Sprintf("%x", string(secretBytes[:])),
-			RolloutStrategy: string(appsapi.RollingUpdateDeploymentStrategyType),
+			RolloutStrategy: string(rolloutStrategy),
 		},
 		Status: imageregistryv1.ImageRegistryStatus{},
 	}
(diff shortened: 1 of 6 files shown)


pkg/storage/swift/swift.go
```diff
@@ -51,6 +51,16 @@ func replaceEmpty(a string, b string) string {
 	return a
 }
 
+// IsSwiftEnabled checks if Swift service is available for OpenStack platform
+func IsSwiftEnabled(listers *regopclient.Listers) bool {
+	driver := NewDriver(&imageregistryv1.ImageRegistryConfigStorageSwift{}, listers)
+	_, err := driver.getSwiftClient()
```
ricardomaraschini: Not every error returned by this function indicates that the Swift is disabled, don't you think?

Fedosin: agree with you. I'll update the patch soon

pkg/operator/bootstrap.go
```diff
@@ -61,6 +66,28 @@ func (c *Controller) Bootstrap() error {
...
+
+	rolloutStrategy := appsapi.RollingUpdateDeploymentStrategyType
+
+	// If Swift service is not available for OpenStack, we have to start using
+	// Cinder with RWO PVC backend. It means that we need to create an RWO claim
+	// and set the rollout strategy to Recreate.
+	switch infra.Status.PlatformStatus.Type {
+	case configapiv1.OpenStackPlatformType:
+		if swift.IsSwiftEnabled(c.listers) {
+			err = c.createPVC(corev1.ReadWriteOnce)
```
dmage: Will this function fail if the PVC already exist? In some cases it might be pre-created (a rare scenario - the operator fails to create the custom resource and it runs Bootstrap the second time).

pkg/operator/bootstrap.go
```diff
@@ -61,6 +66,28 @@ func (c *Controller) Bootstrap() error {
...
+	infra, err := util.GetInfrastructure(c.listers)
+	if err != nil {
+		return err
+	}
+
+	rolloutStrategy := appsapi.RollingUpdateDeploymentStrategyType
+
+	// If Swift service is not available for OpenStack, we have to start using
+	// Cinder with RWO PVC backend. It means that we need to create an RWO claim
+	// and set the rollout strategy to Recreate.
```
bparees: how common is this?  is this really the right default for a customer, or would it be better to come up w/ the registry removed and tell the customer to install swift?

tomassedovic: Cinder was quite common for OpenShift 3.x installs, though of course there the choice between Cinder and Swift was a configuration option.

A lot of customers are not running Swift and for them, the deployment will simply fail right now. This and the self-signed certificates are major blockers for a lot people. They have generally not been receptive to installing Swift so far.

I don't the numbers, but it's not something we thought up. There is a very strong demand for not requiring Swift from our customers.

pkg/storage/swift/swift.go
Fedosin: done

pkg/operator/bootstrap.go
Fedosin: yep, I updated to code to handle this situation properly

pkg/operator/bootstrap.go
bparees: ok

pkg/storage/pvc/pvc.go
```diff
@@ -24,7 +24,10 @@ import (
 const (
 	rootDirectory      = "/registry"
 	randomSecretSize   = 32
-	pvcOwnerAnnotation = "imageregistry.openshift.io"
+	PVCOwnerAnnotation = "imageregistry.openshift.io"
+
+	// PVCImageRegistryName is the defailt name of the claim provisioned for PVC backend
+	PVCImageRegistryName = defaults.ImageRegistryName + "-storage"
```
dmage: Let's move this constant to `defaults`. And in that package it's better to have something like `PVCImageRegistryName = "image-registry-storage"` to make it greppable.

pkg/operator/bootstrap.go
```diff
@@ -94,3 +126,50 @@ func (c *Controller) Bootstrap() error {
...
+	if err == nil {
+		if !pvc.PVCIsCreatedByOperator(claim) {
+			err = fmt.Errorf("could not create this PVC, it already exists and is not owned by the operator")
+			return err
+		}
+		// PVC is already created, no bootstrapping is required
+		return nil
+	} else if errors.IsNotFound(err) {
+		// if PVC with this name doesn't exist, create new PVC
+		storageClassName := "standard"
```
dmage: Is it a special name for our OpenShift on OpenStack?

pkg/storage/pvc/pvc.go
```diff
@@ -24,7 +24,10 @@ import (
 const (
 	rootDirectory      = "/registry"
 	randomSecretSize   = 32
-	pvcOwnerAnnotation = "imageregistry.openshift.io"
+	PVCOwnerAnnotation = "imageregistry.openshift.io"
+
+	// PVCImageRegistryName is the defailt name of the claim provisioned for PVC backend
```
dmage: -defailt
+default

pkg/operator/bootstrap.go
Fedosin: yes, it's kinda hardcoded in kubernetes: https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/storage-class/openstack/default.yaml#L4

pkg/storage/pvc/pvc.go
Fedosin: makes sense! done

pkg/operator/bootstrap.go
```diff
@@ -94,3 +126,50 @@ func (c *Controller) Bootstrap() error {
...
 }
+
+func (c *Controller) createPVC(accessMode corev1.PersistentVolumeAccessMode) error {
+	claimName := pvc.PVCImageRegistryName
+
+	// Check that the claim does not exist before creating it.
+	claim, err := c.clients.Core.PersistentVolumeClaims(defaults.ImageRegistryOperatorNamespace).Get(claimName, metav1.GetOptions{})
+	if err == nil {
+		if !pvc.PVCIsCreatedByOperator(claim) {
+			err = fmt.Errorf("could not create this PVC, it already exists and is not owned by the operator")
```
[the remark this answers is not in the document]
Fedosin: yep, removed it

pkg/storage/pvc/pvc.go
Fedosin: done :)

Fedosin:
@dmage I added a unit test to check that If there is no Swift service, the `ErrEndpointNotFound` error is returned. It's hard to check more, because we need to upgrade our CI for this and create `e2e-openstack-operator` job.
Meanwhile I use https://github.com/openshift/cluster-image-registry-operator/pull/433 to explicitly disable Swift and verify this change on RWO PVC

dmage:
/lgtm
/hold

Holding until we solve problems with upgrades

Fedosin:
e2e-aws-upgrade is green!

PR #443, opened 2020-01-20
Dump related resources when DeployImageRegistry fails

dmage:


dmage:
/assign @ricardomaraschini 

I can't reproduce the "config already exists" problem, let's merge it and see if someone else hit the problem.

PR #445, opened 2020-01-21
Add support for OpenStack clouds with self-signed certificates

mandre:
In the case where OpenStack in deployed with self-signed certificates, we need to initiate the connection to swift using a custom CA certificate.

Let's look for the `ca-bundle.pem` entry in the `openshift-config/cloud-provider-config` config map and use it to initiate the connection if it is provided.

Related to https://github.com/openshift/installer/pull/2932

### DIFF
--- a/pkg/resource/caconfig.go
+++ b/pkg/resource/caconfig.go
@@ -118,6 +118,17 @@ func (gcac *generatorCAConfig) expected() (runtime.Object, error) {
 		}
 	}
 
+	cp_ca, err := gcac.openshiftConfigLister.Get("cloud-provider-config")
+	if errors.IsNotFound(err) {
+		klog.Infof("missing the cloud-provider-config configmap: %s", err)
+	} else if err != nil {
+		return cm, err
+	} else {
+		if cert, ok := cp_ca.Data["ca-bundle.pem"]; ok {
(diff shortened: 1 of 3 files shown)


pkg/resource/caconfig.go
```diff
@@ -118,6 +118,17 @@ func (gcac *generatorCAConfig) expected() (runtime.Object, error) {
...
 	}
 
+	cp_ca, err := gcac.openshiftConfigLister.Get("cloud-provider-config")
+	if errors.IsNotFound(err) {
+		klog.Infof("missing the cloud-provider-config configmap: %s", err)
+	} else if err != nil {
+		return cm, err
+	} else {
+		if cert, ok := cp_ca.Data["ca-bundle.pem"]; ok {
+			cm.Data["cloud-provider-ca-bundle.pem"] = cert
```
dmage: This config map is used by the nodeca daemon and is distributed to all nodes to /etc/docker/certs.d. Therefore its keys should be host names.

Perhaps you need yet another config map to this CA.

mandre: Let's track this in https://bugzilla.redhat.com/show_bug.cgi?id=1794690

PR #446, opened 2020-01-22
Bug 1791934: independent controller for clusteroperators

dmage:


pkg/operator/clusteroperator.go
```diff
@@ -0,0 +1,133 @@
...
+	configv1listers "github.com/openshift/client-go/config/listers/config/v1"
+	imageregistryv1informers "github.com/openshift/client-go/imageregistry/informers/externalversions/imageregistry/v1"
+	imageregistryv1listers "github.com/openshift/client-go/imageregistry/listers/imageregistry/v1"
+
+	"github.com/openshift/cluster-image-registry-operator/defaults"
+	regopclient "github.com/openshift/cluster-image-registry-operator/pkg/client"
+	"github.com/openshift/cluster-image-registry-operator/pkg/resource"
+)
+
+const workQueueKey = "instance"
```
adambkaplan: May want to name this const to be more specific. I may have a similar need for a constant queue key in my metrics PR

pkg/operator/controller.go
```diff
@@ -470,13 +463,23 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
...
+
 	configInformerFactory.Start(stopCh)
 	kubeInformerFactory.Start(stopCh)
 	openshiftConfigKubeInformerFactory.Start(stopCh)
 	kubeSystemKubeInformerFactory.Start(stopCh)
 	regopInformerFactory.Start(stopCh)
 	routeInformerFactory.Start(stopCh)
 
+	// TODO(dmage): This controller should be started from main.
+	go clusterOperatorStatusController.Run(stopCh)
```
adambkaplan: Or from a separate "starter" struct that is effectively main. See https://github.com/openshift/cluster-image-registry-operator/pull/447/commits/eb53b988d47e8a8acfc7f5d00b73bb0fbb19b515

dmage: Sure. I'd prefer to avoid a struct with one method and have just a function. But it doesn't really matter for me.

dmage:
@adambkaplan this should fix upgrade problems, so we'll need to backport it

dmage:
@adambkaplan teardown issues should be fixed by #452.

PR #449, opened 2020-01-24
Report image pruner install status correctly.

adambkaplan:
Report the image pruner's install metric when we sync status.


adambkaplan:
Note that origin has examples on how to query Prometheus in an e2e, but it is not simple.

PR #450, opened 2020-01-27
Bug 1776665: Don't query storage provider for existence on resync

coreydaley:


dmage:
@coreydaley the test failure doesn't look like a flake

> condition StorageExists reason should have been \"S3 Bucket Exists\" but was Creation Successful instead

adambkaplan:
per group D arch discussion - if the storage is created by the operator, but is then removed from underneath the operator, the operator should report itself `Degraded` and not try to re-create the storage.

In the event storage is lost, cluster admins will need to intervene and restore data from a backup.

adambkaplan:
This difficulties in this PR make me wonder if we should refactor to patterns closer to the library-go operators (ex cluster-openshift-apiserver-operator). These tend to be built as follows:

1. One or more controllers which observe the configuration state and write the result to a separate object - like a YAML file in a `ConfigMap`.
2. A sync controller which watches said `ConfigMap` and applies changes to the operands
3. A status controller which observes the operands and reports status up to the `ClusterOperator` object.

coreydaley:
We will still need to query the storage provider x often to make sure that the storage is still there and the settings are still correct for it. Thoughts on that?

adambkaplan:
> We will still need to query the storage provider x often to make sure that the storage is still there and the settings are still correct for it. Thoughts on that?

A controller whose informers have a relist interval of x will give us that. It probably makes sense for the "status" controller to check if the storage exists and report degraded state. IMO a separate "config observer" controller would be responsible for checking if the desired storage configs change (bucket, creds, etc.)

coreydaley:
closing and will create a new one once we move to multiple controllers

PR #451, opened 2020-01-27
Bug 1769927: Persistent Deployment

ricardomaraschini:
This patch makes use of library-go ApplyDeployment() function to stomp over user provided changes. This way we ensure that image registry Deployment is always on the state we need it to be.


pkg/resource/deployment.go
```diff
@@ -104,11 +110,17 @@ func (gd *generatorDeployment) expected() (runtime.Object, error) {
...
+				Type: appsapi.DeploymentStrategyType(deployStrategy),
 			},
 		},
 	}
 
+	dgst, err := strategy.Checksum(deploy)
+	if err != nil {
+		return nil, err
+	}
+	deploy.ObjectMeta.Annotations[parameters.ChecksumOperatorAnnotation] = dgst
```
dmage: Do you need this only for backward compatibility?

pkg/resource/deployment.go
```diff
@@ -117,15 +129,77 @@ func (gd *generatorDeployment) Get() (runtime.Object, error) {
...
+
+	return dep, updated, nil
+}
+
+func (gd *generatorDeployment) UpdateLastGeneration(lastGen int64) {
+	for i, gen := range gd.cr.Status.Generations {
+		if gen.Name == defaults.ImageRegistryName &&
+			gen.Group == gd.GetGroup() &&
+			gen.Resource == gd.GetResource() &&
+			gen.Namespace == defaults.ImageRegistryOperatorNamespace {
```
dmage: gd.GetName(), gd.GetNamespace()

ricardomaraschini: nice catch.

pkg/resource/deployment.go
ricardomaraschini: No, we need this because `resourceapply.ApplyDeployment()`  takes into account only `ObjectMeta` when [detecting changes](https://github.com/openshift/library-go/blob/master/pkg/operator/resource/resourceapply/apps.go#L36).

dmage:
@ricardomaraschini why do we need to bump dependencies in this PR?

dmage:
@ricardomaraschini can you write an e2e test that patches the registry deployment and checks that this change is stomped?

dmage:
/approve
otherwise it looks good

ricardomaraschini:
> @ricardomaraschini why do we need to bump dependencies in this PR?

It happened when i added the new dependency on `resourceapply` package. I noticed it updated other stuff but I thought that it would not be bad to have them updated if possible.

ricardomaraschini:
> @ricardomaraschini can you write an e2e test that patches the registry deployment and checks that this change is stomped?

Yes.

ricardomaraschini:
@dmage test implemented, PTAL.

test/e2e/recreate_deployment_test.go
```diff
@@ -54,3 +56,57 @@ func TestRecreateDeployment(t *testing.T) {
...
+				"path": "/spec/template/spec/containers/0/ports/-",
+				"value": {"name": "foo", "containerPort": 2222}
+			}
+		]`),
+	); err != nil {
+		t.Fatalf("unable to patch image registry deployment: %v", err)
+	}
+
+	// give some room for the operator to restore the deployment.
+	time.Sleep(10 * time.Second)
```
dmage: It's better to watch the resource. Or at least to poll. You never know how slow the CI machine can be.

ricardomaraschini: I added a Poll here, PTAL.

PR #452, opened 2020-01-28
Use Recreate strategy for operator's deployment

dmage:
When the operator's deployment gets changed, we don't want to have two instances of the operator at the same time: 1 replica (old) -> 2 replicas (old and new) -> 1 replica (new). As we don't have leader election and we want to be more predictable behaviour in our e2e tests, it's better to upgrade through scaling down: 1 replica (old) -> 0 replicas -> 1 replica (new).

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -8,6 +8,8 @@ metadata:
     config.openshift.io/inject-proxy: cluster-image-registry-operator 
 spec:
   replicas: 1
+  strategy:
+    type: Recreate
   selector:
     matchLabels:
       name: cluster-image-registry-operator
--- a/test/framework/imageregistry.go
+++ b/test/framework/imageregistry.go
@@ -153,17 +153,29 @@ func deleteImageRegistryResource(client *Clientset) error {
 }
 
 func RemoveImageRegistry(logger Logger, client *Clientset) error {
+	logger.Logf("checking if the operator has only one replica...")
+	pods, err := getOperatorPods(client)
+	if err != nil {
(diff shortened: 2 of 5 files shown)


adambkaplan:
/hold

Are we better off adding leader election to this operator? My worry is that during upgrade the registry is effectively unmanaged.

dmage:
@adambkaplan while I agree with your concern, this PR is intended to fix flakes in our tests, leader election is hard to do right (the operator should stop reconciliation immediately when it's no longer a leader), and other operators (cluster-openshift-apiserver-operator and cluster-kube-apiserver-operator) also use Recreate strategy.

dmage:
failed to wait for bootstrapping to complete
/retest

dmage:
@adambkaplan are you ok with implementing leader election as a follow-up? The way I'd prefer to do it is to merge our refactoring and then migrate to https://github.com/openshift/library-go/tree/master/pkg/controller/controllercmd.

And even if we implement leader election, I still prefer to have 0 replicas for a few seconds during upgrade rather than 2 concurrent replicas. It's much more predictable - we can be sure that if the new replica is alive, the old one is dead and never become active anymore.

ricardomaraschini:
@dmage Is there any hope for this one? I am OK with having leader election deferred to a JIRA card.

adambkaplan:
@ricardomaraschini @dmage fwiw adding leader election absolutely falls into a "stabilize the product" feature.

dmage:
I moved IR-16 (adding leader election) to the top of our backlog. Let's not merge this, after IR-16 we'll need to rollout it anyway.

PR #458, opened 2020-02-17
TestEnv wrapper around testing.T

dmage:
This PR introduces `framework.TestEnv`, a context for test functions.

It carries clients and testing.T, eventually we may want it to implement the context.Context interface. It is supposed to be the first argument of almost every function in the package test/framework. Any function with this context can fail the test (using te.Errorf or te.Fatalf), log additional information (te.Logf), and use k8s/okd clients.

The main idea of this PR is to make our tests more declarative. We usually don't have a fancy error handlers in our e2e tests, they usually look like this:

```
if err != nil {
  return err
}
...
if err != nil {
  t.Fatal(err)
}
```

It's an anti-pattern for the regular code, errors should be handled more carefully and the code should add more context details before returning the error.

But it's OK for tests. Therefore we can use Fatalf inside the helper: something went wrong, there is no reason to continue the test. All these `return err` lines just add visual noise and distract from the test.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -286,6 +286,9 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
 		operatorDegraded.Status = operatorapiv1.ConditionTrue
 		operatorDegraded.Message = applyError.Error()
 		operatorDegraded.Reason = e.Reason
+	} else if cr.Spec.ManagementState == operatorapiv1.Removed {
+		operatorDegraded.Message = "The registry is removed"
+		operatorDegraded.Reason = "Removed"
 	}
 
 	updateCondition(cr, operatorapiv1.OperatorStatusTypeDegraded, operatorDegraded)
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -38,6 +38,9 @@ var (
 )
 
 func TestAWSDefaults(t *testing.T) {
+	te := framework.Setup(t)
+	defer framework.TeardownImageRegistry(te)
+
 	kcfg, err := regopclient.GetConfig()
 	if err != nil {
 		t.Fatalf("Error building kubeconfig: %s", err)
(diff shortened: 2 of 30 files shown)


dmage:
@coreydaley can you take a look at TestOperandProxyConfiguration?

gabemontero:
> This PR introduces `framework.TestEnv`, a context for test functions.
> 
> It carries clients and testing.T, eventually we may want it to implement the context.Context interface. It is supposed to be the first argument of almost every function in the package test/framework. Any function with this context can fail the test (using te.Errorf or te.Fatalf), log additional information (te.Logf), and use k8s/okd clients.
> 
> The main idea of this PR is to make our tests more declarative. We usually don't have a fancy error handlers in our e2e tests, they usually look like this:
> 
> ```
> if err != nil {
>   return err
> }
> ...
> if err != nil {
>   t.Fatal(err)
> }
> ```
> 
> It's an anti-pattern for the regular code, errors should be handled more carefully and the code should add more context details before returning the error.
> 
> But it's OK for tests. Therefore we can use Fatalf inside the helper: something went wrong, there is no reason to continue the test. All these `return err` lines just add visual noise and distract from the test.

I really like where you are headed with this @dmage 

Is it correct to think that after fleshing things out here, pieces like ` test/framework/testenv.go` and anything not image registry specific in `test/framework` are things you'll consider moving to library-go or another repo of similar ilk (maybe even a new one) that can be imported by other operator repos?

thanks

dmage:
@gabemontero yes, parts of this framework were already copied to other repos, so I want to focus on creating common parts that are not specific to this repository (wait until a deployment is rolled out, check environment variables, etc.) and try to reuse them for another operator. If they are reusable, we can create another repo.

gabemontero:
> @gabemontero yes, parts of this framework were already copied to other repos, so I want to focus on creating common parts that are not specific to this repository (wait until a deployment is rolled out, check environment variables, etc.) and try to reuse them for another operator. If they are reusable, we can create another repo.

Awesome @dmage 

When the time comes, if you think an additional person can help you move things along let me know and we can see where I help you out.

adambkaplan:
@dmage generally looks good to me - assuming we have a green run all that's left is to squash commits?

dmage:
@adambkaplan why do you want to squash them?

dmage:
Can we tag it?

adambkaplan:
@dmage do we plan on backporting this to 4.4.0? Do we need this to ensure our 4.4 test suite doesn't flake?

dmage:
@adambkaplan no, we don't need to backport it. It just a refactoring to make our tests less verbose, so it's not supposed to affect flakiness (but an additional check in SetupAvailableImageRegistry can give controllers few more milliseconds and make tests less flaky). I want to use it to make our tests green on vSphere: SetupAvailableImageRegistry should skip the test on vSphere when spec.storage isn't set (because we don't have default on this platform).

We can land it after we land all PRs with tests that we want to backport. Meanwhile I can work on a next commit for vSphere in this PR. What do you think?

adambkaplan:
@dmage in the past I've created a new branch from this point and added new commits there. It's easier to submit a follow-up PR that way 😸

dmage:
@adambkaplan this is ready to be merged

dmage:
@adambkaplan 

e2e-azure-operator - we haven't used it [for a while](https://prow.svc.ci.openshift.org/job-history/origin-ci-test/pr-logs/directory/pull-ci-openshift-cluster-image-registry-operator-master-e2e-azure-operator), I created the [bug 1817119](https://bugzilla.redhat.com/show_bug.cgi?id=1817119) to track problems with this job.
e2e-gcp-operator - some problems with the credentials operator, blocked by [CO-845](https://issues.redhat.com/browse/CO-845).
e2e-vsphere - these tests are not specific to our repo, apparently they are [not stable on vsphere yet](https://prow.svc.ci.openshift.org/?job=*-master-e2e-vsphere).

So these jobs will be fixed as a follow-up.

dmage:
@adambkaplan ptal

PR #459, opened 2020-02-17
Bug 1806686: Image pruner updates: Controller / Prune Registry

coreydaley:
- Move the image pruner to it's own controller
- Implement the --prune-registry flag for the pruner


pkg/operator/controllerimagepruner.go
```diff
@@ -0,0 +1,384 @@
+package operator
+
+import (
+	"fmt"
+	"reflect"
+	"sort"
+	"time"
+
+	regopclient "github.com/openshift/cluster-image-registry-operator/pkg/client"
```
dmage: Usually we use this order of packages:

1. standard packages
2. other packages
3. k8s
4. openshift
5. the current repository

pkg/operator/controllerimagepruner.go
```diff
@@ -0,0 +1,384 @@
...
+				klog.Infof("event from image pruner workqueue successfully processed")
+			}
+		}()
+	}
+}
+
+func (c *ImagePrunerController) handler() cache.ResourceEventHandlerFuncs {
+	return cache.ResourceEventHandlerFuncs{
+		AddFunc: func(o interface{}) {
+			if clusterOperator, ok := o.(*configapiv1.ClusterOperator); ok {
```
dmage: this controller doesn't manage ClusterOperators

pkg/resource/prunercronjob.go
```diff
@@ -116,6 +124,7 @@ func (gcj *generatorPrunerCronJob) expected() (runtime.Object, error) {
 										"--certificate-authority=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt",
 										fmt.Sprintf("--keep-tag-revisions=%d", gcj.getKeepTagRevisions(cr)),
 										fmt.Sprintf("--keep-younger-than=%s", gcj.getKeepYoungerThan(cr)),
+										fmt.Sprintf("--prune-registry %t", gcj.getPruneRegistry(rcr)),
```
dmage: `--prune-registry=%t`, to be consistent.

pkg/operator/controllerimagepruner.go
```diff
@@ -0,0 +1,384 @@
...
+	return c, nil
+}
+
+// ImagePrunerController keeps track of openshift image pruner components.
+type ImagePrunerController struct {
+	kubeconfig *restclient.Config
+	params     parameters.Globals
+	generator  *resource.ImagePrunerGenerator
+	workqueue  workqueue.RateLimitingInterface
+	listers    *regopclient.Listers
```
dmage: Are we going to have a lot of `nil`s there?

pkg/operator/controllerimagepruner.go
coreydaley: I know that.  When the operator was refactored to use the openshift/api everything got in the wrong order and I didn't feel like fixing it all, but I will fix this one.

pkg/operator/controllerimagepruner.go
coreydaley: yes, thoughts?

pkg/operator/controllerimagepruner.go
coreydaley: Yep, this was still listed as a WIP and I am doing cleanup.
Updated in next push

pkg/resource/prunercronjob.go
coreydaley: updated in next push

pkg/operator/controllerimagepruner.go
adambkaplan: two ideas:

1. Create a listers struct for each controller
2. Only use the informer factories as references in the controller struct.

coreydaley: Done, created a listers struct for the image pruner controller

coreydaley:
@adambkaplan @dmage ptal

pkg/operator/controllerimagepruner.go
```diff
@@ -0,0 +1,423 @@
...
+			if err != nil {
+				klog.Errorf("unable to get accessor for old object: %s", err)
+				return
+			}
+			if newAccessor.GetResourceVersion() == oldAccessor.GetResourceVersion() {
+				// Periodic resync will send update events for all known resources.
+				// Two different versions of the same resource will always have different RVs.
+				return
+			}
+			if clusterOperator, ok := o.(*configapiv1.ClusterOperator); ok {
```
dmage: ClusterOperators are not watched by this controller

pkg/operator/controllerimagepruner.go
```diff
@@ -0,0 +1,423 @@
...
+func (b ByCreationTimestamp) Swap(i, j int) {
+	b[i], b[j] = b[j], b[i]
+}
+
+func (b ByCreationTimestamp) Less(i, j int) bool {
+	return !b[j].CreationTimestamp.Time.After(b[i].CreationTimestamp.Time)
+}
+
+// Bootstrap registers this operator with OpenShift by creating an appropriate
+// ClusterOperator custom resource. This function also creates the initial
```
dmage: This function doesn't create ClusterOperator objects.

pkg/operator/controllerimagepruner.go
```diff
@@ -0,0 +1,423 @@
...
+	params     parameters.Globals
+	generator  *resource.ImagePrunerGenerator
+	workqueue  workqueue.RateLimitingInterface
+	listers    *regopclient.ImagePrunerControllerListers
+	clients    *regopclient.Clients
+}
+
+func (c *ImagePrunerController) createOrUpdateResources(cr *imageregistryv1.ImagePruner) error {
+	return c.generator.Apply(cr)
+
```
dmage: Just curious, why do you add empty lines before `}`?

pkg/client/listers.go
```diff
@@ -33,3 +33,14 @@ type Listers struct {
 	ProxyConfigs        configlisters.ProxyLister
 	Infrastructures     configlisters.InfrastructureLister
 }
+
+type ImagePrunerControllerListers struct {
+	Jobs                kjoblisters.JobNamespaceLister
+	CronJobs            kbatchlisters.CronJobNamespaceLister
+	ServiceAccounts     kcorelisters.ServiceAccountNamespaceLister
+	Routes              routelisters.RouteNamespaceLister
```
dmage: This lister is not used by the pruner controller.

pkg/operator/controllerimagepruner.go
coreydaley: I deleted a bunch of stuff from that function but missed the newline.

pkg/client/listers.go
coreydaley: thanks, i missed that one.  i think i want to refactor this to share the listers between controllers in a followup pr.

pkg/operator/controllerimagepruner.go
coreydaley: thanks, i thought i got all of those and removed them.

pkg/operator/controllerimagepruner.go
coreydaley: thanks, thought i got all of those.

coreydaley:
@dmage updated per your comments

pkg/resource/prunercronjob.go
adambkaplan: @coreydaley since 4.4 is not fast forwarding from master, please create a separate PR for the `--prune-registry` flag fix, along with a BZ if QE hasn't created one already.

We will want to fix this behavior before code freeze if possible.

coreydaley:
@adambkaplan ptal, all tests are passing.

pkg/resource/prunercronjob.go
```diff
@@ -130,6 +140,15 @@ func (gcj *generatorPrunerCronJob) expected() (runtime.Object, error) {
 	return cj, nil
 }
 
+func (gcj *generatorPrunerCronJob) getPruneRegistry(cr *imageregistryapiv1.Config) bool {
```
ricardomaraschini: nit: this function could be converted into `return cr.Spec.ManagementState == operatorapi.Managed`. the function name seems misguiding as well as it does not return a "PruneRegistry" :-)

pkg/operator/controllerimagepruner.go
```diff
@@ -0,0 +1,414 @@
...
+		Spec: imageregistryv1.ImagePrunerSpec{
+			Suspend:                    &defaultPrunerSuspend,
+			KeepTagRevisions:           &defaultPrunerKeepTagRevisions,
+			SuccessfulJobsHistoryLimit: &defaultPrunerSuccessfulJobsHistoryLimit,
+			FailedJobsHistoryLimit:     &defaultPrunerFailedJobsHistoryLimit,
+		},
+		Status: imageregistryv1.ImagePrunerStatus{},
+	}
+
+	client, err := regopset.NewForConfig(c.kubeconfig)
```
dmage: You can use `c.clients.RegOp`

pkg/resource/prunercronjob.go
coreydaley: The function handles three cases, Managed, Removed, and UnManaged, so the == would not handle all three.
I feel like the getPruneRegistry is inline with the rest of the get(functionNames) here.  Feel free to suggest a new name.

PR #461, opened 2020-02-18
WIP: Process only part of Mutators on ErrStorageNotConfigured.

ricardomaraschini:
If there is no storage configuration we return only Mutators that
don't depend on the storage to work.


ricardomaraschini:
This will make us not to degrade the operator in case of no storage config. Closing.

PR #462, opened 2020-02-18
Bug 1798618: Make driver on Mutators optional.

ricardomaraschini:
This patch is needed as when the image registry is Removed in an environment where there is no storage backend the operator keeps logging:

"controller.go:353] unable to sync: unable to get generators: storage backend not configured, requeuing"

During removal the driver property in some Mutators is optional. This PR makes the Mutators to work without them, returning error when they are mandatory.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -192,7 +192,7 @@ func (c *ImagePrunerController) syncPrunerStatus(cr *imageregistryv1.ImagePruner
 	}
 }
 
-func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Deployment, applyError error, removed bool) {
+func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Deployment, applyError error) {
 	operatorAvailable := operatorapiv1.OperatorCondition{
 		Status:  operatorapiv1.ConditionFalse,
 		Message: "",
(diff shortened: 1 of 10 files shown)


pkg/operator/finalizer.go
```diff
@@ -14,11 +14,20 @@ import (
...
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage"
 )
 
 func (c *Controller) RemoveResources(o *imageregistryv1.Config) error {
 	c.setStatusRemoving(o)
-	return c.generator.Remove(o)
+	err := c.generator.Remove(o)
+	switch {
+	case err == storage.ErrStorageNotConfigured:
+		return newPermanentError("StorageNotConfigured", err)
```
dmage: This shouldn't block us from removing the resources. We certainly can remove resources that do not require the storage (service accounts, node ca deamon, etc), ideally we should be able to remove Deployment as well.

ricardomaraschini: I have refactored this PR, PTAL. I still  have to change the commit comment though but i prefer to wait until we get into an agreement.

pkg/operator/status.go
```diff
@@ -203,13 +203,13 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
 	} else if deploy == nil {
-		if e, ok := applyError.(permanentError); ok {
-			operatorAvailable.Message = applyError.Error()
-			operatorAvailable.Reason = e.Reason
-		} else if removed {
+		if removed {
 			operatorAvailable.Status = operatorapiv1.ConditionTrue
 			operatorAvailable.Message = "The registry is removed"
 			operatorAvailable.Reason = "Removed"
+		} else if e, ok := applyError.(permanentError); ok {
```
dmage: Why do you want to reorder conditions?

And now I see we have a bug for Progressing when the state is Removed and applyError is not nil. We'll loose the error.

@ricardomaraschini can you remove this `removed` argument from the function and use `cr.Spec.ManagementState`? It should be less confusing.

ricardomaraschini:
/hold

Still need to simulate the behavior over AWS. Will unhold as soon as it is done.

pkg/operator/status.go
ricardomaraschini: Original code restored.

PR #465, opened 2020-02-22
Bug 1806158: Ignore Swift backend if the user doesn't have required permissions

Fedosin:
Now if Swift is available for OpenStack cloud, but the user doesn't have the `swiftoperator` role, installation fails with either 403 or 401 error code from Swift, depending on its configuration.

To prevent this we check if the user has this role, and if it's not the case we fallback to RWO PVC backend.

### DIFF
--- a/pkg/storage/swift/swift_test.go
+++ b/pkg/storage/swift/swift_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers"
 	th "github.com/gophercloud/gophercloud/testhelper"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


dmage:
In 4.x we wanted to reduce configuration variety. How often customers have this kind of configuration? Why don't they want to get swiftoperator role?

pkg/storage/swift/swift_test.go
```diff
@@ -672,3 +673,29 @@ func TestSwiftIsNotAvailable(t *testing.T) {
...
+	d, _ := mockConfig(false, th.Endpoint()+"v3", MockUPISecretNamespaceLister{})
+
+	conn, err := d.getSwiftClient()
+	th.AssertNoErr(t, err)
+
+	// if the user doesn't have permissions, gophercloud should return ErrDefault403
+	listOpts := containers.ListOpts{Full: false}
+	_, err = containers.List(conn, listOpts).AllPages()
+	_, ok := err.(gophercloud.ErrDefault403)
+	th.AssertEquals(t, true, ok)
```
dmage: Ideally IsSwiftEnabled should be checked there.

adambkaplan:
> In 4.x we wanted to reduce configuration variety. How often customers have this kind of configuration? Why don't they want to get swiftoperator role?

+1 to this comment. IMO this is no different than a customer installing on AWS but not giving the installing account permission to create S3 buckets. In these scenarios having user-provisioned install flows is the correct approach. Do we not have UPI for OpenStack yet?

I see that we explicitly call out granting the `swiftoperator` role in our installation docs for OpenStack [1]. Does this need to be modified for 4.4 since block storage (RWO) is an option for the registry?

[1] https://docs.openshift.com/container-platform/4.3/installing/installing_openstack/installing-openstack-installer-custom.html#installation-osp-enabling-swift_installing-openstack-installer-custom

Fedosin:
/hold

@adambkaplan @dmage  sorry for the confusion folks... everything is fine!

Ramon said: 
> With IPI, if a user wants Swift for the registry they’ll have to have the swiftoperator role. That’s documentation then. It’s better with your validation as  IPI installs will not fail and move on to use Cinder IIUC

I'm not closing this PR for now, in case there's any change. But please don't pay attention to it.

adambkaplan:
/close

@Fedosin based on @dmage 's comments I'm inclined to nack this. We have documentation on how to do UPI installs with OpenStack - customers who don't want to set up all of the IPI prerequisites should use this flow instead.

[1] https://docs.openshift.com/container-platform/4.3/installing/installing_openstack/installing-openstack-installer-custom.html

mandre:
Folks, we're re-opening this PR because this allows users to decide whether to use swiftless or not, by having the `swiftoperator` permission. It's common to have swift endpoints present, but not available to the users -- by default the `swiftoperator` permission isn't set in OSP deployments.

The alternative would be to force users to install using UPI, but that's not really practical.

racedo:
The use case here relates to the way RH OSP uses Swift specifically: OSP always has a Swift cluster in the controllers, unless another configuration is requested such as RGW or a dedicated Swift cluster. This default cluster isn't generally meant for tenants and exists to support Glance by default. Because of this, the swiftoperator role has to be given by an admin explicitly to allow tenants to use it (this isn't the case with RGW).

When we install OCP with IPI on RH OSP with default settings then, the installer will see a Swift cluster, only that in many cases it's a Swift cluster not meant to be used by tenants, and because of that it will fail (unless gave admin the swiftoperator role to the tenant).

That's why this check is critical and required, because besides checking if Swift endpoints are available, which will be in most cases, we have to check if it's a Swift cluster meant to be used by tenants. Without this, the check for Swift availability isn't complete and won't solve the issue for a large number of RH OSP deployments.

pkg/storage/swift/swift_test.go
adambkaplan: +1 - we should invoke `IsSwiftEnabled` and expect `false`.

Fedosin: done

PR #466, opened 2020-02-24
DEVEXP-515 - Mounting install pull secrets on image registry.

ricardomaraschini:
This PRs creates a Mutator that migrates installation pull secrets from
openshift-config namespace into registry's namespace. It also mounts
these pull secrets so we can later on leverage them when doing image
pull throughs.

### DIFF
--- a/pkg/resource/pullsecret.go
+++ b/pkg/resource/pullsecret.go
@@ -1,0 +1,101 @@
+package resource
+
+import (
+	"github.com/openshift/cluster-image-registry-operator/defaults"
+	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	coreset "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+var _ Mutator = &generatorPullSecret{}
+
+type generatorPullSecret struct {
+	client    coreset.CoreV1Interface
+	namespace string
+}
+
+func newGeneratorPullSecret(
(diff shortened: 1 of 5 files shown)


pkg/resource/pullsecret.go
```diff
@@ -0,0 +1,95 @@
...
+	return gs.namespace
+}
+
+func (gs *generatorPullSecret) GetName() string {
+	return defaults.InstallationPullSecret
+}
+
+func (gs *generatorPullSecret) expected() (runtime.Object, error) {
+	orig, err := gs.client.Secrets("openshift-config").Get(
+		"pull-secret", metav1.GetOptions{},
```
dmage: Does it always exist?

ricardomaraschini: please check it out.

pkg/resource/pullsecret.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	orig, err := gs.client.Secrets("openshift-config").Get(
+		"pull-secret", metav1.GetOptions{},
+	)
+	if err != nil && !errors.IsNotFound(err) {
+		return nil, err
+	} else if errors.IsNotFound(err) {
+		return sec, nil
+	}
```
dmage:
```
	if errors.IsNotFound(err) {
		return sec, nil
	} else if err != nil {
		return nil, err
	}
```

ricardomaraschini: much better. done.

gabemontero:
@dmage @ricardomaraschini  - a question occurred to me wrt this ... does this mean that the samples operator no longer needs to copy the install pull secret into the openshift namespace to facilitate sample imagestreams that pull from `registry.redhat.io` ?

If this came up during planning for DEVEXP-515 and either I missed it or forgot, apologies for the noise.

dmage:
@gabemontero we didn't discuss it, but yes, when we land this PR, the samples operator won't need to care about the install pull secret.

PR #469, opened 2020-02-26
Bug 1776665: Remove check for existence of backend storage

coreydaley:
Remove the check for the existence of the backend storage for the image registry


coreydaley:
I am open to the removal of the public StorageExists function for each storage type, but it is probably worth leaving around just in case we need it in the future.

adambkaplan:
@coreydaley operator tests failing because the expected status reasons are now gone. This leads to an important question - do we (as in OpenShift/k8s) consider status conditions API? Can we break a customer who is watching the reason codes for some kind of internal monitoring/scripting?

coreydaley:
I would think that the status is more important than the reason for the purposes of a customer watching it.

coreydaley:
/close
Superceded by #470

PR #470, opened 2020-02-27
Bug 1776665: Remove watch for KubeSystem ConfigMaps

coreydaley:
We should no longer rely on the installConfig configmap in the kubesystem.
Removing the watch for the kubesystem configmaps will fix the out of control
event processing every 4 seconds

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -378,7 +378,6 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 	configInformerFactory := configinformers.NewSharedInformerFactory(configClient, defaultResyncDuration)
 	kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(c.clients.Kube, defaultResyncDuration, kubeinformers.WithNamespace(c.params.Deployment.Namespace))
 	openshiftConfigKubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(c.clients.Kube, defaultResyncDuration, kubeinformers.WithNamespace(openshiftConfigNamespace))
-	kubeSystemKubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(c.clients.Kube, defaultResyncDuration, kubeinformers.WithNamespace(kubeSystemNamespace))
 	regopInformerFactory := regopinformers.NewSharedInformerFactory(c.clients.RegOp, defaultResyncDuration)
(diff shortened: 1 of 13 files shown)


pkg/storage/util/util.go
```diff
@@ -61,49 +58,6 @@ func UpdateCondition(cr *imageregistryv1.Config, conditionType string, status op
 	cr.Status.Conditions = conditions
 }
 
-// GetInfrastructure gets information about the cloud platform that the cluster is
-// installed on including the Type, Region, and other platform specific information.
-// Currently the install config is used as a backup to be compatible with upgrades
-// from 4.1 -> 4.2 when platformStatus did not exist, but should be able to be removed
-// in the future.
```
adambkaplan: the future is now!

pkg/operator/controller.go
```diff
@@ -459,11 +458,6 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 			c.listers.ImagePrunerConfigs = informer.Lister()
 			return informer.Informer()
 		},
-		func() cache.SharedIndexInformer {
-			informer := kubeSystemKubeInformerFactory.Core().V1().ConfigMaps()
-			c.listers.InstallerConfigMaps = informer.Lister().ConfigMaps(kubeSystemNamespace)
```
adambkaplan: Can we get rid fo `InstallerConfigMaps` in the listers object?

coreydaley: yes!

coreydaley: done

coreydaley:
@adambkaplan @dmage Can i get another lgtm? I had to make a small change.

pkg/storage/util/util.go
wking: The future isn't quite yet (blocked on [this][1] internal ticket).  Currently you can still get into trouble on 4.3+ with clusters that installed on 4.1 and have updated 4.1 -> 4.2 -> 4.3, because there's no migration code to consume the install-config and populate `PlatformStatus` (e.g. [here][2]).  Can we revert this, and open up some different fix to bug 1776665?  Or am I missing something?

[1]: https://issues.redhat.com/browse/CORS-1173
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1813343

coreydaley: Yes, we can revert this and apply the same fix that we did for 4.2 that just filters the events.

wking: The fix from #477 would avoid the issue with clusters which were originally 4.1.

PR #476, opened 2020-02-28
remove gosec check

coreydaley:
Per an email thread on aos-devel, gosec does not meet our security/compliance needs, and has already been removed from newer branches of the ciro code including master.  This is just a backport so that make build/make test doesn't take so long to run.


pierreprinetti:
@coreydaley 
Do you mind adding a bit of rationale/context to the commit message? That might help others making the right decision when planning tests for other repos

coreydaley:
@pierreprinetti updated description

pierreprinetti:
Thank you!
/lgtm

PR #479, opened 2020-03-03
Bug 1809467: Making storage driver optional on Mutators.

ricardomaraschini:
Making storage drivers optional on Mutator allows us to the latter when
deleting resources.

This is a stripped down version of:

9869211040b8a038a85f85802d5cf858371b58cb

The original commit contains refactoring that is not needed on a
backport, by doing so we avoid a merge conflict against 4.4 release.

### DIFF
--- a/pkg/resource/generator.go
+++ b/pkg/resource/generator.go
@@ -93,8 +93,10 @@ func (g *Generator) listRoutes(cr *imageregistryv1.Config) []Mutator {
 
 func (g *Generator) List(cr *imageregistryv1.Config) ([]Mutator, error) {
 	driver, err := storage.NewDriver(&cr.Spec.Storage, g.kubeconfig, g.listers)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/resource/generator.go
```diff
@@ -93,8 +93,10 @@ func (g *Generator) listRoutes(cr *imageregistryv1.Config) []Mutator {
 
 func (g *Generator) List(cr *imageregistryv1.Config) ([]Mutator, error) {
 	driver, err := storage.NewDriver(&cr.Spec.Storage, g.kubeconfig, g.listers)
-	if err != nil {
+	if err != nil && err != storage.ErrStorageNotConfigured {
 		return nil, err
+	} else if err == storage.ErrStorageNotConfigured {
+		klog.V(6).Info("storage not configured, some mutators might not work.")
```
dmage: 5 is the highest level that we should use in our code. So far we have only `1` and `4`.

PR #480, opened 2020-03-03
Refactor controllers to use shared clients, listers, and informers

coreydaley:


coreydaley:
/assign @dmage @ricardomaraschini @adambkaplan 

This PR also includes the import sorting from #481 so you can ignore that.
I know the tests aren't all passing yet, but it seems like those are mostly flakes, except for the unit tests which I am currently investigating.

pkg/client/generator.go
```diff
@@ -0,0 +1,216 @@
...
+	regopinformers "github.com/openshift/client-go/imageregistry/informers/externalversions"
+	routeset "github.com/openshift/client-go/route/clientset/versioned"
+	routesetv1 "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+	routeinformers "github.com/openshift/client-go/route/informers/externalversions"
+
+	"github.com/openshift/cluster-image-registry-operator/defaults"
+	"github.com/openshift/cluster-image-registry-operator/pkg/parameters"
+)
+
+type Generator struct {
```
dmage: Please give it another name, it's terrible to have the same meaningless name in two packages.

PR #481, opened 2020-03-03
Re-grouping and re-ordering go imports

coreydaley:
Re-grouping and re-ordering go imports according to our team preferences:
 - standard
 - other
 - kubernetes
 - openshift
 - module

### DIFF
--- a/cmd/cluster-image-registry-operator/main.go
+++ b/cmd/cluster-image-registry-operator/main.go
@@ -8,16 +8,17 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+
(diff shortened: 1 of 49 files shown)


tools.go
```diff
@@ -6,18 +6,19 @@
 package tools
 
 import (
-	// Makefile
 	"github.com/jteeuwen/go-bindata/go-bindata"
-	"github.com/openshift/library-go/alpha-build-machinery/make/targets/openshift"
+	// Makefile
```
dmage: this one is better to keep as is

coreydaley: Fixed, good catch.  I need to update the script to preserve comments, or allow an exclude flag

PR #482, opened 2020-03-05
Bug 1810317: bootstrap HA configuration

dmage:


pkg/operator/bootstrap.go
```diff
@@ -102,8 +102,8 @@ func (c *Controller) Bootstrap() error {
 		Spec: imageregistryv1.ImageRegistrySpec{
 			ManagementState: mgmtState,
 			LogLevel:        2,
-			Storage:         imageregistryv1.ImageRegistryConfigStorage{},
-			Replicas:        1,
+			Storage:         platformStorage,
+			Replicas:        replicas,
 			HTTPSecret:      fmt.Sprintf("%x", string(secretBytes[:])),
 			RolloutStrategy: string(rolloutStrategy),
```
adambkaplan: A note for others - we separately set the RolloutStrategy to `Recreate` if the registry is using block storage on OpenStack.

adambkaplan:
/cherrypick release-4.4

Note that we do not intend to ship the backport to 4.4.0 (we'd like this to be fully verified in 4.5.0 first).

@dmage are there concerns with a 4.3 backport for OpenStack? I don't think the fallbacks for swift storage are in that version.

dmage:
@adambkaplan backport for 4.3 will be manual because of these fallbacks. In 4.3 we bootstrap on OpenStack only with Swift, so we always can have 2 replicas.

PR #496, opened 2020-03-17
Bug 1776665: Ignore most kube system events

coreydaley:
Corrects issues that could arise from 4.1 clusters being upgraded through 4.5

 - Reverts #470 
 - Ignores events from the kube-system namespace except for the cluster-config-v1 configmap

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/openshift/api v0.0.0-20200311183032-85e16cc5dd7c
 	github.com/openshift/client-go v0.0.0-20200116152001-92a2713fa240
 	github.com/openshift/crd-schema-gen v1.0.0
-	github.com/openshift/installer v0.9.0-master.0.20190726121806-6e8f9c335410 // indirect
+	github.com/openshift/installer v0.9.0-master.0.20190726121806-6e8f9c335410
 	github.com/openshift/library-go v0.0.0-20200127110935-527e40ed17d9
 	github.com/prometheus/client_golang v1.1.0
 	github.com/prometheus/client_model v0.0.0-20191202183732-d1d2010b5bee
--- a/pkg/storage/util/util.go
+++ b/pkg/storage/util/util.go
@@ -8,9 +8,12 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metaapi "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
 
+	configv1 "github.com/openshift/api/config/v1"
 	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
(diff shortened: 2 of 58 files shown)


pkg/storage/util/util.go
```diff
@@ -59,6 +62,49 @@ func UpdateCondition(cr *imageregistryv1.Config, conditionType string, status op
 	cr.Status.Conditions = conditions
 }
 
+// GetInfrastructure gets information about the cloud platform that the cluster is
+// installed on including the Type, Region, and other platform specific information.
+// Currently the install config is used as a backup to be compatible with upgrades
+// from 4.1 -> 4.2 when platformStatus did not exist, but should be able to be removed
+// in the future.
```
bparees: @adambkaplan @coreydaley @dmage so we decided we still need this because a cluster from 4.1 doesn't have the new config info?

We should be pushing on the install team to ensure the new apis are populated during upgrade.  We cannot continue to carry 4.1 specific logic "just in case", we need to be able to assume that a 4.3 cluster looks like any other 4.3 cluster, regardless of what it started life as.

So if there are differences between a 4.1->4.2->4.3 cluster and a 4.3 cluster, those should be bugs against whoever owns the api that is manifesting differences.

dmage: @bparees yes, we have to carry this until they fix it. There is ticket to fix it: https://issues.redhat.com/browse/CORS-1173

bparees: perfect, thanks @dmage

coreydaley: @bparees @dmage We still need to merge this though.

go.mod
```diff
@@ -39,7 +39,7 @@ require (
 	github.com/openshift/api v0.0.0-20200311183032-85e16cc5dd7c
 	github.com/openshift/client-go v0.0.0-20200116152001-92a2713fa240
 	github.com/openshift/crd-schema-gen v1.0.0
-	github.com/openshift/installer v0.9.0-master.0.20190726121806-6e8f9c335410 // indirect
+	github.com/openshift/installer v0.9.0-master.0.20190726121806-6e8f9c335410
```
adambkaplan: @coreydaley why do we need an installer bump when the version listed in `go.mod` hasn't changed? I also don't see a corresponding change in `go.sum`.

coreydaley: IT was changed from being indirect to direct

coreydaley: the ciro would not compile without that update and pulling the installer back in as a direct dependency

dmage: because we use installer types in pkg/storage/util/util.go

pkg/storage/util/util.go
bparees: yes, understood. just wanted to make sure we were doing the right thing long term.

coreydaley:
Looks like I'll be creating manual pull requests for 4.3 & 4.4 :-1:

PR #505, opened 2020-03-27
Bug 1814709: Don't revert S3 Encrypt changes by user

coreydaley:
Set Encrypt:true if the S3 storage configuration is blank
Don't revert Encrypt:false if set by the user

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -679,6 +679,27 @@ func TestAWSChangeS3Encryption(t *testing.T) {
 			t.Errorf("unable to find environment variable: wanted %s", val.Name)
 		}
 	}
+
+	// If a user sets Encrypt: false, don't overwrite it
+	if _, err = client.Configs().Patch(defaults.ImageRegistryResourceName, types.MergePatchType, []byte(`{"spec": {"storage": {"s3": {"Encrypt": "false"}}}}`)); err != nil {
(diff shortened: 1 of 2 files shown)


test/e2e/aws_test.go
```diff
@@ -679,6 +679,27 @@ func TestAWSChangeS3Encryption(t *testing.T) {
 			t.Errorf("unable to find environment variable: wanted %s", val.Name)
 		}
 	}
+
+	// If a user sets Encrypt: false, don't overwrite it
```
dmage: That still changes the behaviour of the operator. In the current version a customer can provide only `keyID` or `cloudFront`, the operator will populate everything else and set `encrypt` to `true`.

I don't think we can fix this problem without breaking backward compatibility and without introducing a new API field.

ricardomaraschini: An idea, can't we make the encrypt field a pointer to a bool? At least I have seen this approach being used to differentiate between "user has set it to false" from "user did not set it". My two cents.

dmage: No, it's a bad idea to have different logic for `null` and `false`. We've already discussed it,  @deads2k suggested to deprecate the existing field and add `insecureDisableEncryptedStorage`.

ricardomaraschini: Makes sense.

PR #507, opened 2020-03-31
Bug 1807471: add always-on controller for image-registry-certificates

dmage:


dmage:
```
NodeControllerDegraded: The master nodes not ready: node \"ip-10-0-133-125.us-west-2.compute.internal\" not ready since 2020-03-31 17:31:49 +0000 UTC because NodeStatusUnknown (Kubelet stopped posting node status.)\nEtcdMembersDegraded: ip-10-0-133-125.us-west-2.compute.internal members are unhealthy,  members are unknown
```
/retest

pkg/operator/imageregistrycertificates.go
```diff
@@ -0,0 +1,115 @@
...
+		klog.Errorf("ImageRegistryCertificatesController: unable to sync: %s, requeuing", err)
+	} else {
+		c.queue.Forget(obj)
+		klog.Infof("ImageRegistryCertificatesController: event from workqueue successfully processed")
+	}
+	return true
+}
+
+func (c *ImageRegistryCertificatesController) sync() error {
+	g := resource.NewGeneratorCAConfig(c.configMapLister, c.imageConfigLister, c.openshiftConfigLister, c.serviceLister, c.coreClient)
```
ricardomaraschini, line 97: Could this be done when creating the object on `NewImageRegistryCertificatesController()` ?

test/framework/imageregistry.go
```diff
@@ -154,23 +155,88 @@ func deleteImageRegistryResource(te TestEnv) {
...
+		metav1.GetOptions{},
+	)
+	if errors.IsNotFound(err) {
+		return
+	}
+	if err != nil {
+		te.Fatalf("unable to get the current state of daemonset/node-ca: %s", err)
+	}
+
+	policy := metav1.DeletePropagationBackground
```
ricardomaraschini, line 181: Isn't it possible to use `DeletePropagationForeground` here so we don't need to wait with `WaitUntilFinalized()`?

pkg/operator/controller.go
```diff
@@ -477,6 +476,14 @@ func (c *Controller) Run(stopCh <-chan struct{}) error {
 		c.kubeconfig, c.clients, c.listers,
 	)
 
+	imageRegistryCertificatesController := NewImageRegistryCertificatesController(
+		c.clients.Core,
+		kubeInformerFactory.Core().V1().ConfigMaps(),
+		kubeInformerFactory.Core().V1().Services(),
```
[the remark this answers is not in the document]
dmage: I try to use the upstream approach for controllers. I think their code is cleaner and we have a lot to learn from them.

https://github.com/kubernetes/kubernetes/blob/3758426884e3c82cbd99c72e8015f4396f21fde2/cmd/kube-controller-manager/app/apps.go#L40

In this case `v1.Interface` instead of two informers will be much more opaque dependency. I don't want to read this code few month later and think:

Do it really wants to use all corev1 types?! I don't think so, but what kind of objects does this controller really need? Meh, I need to read its code...

test/framework/imageregistry.go
dmage: DeletePropagationForeground doesn't wait the object is finalized anyway, it's an instruction for GC. For daemonsets deleting its children can be slow. I'd prefer it to stay as a background job.

pkg/operator/imageregistrycertificates.go
dmage: Generators are lightweight objects (i.e. their constructors are trivial) and in this case it should be created on stack, so it doesn't matter. Eventually I'd expect us to get rid of these generators and inline them into controllers.

test/e2e/graceful_shutdown_test.go
```diff
@@ -27,8 +34,6 @@ func TestNodeCAGracefulShutdown(t *testing.T) {
 		t.Fatalf("unable to list pods: %v", err)
 	}
```
ricardomaraschini, line 36: I was looking at the error being reported by CI on this test and it seems like we could not find any pod to attach to(that would be very strange). Would you mind incorporating this slight change to make things clear? 

```suggestion
	if len(pods.Items) == 0 {
		t.Fatalf("no node-ca pods found")
	}
```

dmage:
@ricardomaraschini ptal again

PR #509, opened 2020-04-02
Bug 1809245: Try to attach to all pods during test

ricardomaraschini:
Graceful shutdown tests now attempt to attach to any available pod before failing.

This should mitigate the problem we are seeing during node-ca GracefulShutdown tests..

### DIFF
--- a/test/framework/logs.go
+++ b/test/framework/logs.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"regexp"
-	"testing"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


test/framework/logs.go
```diff
@@ -99,19 +98,23 @@ func DumpPodLogs(logger Logger, podLogs PodSetLogs) {
...
-// If an error happens when fetching pod's Stream() this function calls t.Fatal
-// while if a failure happens during pods log read the error is sent back to
-// the caller through an error channel.
-func MustFollowPodLog(t *testing.T, pod corev1.Pod) (<-chan string, <-chan error) {
-	client := MustNewClientset(t, nil)
+// If an error happens when fetching pod's Stream() this function returns
+// immediately. If a failure happens during pods log read the error is sent back
+// to the caller through an error channel.
+func FollowPodLog(pod corev1.Pod) (<-chan string, <-chan error, error) {
+	client, err := NewClientset(nil)
```
dmage, line 108: It's better to reuse the client (pass it as a parameter) and not to recreate it from scratch in every function.

ricardomaraschini: done.

PR #510, opened 2020-04-02
Using random name for PVC during tests

ricardomaraschini:
Allows for multiple consecutive executions of the same test.

### DIFF
--- a/test/e2e/pvc_test.go
+++ b/test/e2e/pvc_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
+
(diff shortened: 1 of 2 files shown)


test/e2e/pvc_test.go
```diff
@@ -189,7 +191,7 @@ func TestDefaultPVC(t *testing.T) {
 }
 
 func TestCustomRWXPVC(t *testing.T) {
-	claimName := "test-custom-rwx-pvc"
+	claimName := uuid.New().String()
```
dmage, line 194: Add some prefix so we can know where it came from in case of problems.

ricardomaraschini: done

PR #516, opened 2020-04-13
Bump gophercloud and related utils

Fedosin:


adambkaplan:
@Fedosin are there particular OpenStack fixes that the image registry operator needs with this bump?

Fedosin:
@adambkaplan hey! there is a strange dns issue with image registry operator when it can't authenticate in Keystone. I though gophercloud bump might help, but I was wrong.
I'm going to close this pr.

PR #518, opened 2020-04-14
Bug 1809005: Always set node selector kubernetes.io/os to linux

ricardomaraschini:
This assures we only run the operator and the operand on linux nodes.

### DIFF
--- a/pkg/resource/podtemplatespec.go
+++ b/pkg/resource/podtemplatespec.go
@@ -336,13 +336,21 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 		}
 	}
 
+	nodeSelectors := map[string]string{}
+	for k, v := range cr.Spec.NodeSelector {
+		nodeSelectors[k] = v
+	}
+	if _, ok := nodeSelectors["kubernetes.io/os"]; !ok {
+		nodeSelectors["kubernetes.io/os"] = "linux"
+	}
+
 	spec := corev1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: defaults.DeploymentLabels,
 		},
 		Spec: corev1.PodSpec{
 			Tolerations:       cr.Spec.Tolerations,
-			NodeSelector:      cr.Spec.NodeSelector,
+			NodeSelector:      nodeSelectors,
 			PriorityClassName: "system-cluster-critical",
 			Containers: []corev1.Container{
 				{


pkg/resource/podtemplatespec.go
```diff
@@ -336,13 +336,19 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 		}
 	}
 
+	nodeSelectors := map[string]string{}
+	for k, v := range cr.Spec.NodeSelector {
+		nodeSelectors[k] = v
+	}
+	nodeSelectors["kubernetes.io/os"] = "linux"
```
dmage, line 343: Shouldn't we allow users to override our default?

ricardomaraschini: I think it makes sense. I have adjusted it to avoid overwrite.

manifests/07-operator.yaml
```diff
@@ -21,6 +21,7 @@ spec:
       priorityClassName: system-cluster-critical
       nodeSelector:
         node-role.kubernetes.io/master: ""
+        kubernetes.io/os: linux
```
adambkaplan, line 24: This is somewhat redundant - we only support RHCOS masters I believe.

gabemontero: I vaguely recall a Jira story where we had to do something similar, i.e. make sure it doesn't run on windows, but I can't find the Jira story or analogous yaml in the repos I think I might have had to do this for.

pkg/resource/podtemplatespec.go
```diff
@@ -336,13 +336,21 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 		}
 	}
 
+	nodeSelectors := map[string]string{}
```
coreydaley, line 339: I'm not sure how I feel about this.  The image registry has to run on linux, but we are saying that an administrator can set a different os, even if it won't run there? Seems sketchy.

dmage: Settings `kubernetes.io/os` in this selector to anything other than `linux` is not supported today. However, if a customer decides to change its value anyway, the configuration should be broken today.

I can envision a situation when someone sets `kubernetes.io/os` to `windows` and forget about it, it works. Later we add support of this OS and his configuration suddenly has a new meaning.

adambkaplan: > Settings kubernetes.io/os in this selector to anything other than linux is not supported today

I don't think that is true. We are moving towards supporting Windows Containers (and that means Windows nodes) in the near future. My understanding is that this is "dev preview" in 4.4.

dmage: @adambkaplan are you saying that someone is trying to build the image registry for Windows?

adambkaplan: @dmage - at this point no, just that there will be Windows workers and we need to ensure we don't schedule or workloads on them by default. If a customer sets the node selector to be `windows`, their cluster goes degraded. So 🤷

ricardomaraschini: I agree with @dmage. It is better to crash with a wrong configuration now. This way we assure that the configuration reflects user intentions once we start supporting Window hosts.

manifests/07-operator.yaml
ricardomaraschini: I see your point @adambkaplan. I will get rid of this.

pkg/resource/podtemplatespec.go
ricardomaraschini: Hey guys, any extra consideration here? I would like to get this moving ahead.

PR #520, opened 2020-04-15
DEVEXP-368: get S3 endpoint from infrastructure status

dmage:


pkg/storage/s3/s3.go
```diff
@@ -190,6 +204,7 @@ func (d *driver) ConfigEnv() (envs envvar.List, err error) {
 		envvar.EnvVar{Name: "REGISTRY_STORAGE_S3_BUCKET", Value: d.Config.Bucket},
 		envvar.EnvVar{Name: "REGISTRY_STORAGE_S3_REGION", Value: d.Config.Region},
 		envvar.EnvVar{Name: "REGISTRY_STORAGE_S3_ENCRYPT", Value: d.Config.Encrypt},
+		envvar.EnvVar{Name: "REGISTRY_STORAGE_S3_VIRTUALHOSTEDSTYLE", Value: d.Config.VirtualHostedStyle},
```
adambkaplan, line 207: Do we want to always set this env var, or only set it if the custom region endpoint is set?

pkg/storage/s3/s3.go
```diff
@@ -131,7 +144,9 @@ func (d *driver) getS3Service() (*s3.S3, error) {
 	}
 	awsConfig.WithUseDualStack(true)
 	if d.Config.RegionEndpoint != "" {
-		awsConfig.WithS3ForcePathStyle(true)
+		if !d.Config.VirtualHostedStyle {
+			awsConfig.WithS3ForcePathStyle(true)
+		}
```
adambkaplan, line 149: looks good - matches https://github.com/openshift/docker-distribution/blob/image-registry-4.0-distribution-16128bb/registry/storage/driver/s3-aws/s3.go#L463-L468

pkg/storage/s3/s3.go
coreydaley: It shouldn't hurt anything, but we will need to make sure that the documentation details that it only works with the custom region endpoint.

dmage: Interesting point, we need to update https://github.com/docker/distribution/blob/master/docs/configuration.md in our upstream PR and eventually https://github.com/docker/docker.github.io/blob/master/registry/storage-drivers/s3.md

PR #521, opened 2020-04-17
Bug 1804160: Allow default service account to Get() namespaces

ricardomaraschini:


dmage:
@ricardomaraschini it doesn't look like necessary privilege, why do you need to access all namespaces?

ricardomaraschini:
@dmage I need to know if this is an acceptable change before I move on with another PR for the same bug on image-registry.

ricardomaraschini:
> @ricardomaraschini it doesn't look like necessary privilege, why do you need to access all namespaces?

To check if a namespace exists and return appropriate error. I can check that later on(when creating the ImageStream) but I don't consider it ideal as it makes image-registry to take more time to reply back.

dmage:
@ricardomaraschini you should interact with namespaces using client's credentials.

ricardomaraschini:
> @ricardomaraschini you should interact with namespaces using client's credentials.

I was not 100% sure it would work with client's credentials in all scenarios. I will implement using them and see how it is goes. Thanks.

dmage:
@ricardomaraschini it's hard to say without implementation on the image-registry side

PR #522, opened 2020-04-19
bindata: Run node-ca as hostNetwork to avoid NetworkNotReady

smarterclayton:
The node-ca daemon is expected to run on all nodes soon after start
and has no need of pod network access. This causes a large number
of NetworkNotReady events to be sent and is unnecessary - the pod
should start as soon as the node is up.


smarterclayton:
Hrm, this repo doesn't run the serial test.

PR #525, opened 2020-04-22
WIP - Bug 1825055: Explicitly setting route timeout to 30 seconds

ricardomaraschini:
Disclaimer - Still pending check as I could not deploy a cluster this morning.

### DIFF
--- a/pkg/resource/route.go
+++ b/pkg/resource/route.go
@@ -15,7 +15,10 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/defaults"
 )
 
-const RouteOwnerAnnotation = "imageregistry.openshift.io"
+const (
+	RouteOwnerAnnotation   = "imageregistry.openshift.io"
+	RouteTimeoutAnnotation = "haproxy.router.openshift.io/timeout"
+)
 
 func RouteIsCreatedByOperator(route *routeapi.Route) bool {
 	_, ok := route.Annotations[RouteOwnerAnnotation]
@@ -67,9 +70,12 @@ func (gr *generatorRoute) GetName() string {
 func (gr *generatorRoute) expected() (runtime.Object, error) {
 	r := &routeapi.Route{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        gr.GetName(),
-			Namespace:   gr.GetNamespace(),
-			Annotations: map[string]string{RouteOwnerAnnotation: "true"},
+			Name:      gr.GetName(),
+			Namespace: gr.GetNamespace(),
+			Annotations: map[string]string{
+				RouteTimeoutAnnotation: "30s",
+				RouteOwnerAnnotation:   "true",
+			},
 		},
 		Spec: routeapi.RouteSpec{
 			Host: gr.route.Hostname,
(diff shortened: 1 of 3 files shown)


pkg/resource/route.go
```diff
@@ -67,9 +70,12 @@ func (gr *generatorRoute) GetName() string {
 func (gr *generatorRoute) expected() (runtime.Object, error) {
 	r := &routeapi.Route{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        gr.GetName(),
-			Namespace:   gr.GetNamespace(),
-			Annotations: map[string]string{RouteOwnerAnnotation: "true"},
+			Name:      gr.GetName(),
+			Namespace: gr.GetNamespace(),
+			Annotations: map[string]string{
+				RouteTimeoutAnnotation: "30s",
```
bparees, line 76: given the issues we hit with this, do we perhaps want to set it even higher by default?

bparees: also will the operator stomp this value if the user changes it?  if so it seems like we need a config api for it (though setting it explicitly to a sufficiently high value by default can go in as the immediate fix since it's better than nothing)

dmage:
/hold

We need to investigate why we had this problem on the CI cluster. It's abnormal for the registry to keep silence for more than 30 seconds.

bparees:
> We need to investigate why we had this problem on the CI cluster. It's abnormal for the registry to keep silence for more than 30 seconds.

is the route timeout "30s for the request, total" or "30s between data packets"?

pkg/resource/route.go
dmage, line 76: Let's not create constants for strings that we use only once and that are not part of our API.

dmage:
@bparees it should be "30s between data packets", and it's also 30s before the first packet.

pkg/resource/route.go
ricardomaraschini: After some long discussion with @dmage allied to my last findings on this issue I am inclined to just close this PR. The default timeout of 30 seconds is, as per ha-proxy documentation, for the server (in this case the image-registry) to start to reply to a request. To make this value explicit does not help, neither help to increase it as I will try to explain below.

I have done some debug around this issue and I discovered that under heavy CPU pressure image-registry may take more than 30 seconds to start the reply to a "get manifest" request (we are currently requesting for only 100m of CPU by default) and I think that a 504 response on this scenario is a valuable piece of information regarding resource exhaustion.

As per version 4.4, even with the connection drop (that happens with the 504), image registry still tries to process the request completely. This behavior has changed on 4.5 so it is important to keep this timeout lower in order to stop processing requests for which nobody is there to read the result for. IMO if we are taking more than 30 seconds to start replying one needs to scale the registry horizontally.

I would like to hear inputs on this one. I will open a separate PR to increase default memory and CPU requests, we can discuss these aspects there.

cc @dmage @adambkaplan

bparees: your assessment makes sense to me.  if it were 30s for the total response i think we'd need something, but since it's 30s between packets(and initial packet) it doesn't seem necessary.

That said, i can envision cases where we might take 30s to do the initial response because our initial rbac lookup/layer-locating logic can be so expensive, particularly if your imagestreams are large and the apiserver is overwhelmed.  But if that's happening, the admin would, as you say, want to fix that, not raise the timeout.

ricardomaraschini:
As per https://github.com/openshift/cluster-image-registry-operator/pull/525#discussion_r413782391 I am closing this PR.

PR #537, opened 2020-04-23
Bug 1825055: Bumping default image registry resources

ricardomaraschini:
After some discussions with @dmage we decided that a bump on these values would be a good thing as it would restrict a little the number of other pods that may be running on the same node image registry pods are.

Any input on this is very welcomed as we run two image registry replicas by default, that would makes us to consume 2 vCPU and ~2Gb of memory from the day 1.

This relates to BZ 1825055 because I have noticed the erroneous behavior described on it when under CPU pressure.

### DIFF
--- a/pkg/resource/podtemplatespec.go
+++ b/pkg/resource/podtemplatespec.go
@@ -305,8 +305,8 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 
 	resources := corev1.ResourceRequirements{
 		Requests: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("100m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    resource.MustParse("1"),
+			corev1.ResourceMemory: resource.MustParse("1Gi"),
 		},
 	}
 	if cr.Spec.Resources != nil {


ricardomaraschini:
/assign @dmage @adambkaplan @bparees 

This is a branch of https://github.com/openshift/cluster-image-registry-operator/pull/525

pkg/resource/podtemplatespec.go
```diff
@@ -305,8 +305,8 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 
 	resources := corev1.ResourceRequirements{
 		Requests: corev1.ResourceList{
-			corev1.ResourceCPU:    resource.MustParse("100m"),
-			corev1.ResourceMemory: resource.MustParse("256Mi"),
+			corev1.ResourceCPU:    resource.MustParse("1"),
+			corev1.ResourceMemory: resource.MustParse("1Gi"),
```
bparees, line 309: hm, that's a big cpu number.  we've been working lately to reduce the requests most components are making down to something closer to what they are actually using normally because our footprint is getting huge and its preventing control plane workloads from getting scheduled.  

Do we have prometheus data showing the registry will normally need this much cpu?  Do we just need t o provide better tuning instructions to admins suggesting they raise these values under certain circumstances?

dmage: @bparees what is the normal usage? An empty cluster that doesn't use the registry at all, or our CI cluster that heavily relies on the registry?

bparees: i think we need to aim for some sort of useful default that covers typical week 1 usage for a customer.  So not the "no usage" case, but definitely not the api.ci usage.  api.ci needs to tune their cluster to their utilization.

ricardomaraschini:
After some research we find out that the defaults we have are appropriate for basic usage. I am closing this.

PR #542, opened 2020-04-30
Bug 1823676: Update relatedObjects and add operator starter (refactoring)

dmage:


dmage:
registry.svc.ci.openshift.org timeout
/retest

pkg/client/clients.go
```diff
@@ -14,13 +14,13 @@ import (
...
-	Core   *coreset.CoreV1Client
-	Apps   *appsset.AppsV1Client
-	RBAC   *rbacset.RbacV1Client
-	Batch  *batchset.BatchV1beta1Client
-	Job    *jobset.BatchV1Client
+	Core   coreset.CoreV1Interface
+	Apps   appsset.AppsV1Interface
+	RBAC   rbacset.RbacV1Interface
+	Batch  batchset.BatchV1beta1Interface
+	Job    jobset.BatchV1Interface
```
ricardomaraschini, line 25: This is neat.

cmd/cluster-image-registry-operator/main.go
```diff
@@ -67,19 +68,14 @@ func runOperator(cmd *cobra.Command, args []string) {
...
-	go imagePrunerController.Run(stopCh)
-
-	controller, err := operator.NewController(cfg)
-	if err != nil {
-		klog.Fatal(err)
-	}
+	ctx, cancel := context.WithCancel(context.Background())
+	go func() {
+		defer cancel()
+		<-stopCh
```
ricardomaraschini, line 74: Not sure if this applies here but as you are refactoring I have a question about our controllers in general.

What happens when we close this channel? In other words: do controllers stop abruptly or they finish processing whatever they are processing and then return? 

I think that is the first, not the latter. Wouldn't it be better to make them finish whatever they are doing before returning? A WaitGroup of some kind.

pkg/resource/prunerclusterrolebinding.go
```diff
@@ -57,7 +49,7 @@ func (gcrb *generatorPrunerClusterRoleBinding) expected() (runtime.Object, error
 			{
 				Kind:      "ServiceAccount",
 				Name:      "pruner",
-				Namespace: gcrb.GetNamespace(),
+				Namespace: "openshift-image-registry",
```
ricardomaraschini, line 52: Don't we have a variable on `defaults` with this?

cmd/cluster-image-registry-operator/main.go
dmage: Today the operator will exit immediately and it shouldn't be a problem (it may be killed or experience power outage at any moment).

https://github.com/kubernetes/kubernetes/blob/eec809aa93863afde30b40f03f8f380f006db794/cmd/kube-controller-manager/app/controllermanager.go#L117
kube-controller-manager doesn't care about it

https://github.com/kubernetes/kubernetes/blob/eec809aa93863afde30b40f03f8f380f006db794/cmd/kube-controller-manager/app/controllermanager.go#L278
when it looses the leadership (afaik SIGTERM also causes that), it also immediately stops.

That said, some controllers in library-go start to use WaitGroups. That might be useful for testing where you don't want to have alive controllers between tests.

ricardomaraschini: Interesting. So we could use WaitGroups but we don't want to have them necessarily on this PR?

dmage: Yes, eventually we may want to use WaitGroups and write unit tests for controllers.

PR #547, opened 2020-05-12
Bug 1813894: Stop using service ca from service account token

marun:
/cc @sttts @mfojtik @deads2k

### DIFF
--- a/pkg/client/listers.go
+++ b/pkg/client/listers.go
@@ -42,4 +42,5 @@ type ImagePrunerControllerListers struct {
 	ClusterRoleBindings krbaclisters.ClusterRoleBindingLister
 	RegistryConfigs     regoplisters.ConfigLister
 	ImagePrunerConfigs  regoplisters.ImagePrunerLister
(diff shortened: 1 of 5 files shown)


manifests/04-ca-service.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  annotations:
+    release.openshift.io/create-only: "true"
+    service.beta.openshift.io/inject-cabundle: "true"
+  name: service-ca
```
dmage, line 8: The operator creates `serviceca` (but only when the registry is managed). I'd rather reuse the existing and make the operator to always create it.

marun: @dmage Can you point me at where this creation occurs? As far as I can tell the operator is not creating an config map annotated for service ca injection? The only reference I can find to `service.beta.openshift.io/inject-cabundle` is for testing: https://github.com/openshift/cluster-image-registry-operator/blob/master/test/framework/configmap.go#L18

PR #549, opened 2020-05-13
WIP - Bug 1797512: Checking for RequestFailure return from AWS

ricardomaraschini:
If when requesting AWS endpoint(that may be a custom endpoint) we receive an error back related to the request result(e.g. a 404) we report appropriately on our operator status.


ricardomaraschini:
Working on an alternative solution(https://github.com/openshift/cluster-image-registry-operator/pull/555), closing this one as it crashed our tests completely.

PR #552, opened 2020-05-14
Bug 1700206: Use sentence case for error messages

dmage:
ClusterOperator condition messages should be correct sentences.


pierreprinetti:
Interesting. This would mean that from now on, if I lookup an error message string in the openshift codebase, I shall remember to turn on case-insensitive matching. Isn't that a bit confusing? Why would we want to arbitrarily changing error messages coming from other parts of the system?

If having consistent casing in the error messages is important, perhaps we may agree on a convention, rather than fixing at runtime.

dmage:
@pierreprinetti, updated, that should be more classical approach :)

PR #553, opened 2020-05-15
Bug 1834243: Refactored GCS day 2 e2e tests

ricardomaraschini:
GCS test was broken. Now it creates a bucket and tests if user provided
credentials propagate to running registry.

### DIFF
--- a/test/e2e/gcs_test.go
+++ b/test/e2e/gcs_test.go
@@ -1,9 +1,18 @@
 package e2e
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
 	"testing"
 	"time"
+
+	gstorage "cloud.google.com/go/storage"
+	"github.com/google/uuid"
+	goauth2 "golang.org/x/oauth2/google"
+	goption "google.golang.org/api/option"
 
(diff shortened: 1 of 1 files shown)


test/e2e/gcs_test.go
```diff
@@ -67,12 +56,61 @@ func TestGCSMinimal(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("error creating gcs credentials: %v", err)
+	}
+
+	gcli, err := gstorage.NewClient(ctx, goption.WithCredentials(credentials))
+	if err != nil {
+		t.Fatalf("error creating gcs client: %v", err)
+	}
+
+	bucketName := fmt.Sprintf("registry-test-%s", uuid.New().String())
```
dmage, line 91: It's better to prefix it with ClusterID aka infra.Status.InfrastructureName, in this case the installer will know that it's our cluster and protect our from leaks (in case of timeouts, OOMKilled, and other problems).

https://github.com/openshift/installer/blob/aecb9201459e3e92f12344e78aaaf03b6550fde7/pkg/destroy/gcp/bucket.go#L18

ricardomaraschini:
/test e2e-aws

"Failed to import expected imagestreams"

PR #554, opened 2020-05-15
Bug 1827075: Drain GCS buckets before removal

adambkaplan:
When removing GCS storage, ensure that there are no objects in the bucket prior to deletion.
Otherwise, removing the storage will fail with 409 CONFLICT errors.


adambkaplan:
/assign @dmage 

/cc @ricardomaraschini @coreydaley 

/hold

I'd like to discuss how we can effectively test this. As @coreydaley knows, the GCS SDK does not have good test doubles that allow us to easily unit test. My thought is that I should add a `TestRemoveStorage` suite that does the following:

1. Sets up the image registry with the defaults for the platform we are running on
2. Create a namespace
3. Imports the UBI 7 and 8 imagestreams with pullthrough into the namespace
4. Remove the image registry - thereby removing the storage.

Do we already have something like this?

adambkaplan:
/hold cancel

Added e2e test

adambkaplan:
@dmage @ricardomaraschini ptal - added an e2e test to ensure the image registry can be removed after images had been imported to it.

adambkaplan:
@dmage @ricardomaraschini ptal - I added an e2e to ensure we can tear down the registry if images were added to the storage (via pullthrough in this case).

dmage:
@adambkaplan you don't use pull-through there. Image API doesn't interact with the registry.

Pushing a blob or using builds might be easier rather than using pull-through. Maybe it's better to add a Disruptive test to openshift-tests?

adambkaplan:
> using builds might be easier rather than using pull-through

I can give that a go here - you can kick off a build using raw `Build` objects (`ci-operator` does this today)

adambkaplan:
@dmage updated to use a build, and the specific test had clean runs on GCP and AWS. PTAL

adambkaplan:
/retest

Flake (?) on `TestOperandProxyConfiguration`

PR #555, opened 2020-05-16
Bug 1797512: Reporting RequestFailure errors

ricardomaraschini:
If we fail to reach AWS endpoint (or custom region endpoint) report this
clearly on operator conditions.

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -164,7 +164,7 @@ func (d *driver) getS3Service() (*s3.S3, error) {
 func isBucketNotFound(err interface{}) bool {
 	switch s3Err := err.(type) {
 	case awserr.Error:
-		if s3Err.Code() == "NoSuchBucket" {
+		if s3Err.Code() == s3.ErrCodeNoSuchBucket {
 			return true
 		}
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
/test e2e-aws

"Failed to import expected imagestreams"

pkg/storage/s3/s3.go
```diff
@@ -369,7 +369,11 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 				Bucket: aws.String(d.Config.Bucket),
 			})
 			if err != nil {
-				if aerr, ok := err.(awserr.Error); ok {
+				if aerr, ok := err.(awserr.RequestFailure); ok {
+					err := fmt.Errorf("error requesting endpoint: %s(%d)", aerr.Code(), aerr.StatusCode())
```
dmage, line 373: I would rather not fix the bug than lose the error message.

ricardomaraschini: Let's log it?

ricardomaraschini:
/close

We have decided to keep the complete error being reported on Conditions.

PR #560, opened 2020-05-25
Bug 1835004: Use keepYoungerThanDuration

dmage:


go.mod
```diff
@@ -36,7 +36,7 @@ require (
 	github.com/hashicorp/golang-lru v0.5.3 // indirect
 	github.com/imdario/mergo v0.3.8 // indirect
 	github.com/onsi/gomega v1.7.1 // indirect
-	github.com/openshift/api v0.0.0-20200424083944-0422dc17083e
```
ricardomaraschini, line 39: I don't think changes on go.{mod,sum} belong to this commit.

dmage: That's an open question where go.mod changes should belong to. I tend to commit go.mod with the meaningful changes, as this is something that I changed manually.

dmage: fixed

PR #561, opened 2020-05-26
Bug 1828608: Removing ImageRegistryRemoved alert

ricardomaraschini:
Dear reviewer

In the past we have decided do bootstrap the registry as removed in some platforms, around the same time we also added an alert to notify the administrator if the registry was removed. Now we are having problems with some CI setup on VSphere(see related BZ) because the registry is bootstrapping as removed therefore this alarm is being triggered.

This PR removes this alert completely. Another option would be to postpone it to be triggered only a few hours later, please keep that in mind when reviewing this.

I have opted for removing this as the Registry being in a removed state is a "normal" situation, in the sense that any administrator may remove it by simply editing the config.

### DIFF
--- a/manifests/09-prometheus-rules.yaml
+++ b/manifests/09-prometheus-rules.yaml
@@ -17,17 +17,6 @@ spec:
         message: |
           Image Registry Storage configuration has changed in the last 30
           minutes. This change may have caused data loss.
-    - alert: ImageRegistryRemoved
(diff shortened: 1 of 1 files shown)


bparees:
Removing the alert seems sensible to me, i don't recall why we felt we needed one.  Using emptydir storage seems potentially worth an alert, but registry being explicitly removed should be quite apparent the the admin if it's something they care about.

adambkaplan:
@abhinavdahiya note that vSphere deployments will still need to configure registry storage on day 2.

PR #563, opened 2020-05-26
[WIP] Bug 1833109: Storage should be unmanaged when its configuration is changed by customer

dmage:


ricardomaraschini (changes requested):
May we close this one?

PR #564, opened 2020-05-26
Bug 1836753: Use release-4.3 api

dmage:


dmage:
We use release-4.4 API in the release-4.3 branch. We may want to backport validator annotations, `preserveUnknownFields: false`, but in this case we'll have to backport omitempty as well.

dmage:
Waiting for https://github.com/openshift/api/pull/547 to land

ashcrow:
I was unable to figure out why bumping api in vendor fixes the issue reading this PR as well as the bug. However, bumping has occurred in master and in 4.4 (to different versions specific to the release it looks like). In the future please consider adding the context to the PR or bug to make it easier for review and approval :pray:.

PR #565, opened 2020-05-28
Bug 1841080: Increasing default startingDeadlineSeconds

ricardomaraschini:
A value of 60 seconds might make the cronjob not to be started in time.
This patch increases the deadline to 10 minutes, as we run the pruner job
only once a day a delay increase here is acceptable.


dmage:
@ricardomaraschini why do we need this deadline? I'd rather remove it.

ricardomaraschini:
> @ricardomaraschini why do we need this deadline? I'd rather remove it.

I have bumped this to one hour, let me know what you think.

PR #573, opened 2020-06-26
IR-99: Added PrometheusRule to alert on v1 image imports

ricardomaraschini:
Registry protocol v1 has since long been deprecated and as we plan to same in the near future this commit adds an alert to inform users if they are still using v1. Alerts will last active for 15 minutes.

Metrics for this alert is exported by [openshift-apiserver](https://github.com/openshift/openshift-apiserver/pull/121)

### DIFF
--- a/manifests/09-prometheus-rules.yaml
+++ b/manifests/09-prometheus-rules.yaml
@@ -17,3 +17,21 @@ spec:
         message: |
           Image Registry Storage configuration has changed in the last 30
           minutes. This change may have caused data loss.
+    - alert: ImportImageUsingV1Protocol
+      expr: |
+        (
+            sum(
+                max_over_time(apiserver_v1_image_imports_total[15m]) or apiserver_v1_image_imports_total * 0
+            ) by (repository)
+            -
+            sum(
+                max_over_time(apiserver_v1_image_imports_total[15m] offset 15m) or apiserver_v1_image_imports_total * 0
+            ) by (repository)
+        ) > 0
+      labels:
+        severity: warning
+      annotations:
+        message: |
+          Registry repository {{ $labels.repository }} was imported using deprecated registry
+          protocol v1. Support for protocol v1 will be deprecated, it is recommended to use an
+          image registry that supports the newer v2 protocol.


manifests/09-prometheus-rules.yaml
```diff
@@ -17,3 +17,15 @@ spec:
...
           minutes. This change may have caused data loss.
+    - alert: ImportImageUsingV1Protocol
+      expr: increase(apiserver_v1_image_imports[15m]) > 0
+      labels:
+        severity: warning
+      annotations:
+        message: |
+          An attempt to import an image using deprecated registry protocol v1
+          was detected on pod {{ $labels.pod }}. Support for protocol v1 will
+          be deprecated, it is recommended to use an image registry who supports
```
bparees, line 28:
```suggestion
          be deprecated, it is recommended to use an image registry that supports
```

manifests/09-prometheus-rules.yaml
```diff
@@ -17,3 +17,15 @@ spec:
...
+          a newer and safest protocol. You can see what image got imported using
```
bparees, line 29:
```suggestion
          the newer v2 protocol. You can see what image was imported using
```

i'd rather not get into judgement type evaluations here.

dmage:
@bparees our registry has never supported v1 API. CRI-O has never supported it. There are no well-known public v1-only registries. I'd rather remove this support without any metrics/deprecation, but... In theory there might be a customer who still somehow able to use this feature.

lilic:
@dmage 
To answer the question from the other PR:

> @lilic if it's too noisy, the customer has a huge problem: they rely on something that will be gone.

So the problem is more about the false noises, rather than having this image imports version, as @simonpasquier mentioned in the comment:
> Overall I'm concerned that the alert isn't really actionable by a cluster admin and triggers alert fatigue. At the very least I would expect the severity label to be info rather than warning. Also looking at the definition of the alert, it could be that the alert starts firing and then resolves itself because no import using protocol v1 happened in the last 30 minutes.

https://github.com/openshift/openshift-apiserver/pull/121#discussion_r451630940

> It'll be mentioned in the release docs, but how customers will know that they actually use this feature?

We can tell them which metric to look at, in the release docs?

Another question I have is, could this alert be triggered by non user components? e.g. things outside of their control, like an operator they install or our own workload? This was a problem last time we did warning alerts for deprecated API.

ricardomaraschini:
Taking into account the feedback received on this and considering the outcome of some other discussions we have decided to close this PR without merging it. We will get the metric merged (https://github.com/openshift/openshift-apiserver/pull/121) and analyse the data from it later on, without generating this alert.

PR #574, opened 2020-07-01
IR-16: Leader election

ricardomaraschini:
This PR implements leader election for this operator, allowing multiple instances to run at the same time.


dmage:
@ricardomaraschini why don't you want to use https://godoc.org/github.com/openshift/library-go/pkg/controller/controllercmd ?

PR #575, opened 2020-07-02
IR-16: Implementing Leader Election

ricardomaraschini:
Using controllercmd package to implement:

1. Leader election
2. Gain file watching  capabilities.


ricardomaraschini:
/test e2e-aws-image-registry


cp: cannot create regular file '/var/run/secrets/ci.openshift.io/cluster-profile/day-2-manifests-insights-live.yaml': Read-only file system

ricardomaraschini:
/test e2e-aws

cp: cannot create regular file '/var/run/secrets/ci.openshift.io/cluster-profile/day-2-manifests-insights-live.yaml': Read-only file system

ricardomaraschini:
/retest

Error: Unable to find matching route for Route Table (rtb-0beb26e67e23ea820) and destination CIDR block (0.0.0.0/0).

cmd/cluster-image-registry-operator/main.go
```diff
@@ -24,59 +28,93 @@ import (
...
+			if len(filesToWatch) > 0 {
+				fileContents := make(map[string][]byte)
+				watchPaths := make([]string, 0)
+				for _, path := range filesToWatch {
+					fcontent, err := ioutil.ReadFile(path)
+					if err != nil {
+						klog.Warningf("Unable to watch %s: %v", path, err)
+						continue
+					}
+					fileContents[path] = fcontent
```
dmage, line 98: You don't need to provide this map unless you want to avoid race conditions. But to avoid race conditions, the file consumer should provide the value that was read from the file.

If the consumer cannot provide it, these is no reason to read it manually.

cmd/cluster-image-registry-operator/main.go
```diff
@@ -24,59 +28,93 @@ import (
...
+				func(ctx context.Context, cctx *controllercmd.ControllerContext) error {
+					printVersion()
+					rand.Seed(time.Now().UnixNano())
+					go metrics.RunServer(metricsPort)
+					return operator.RunOperator(ctx, cctx.KubeConfig)
+				},
+			).WithLeaderElection(
+				configv1.LeaderElection{
+					LeaseDuration: metav1.Duration{
+						Duration: 15 * time.Second,
```
dmage, line 76: Can we use default values?

cmd/cluster-image-registry-operator/main.go
```diff
@@ -24,59 +28,93 @@ import (
...
+	cmd := &cobra.Command{
+		Use:   "cluster-image-registry-operator",
+		Short: "OpenShift cluster image registry operator",
+		Run: func(cmd *cobra.Command, args []string) {
+			ctrl := controllercmd.NewController(
+				"image-registry-operator",
+				func(ctx context.Context, cctx *controllercmd.ControllerContext) error {
+					printVersion()
+					rand.Seed(time.Now().UnixNano())
+					go metrics.RunServer(metricsPort)
```
dmage, line 70: You can keep shopCh and pass `ctx.Done()`

ricardomaraschini: That was my initial approach but I decided to get rid of it, it is not necessary.

cmd/cluster-image-registry-operator/main.go
ricardomaraschini: seems like we can, patching it.

cmd/cluster-image-registry-operator/main.go
ricardomaraschini: This was rather confusing but yeah, seems like if we don't pass it it calculates on its own. Patching it.

PR #576, opened 2020-07-07
WIP - Bug 1853734: Implementing storage account access key cache

ricardomaraschini:
This commit implements a cache to avoid going to Azure remotely to list
storage account keys everytime. The cache keeps the information for one
minute.


dmage:
@ricardomaraschini this approach won't work as the driver is re-created on every event.

ricardomaraschini:
> @ricardomaraschini this approach won't work as the driver is re-created on every event.

Yes, don't know what I was thinking.

PR #577, opened 2020-07-09
Implementing unit tests for GetConfig

ricardomaraschini:
This PR adds unit tests for function `GetConfig()`. As I am moving now to work on our StorageInterface and this is kinda done I thought it would be good to have this in.

### DIFF
--- a/pkg/storage/azure/azure.go
+++ b/pkg/storage/azure/azure.go
@@ -19,6 +19,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/rand"
+	kcorelisters "k8s.io/client-go/listers/core/v1"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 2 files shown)


pkg/storage/azure/azure.go
```diff
@@ -75,47 +77,47 @@ func (e *errDoesNotExist) Error() string {
 	return e.Err.Error()
 }
 
-func getAzureConfigFromCloudSecret(creds *corev1.Secret) (*Azure, error) {
```
dmage, line 78: What's the point of merging these functions?

ricardomaraschini: I did not see the point of having one function to populate one struct.

dmage: To be sure that we don't use a wrong secret.

dmage: We had a bug about that in the past.

dmage: But tests should level the problem

PR #583, opened 2020-07-23
Enable registry by default for ovirt

rgolangh:
As part of the upcommig CSI Driver support a registry should be enabled by
default.
The CSI driver support is part of Cluster Storage Operator support in
external providers. See
https://github.com/openshift/enhancements/pull/352

Signed-off-by: Roy Golan <rgolan@redhat.com>

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -133,7 +133,6 @@ func GetPlatformStorage(listers *regopclient.Listers) (imageregistryv1.ImageRegi
 	// These are the platforms we don't configure any backend for, on these
 	// we should bootstrap the image registry as "Removed".
 	case configapiv1.BareMetalPlatformType,
-		configapiv1.OvirtPlatformType,
 		configapiv1.VSpherePlatformType,
 		configapiv1.NonePlatformType:
 		break
@@ -155,6 +154,11 @@ func GetPlatformStorage(listers *regopclient.Listers) (imageregistryv1.ImageRegi
 			replicas = 2
 			break
 		}
+		cfg.PVC = &imageregistryv1.ImageRegistryConfigStoragePVC{
+			Claim: defaults.PVCImageRegistryName,
+		}
+		replicas = 1
+	case configapiv1.OvirtPlatformType:
 		cfg.PVC = &imageregistryv1.ImageRegistryConfigStoragePVC{
 			Claim: defaults.PVCImageRegistryName,
 		}


rgolangh:
This depends on oVirt csi operator support in CSO - I'll link the relevant PR 
/hold

rgolangh:
@bennyz @Gal-Zaidman @gekorob please have a look

pkg/storage/storage.go
```diff
@@ -159,6 +158,11 @@ func GetPlatformStorage(listers *regopclient.Listers) (imageregistryv1.ImageRegi
 			Claim: defaults.PVCImageRegistryName,
 		}
 		replicas = 1
+	case configapiv1.OvirtPlatformType:
+		cfg.PVC = &imageregistryv1.ImageRegistryConfigStoragePVC{
+			Claim: defaults.PVCImageRegistryName,
```
dmage, line 163: When the claim name is present, the operator expects PVC to be provided.

pkg/storage/storage.go
```diff
@@ -159,6 +158,11 @@ func GetPlatformStorage(listers *regopclient.Listers) (imageregistryv1.ImageRegi
...
+		}
+		replicas = 1
```
dmage, line 165: Does oVirt storage support RWX?

dmage:
Can you add e2e-ovirt (always_run: false, optional: true) to this repo so that we can check this change?

pkg/storage/storage.go
rgolangh: No - that would be a shared disk from RHV side. Do we have other options?

rgolangh: I thought replica 1 would work with RWO?

dmage: We want to have at least 2 replicas when it's possible. But in this case the operator can have only 1 replica, so it's ok.

Gal-Zaidman:
> Can you add e2e-ovirt (always_run: false, optional: true) to this repo so that we can check this change?

created https://github.com/openshift/release/pull/10437

pkg/storage/storage.go
rgolangh: Noted. We would track this and use this case as a main driver for supporting RWX

rgolangh:
ovirt e2e won't pass because of the cluster storage operator work related to ovirt  https://github.com/openshift/cluster-storage-operator/pull/65
Alternatively a workaround for that e2e job may fill the gap https://github.com/openshift/release/pull/10233

PR #584, opened 2020-07-31
IR-119: Support IgnoreInvalidImageReferences

dmage:
When the image pruner's spec.ignoreInvalidImageReferences is true, the pruner should be run with --ingore-invalid-refs=true.

This option is enabled by default for new instances.

See also https://github.com/openshift/api/pull/708.


test/e2e/pruner_test.go
```diff
@@ -242,3 +251,90 @@ func TestPrunerPodCompletes(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	origSpec := cr.Spec.DeepCopy()
+
+	suspend := false
+	cr.Spec.Suspend = &suspend
+	cr.Spec.Schedule = "* * * * *"
+	cr.Spec.IgnoreInvalidImageReferences = true
```
bparees, line 270: do you have an inverse test?  (ensures the arg is not passed, or is passed w/ a value of false, when this is not set?)

dmage: now I have, fixed

ricardomaraschini:
@dmage aren't we planning to backport this to release-4.5 ? See https://bugzilla.redhat.com/show_bug.cgi?id=1852501#c39

dmage:
@ricardomaraschini, no, we don't backport API changes

PR #585, opened 2020-08-03
BUG 1862991: Remove storageclass from PVC creation

Gal-Zaidman:
When the csi driver for ovirt was merged into the storage operator[1] there is no "standard" StorageClass in the cluster, but there is a ovirt-csi-sc storage class which is the default storage class.
This cause the installer to fail due to image registry not being ready, since it was panding due to the pending pvc.
Instead the default storageclass should be used, see thread on forum-storage[2], not specifing the storage class[3] catches the default storage class.

[1] https://github.com/openshift/cluster-storage-operator/pull/65
[2] https://coreos.slack.com/archives/CBQHQFU0N/p1596452181198600
[3] https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1

Signed-off-by: Gal-Zaidman <gzaidman@redhat.com>

### DIFF
--- a/pkg/operator/bootstrap.go
+++ b/pkg/operator/bootstrap.go
@@ -12,12 +12,14 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 
+	configapiv1 "github.com/openshift/api/config/v1"
 	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
 	operatorapi "github.com/openshift/api/operator/v1"
 
 	"github.com/openshift/cluster-image-registry-operator/pkg/defaults"
 	"github.com/openshift/cluster-image-registry-operator/pkg/storage"
 	"github.com/openshift/cluster-image-registry-operator/pkg/storage/pvc"
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage/util"
 )
 
 // randomSecretSize is the number of random bytes to generate
@@ -107,6 +109,13 @@ func (c *Controller) createPVC(accessMode corev1.PersistentVolumeAccessMode, cla
 
 	// "standard" is the default StorageClass name, that was provisioned by the cloud provider
 	storageClassName := "standard"
+
+	// This is a Workaround for Bug#1862991 Tracker for removel on Bug#1866240
+	if infra, err := util.GetInfrastructure(c.listers); err != nil {
+		return err
+	} else if infra.Status.PlatformStatus.Type == configapiv1.OvirtPlatformType {
+		storageClassName = "ovirt-csi-sc"
+	}
 
 	claim := &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{


Gal-Zaidman:
e2e ovirt will not pass due to[1] but I will connect to the cluster and make sure that the ovirt-image-registry is up

[1] https://github.com/openshift/release/pull/10560

pkg/operator/bootstrap.go
```diff
@@ -105,9 +105,6 @@ func (c *Controller) createPVC(accessMode corev1.PersistentVolumeAccessMode, cla
 		return err
 	}
 
-	// "standard" is the default StorageClass name, that was provisioned by the cloud provider
```
rgolangh, line 108: true for ovirt. you need to switch case our platform

Gal-Zaidman: We talked offline but adding it here for reference:
the switch case is[1] and it means that only openstack and ovirt will get a PVC created since we both have a default storage class we should be ok

[1] https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/storage.go#L157-L162

dmage:
/hold

This change may break OpenStack. @Fedosin can you take a look?

Gal-Zaidman:
We are failing because the pvc is created before the storage class is available, see:
```
sh-4.2$ oc get pvc -A
NAMESPACE                  NAME                     STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   AGE
openshift-image-registry   image-registry-storage   Pending                                                     25m
sh-4.2$ oc describe pvc/image-registry-storage -n openshift-image-registry
Name:          image-registry-storage
Namespace:     openshift-image-registry
StorageClass:
Status:        Pending
Volume:
Labels:        <none>
Annotations:   imageregistry.openshift.io: true
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:
Access Modes:
VolumeMode:    Filesystem
Mounted By:    image-registry-68d4966679-mqgmf
Events:
  Type    Reason         Age                 From                         Message
  ----    ------         ----                ----                         -------
  Normal  FailedBinding  24m (x8 over 26m)   persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
  Normal  FailedBinding  23m (x3 over 23m)   persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
  Normal  FailedBinding  16m (x26 over 22m)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
  Normal  FailedBinding  44s (x62 over 15m)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
sh-4.2$ oc get storageclass -A
NAME                     PROVISIONER     RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
ovirt-csi-sc (default)   csi.ovirt.org   Delete          Immediate           false                  19m
```

does it make sense to require the image registry operator to start after the storage operator?

pkg/operator/bootstrap.go
```diff
@@ -4,14 +4,15 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
-
+	"github.com/openshift/cluster-image-registry-operator/pkg/storage/util"
```
dmage, line 7: nit, move this line to other `github.com/openshift/cluster-image-registry-operator` imports

Gal-Zaidman: fixed, it was there at the beginning I think that gofmt replaced the order for some reason :/

pkg/operator/bootstrap.go
```diff
@@ -4,20 +4,21 @@ import (
 	"context"
 	"crypto/rand"
 	"fmt"
-
```
ricardomaraschini, line 7: nit: this empty line should not be removed.

Gal-Zaidman: Done

Gal-Zaidman:
oVirt test passed :)

PR #586, opened 2020-08-05
Bug 1857684: Using last finished job status and disabling retries

ricardomaraschini:
This commit disables retries on pruning jobs, it also uses now the last
finished job status when checking for Condition (the operator becomes
deprecated only if the last finished job failed).


dmage:
@ricardomaraschini will the failed pod stay accessible after this change?

ricardomaraschini:
> @ricardomaraschini will the failed pod stay accessible after this change?

Yes, pods remain available for log checking. Last three successes and last three failures.

PR #587, opened 2020-08-05
Removed shared process namespace

ricardomaraschini:
As we don't have two containers running on this POD anymore there is no
need to have this shared namespace.

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -17,7 +17,6 @@ spec:
         name: cluster-image-registry-operator
     spec:
       serviceAccountName: cluster-image-registry-operator
-      shareProcessNamespace: true
       priorityClassName: system-cluster-critical
       nodeSelector:
         node-role.kubernetes.io/master: ""


manifests/07-operator.yaml
```diff
@@ -17,7 +17,6 @@ spec:
         name: cluster-image-registry-operator
     spec:
       serviceAccountName: cluster-image-registry-operator
-      shareProcessNamespace: true
```
wking, line 20: Kubernetes API for this property [defaults to `false`][1], but the cluster-version operator treats unset pointers in manifests as "[operator maintainers have no opinion][2]".  If you want to actually clear this, you should add an explicit `shareProcessNamespace`.  If you are fine with existing clusters rolling forward with `shareProcessNamespace: true`, then just removing it like you do in this PR is fine.  Or we can revisit the CVO's merge logic for this particular pointer field.  Thoughts?

[1]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podspec-v1-core
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1866554#c16

ricardomaraschini: Thanks for the heads up. I will create a PR to set this to false, I think it is better to keep enabled only features we actually use.

PR #588, opened 2020-08-17
Bug 1833109: Refactoring for Azure Driver

ricardomaraschini:
Yak shaving for future implementation of storage ManagementState.

### DIFF
--- a/pkg/storage/azure/azure.go
+++ b/pkg/storage/azure/azure.go
@@ -36,7 +36,6 @@ import (
 
 const (
 	storageExistsReasonNotConfigured     = "StorageNotConfigured"
-	storageExistsReasonPermissionDenied  = "StoragePermissionDenied"
 	storageExistsReasonConfigError       = "ConfigError"
 	storageExistsReasonUserManaged       = "UserManaged"
 	storageExistsReasonAzureError        = "AzureError"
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
/test e2e-azure-operator

These are passing on a development cluster, let's give it another try.

ricardomaraschini:
/retest

we had some Azure throttling issues, restart.

pkg/storage/azure/azure.go
```diff
@@ -454,162 +453,228 @@ func (d *driver) StorageChanged(cr *imageregistryv1.Config) bool {
...
+			cr,
+			defaults.StorageExists,
+			operatorapiv1.ConditionFalse,
+			storageExistsReasonNotConfigured,
+			"Storage account is provided, but container is not specified",
+		)
+		return
+	}
+
+	cr.Status.StorageManaged = false
```
dmage, line 578: that's not verification, the function name is misleading

pkg/storage/azure/azure.go
```diff
@@ -454,162 +453,228 @@ func (d *driver) StorageChanged(cr *imageregistryv1.Config) bool {
...
+
+	cr.Spec.Storage.Azure = d.Config.DeepCopy()
+	cr.Status.Storage = imageregistryv1.ImageRegistryConfigStorage{
+		Azure: d.Config.DeepCopy(),
+	}
+
+	if storageAccountCreated && containerCreated {
+		cr.Status.StorageManaged = true
+	} else {
+		cr.Status.StorageManaged = false
```
dmage, line 668: afaik the previous version didn't do this, and this might be wrong
if someone decided to change the container name and keep autogenerated account, the storage should still be managed by the operator (because everything is created by the operator)

pkg/storage/azure/azure.go
ricardomaraschini: Yes, makes sense. Updated.

PR #589, opened 2020-08-18
Bug 1870513: Fix wrong usage of ErrBucketNotExist

ricardomaraschini:
ErrBucketNotExist is never of type
"google.golang.org/api/googleapi".Error

Removed an unused piece code as well.


ricardomaraschini:
@dmage PTAL

dmage:
/lgtm
You need a BZ for that.

PR #591, opened 2020-08-19
Bug 1857782: Setting shareProcessNamespace to false

ricardomaraschini:
Cluster version operator treats unset pointers in manifests as "operator
maintainers have no opinion". If we want to actually clear this, we
need explicitly set shareProcessNamespace to false. As this property was
set to true we need now to set it to false.

For reference https://github.com/openshift/cluster-image-registry-operator/pull/587#discussion_r472576936

### DIFF
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -17,6 +17,7 @@ spec:
         name: cluster-image-registry-operator
     spec:
       serviceAccountName: cluster-image-registry-operator
+      shareProcessNamespace: false
       priorityClassName: system-cluster-critical
       nodeSelector:
         node-role.kubernetes.io/master: ""


ricardomaraschini:
@dmage for reference: https://github.com/openshift/cluster-image-registry-operator/pull/587#discussion_r472576936

wking:
`tide Pending — Not mergeable. Needs bugzilla/valid-bug, lgtm labels.`, so this will need to get hung on a bug to land.  Maybe [1857782][1] or [1866554][2]?  Or create a new one.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1857782
[2]: https://bugzilla.redhat.com/show_bug.cgi?id=1866554

PR #593, opened 2020-08-24
WIP - Bug 1808118: Implemented operator log level controller

ricardomaraschini:
LogLevel controller keeps track of Operator and Operand log levels.


pkg/operator/bootstrap.go
```diff
@@ -78,8 +78,11 @@ func (c *Controller) Bootstrap() error {
 			Finalizers: []string{defaults.ImageRegistryOperatorResourceFinalizer},
 		},
 		Spec: imageregistryv1.ImageRegistrySpec{
+			OperatorSpec: operatorapi.OperatorSpec{
+				LogLevel:         operatorapi.Normal,
+				OperatorLogLevel: operatorapi.Normal,
+			},
```
bparees, line 84: this is a somewhat awkward way to do this since you're putting operator config info into an operand config resource.

while it may work ok for now since registry is a singleton, it would be a problem if you wanted to allow multiple registry instances and in general doesn't feel quite right.

have any other operators addressed this?  what pattern did the use?

dmage: operator/v1.OperatorSpec is a generic Spec for all operators, so many other operators work the same way.

We won't want to have multiple registry instances because the registry itself is a singleton: it uses k8s api as to storage metadata (ImageStreams).

bparees: i know multiple registry instances will almost certainly never be a thing, but it still seems slightly odd for the operand config resource to control the operator, so i'd like to know we have other operators that work that way. 

after a bit of poking around it seems like we do, so i guess this is fine, if slightly unfortunate imho

dmage: @bparees I agree, that's because this config is started as a config for the operand, and other similar resources are started as configs for the operators. To make this config similar to other configs, it was transformed into a config for both: the operator (mostly status) and the operand (mostly spec). Both are singletons, so it works. Maybe eventually we'll split it and have two separate resources.

ricardomaraschini:
Closed in favor of https://github.com/openshift/cluster-image-registry-operator/pull/601

PR #595, opened 2020-08-26
Bug 1869613: Swift: add a unit test for bulk deleting failure

Fedosin:
This test emulates Swift behavior if it fails to delete one of the given objects in bulk mode.

### DIFF
--- a/pkg/storage/swift/swift_test.go
+++ b/pkg/storage/swift/swift_test.go
@@ -434,6 +434,51 @@ func TestSwiftRemoveStorageWithContent(t *testing.T) {
 	th.AssertEquals(t, true, containerDeleted)
 }
 
+func TestSwiftRemoveStorageWithContentFailure(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+	handleAuthentication(t, "container")
+
+	var containerContentListed bool
+	var containerContentDeleted bool
+	var containerDeleted bool
+	th.Mux.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
+		case http.MethodGet:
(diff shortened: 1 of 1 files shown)


pkg/storage/swift/swift_test.go
```diff
@@ -335,6 +335,51 @@ func TestSwiftRemoveStorageWithContent(t *testing.T) {
...
+			w.WriteHeader(http.StatusOK)
+			respBody := `{"Number Not Found": 0, "Response Status": "200 OK", "Errors": [["obj0", "Internal Error"]], "Number Deleted": 1, "Response Body": ""}`
+			_, _ = w.Write([]byte(respBody))
+		case http.MethodDelete:
+			w.WriteHeader(http.StatusNoContent)
+			containerDeleted = true
+		}
+	})
+
+	d, installConfig := mockConfig(true, th.Endpoint()+"v3", MockUPISecretNamespaceLister{})
```
mandre, line 371: `mockConfig`'s signature was changed with https://github.com/openshift/cluster-image-registry-operator/pull/596/files?file-filters%5B%5D=.go#diff-6cddc54abe4f6fba085c6fc5b982b9d9 and this call needs to be updated.

dmage:
/hold
@Fedosin is it expected that e2e-openstack fails that often?

Fedosin:
@dmage yes, openstack ci is highly unstable. it's not related to this PR, since it create a unit test only.

dmage:
/hold cancel
Let's see if it manages to land.

mandre:
e2e-openstack job was missing the optional attribute, that's the reason why prow it testing it over and over even though it's not shown as `Required` in the github UI.
Patch at https://github.com/openshift/release/pull/11521 should fix it.

PR #596, opened 2020-08-26
Bug 1833109: Spec.Storage.ManagementState implementation

ricardomaraschini:
`Spec.Storage.ManagementState` dictates how the underlying storage unit should be treated by the registry operator. It accepts two values, `Managed` or `Unmanaged` with the following consequences:

**Managed**
Indicates that the operator should "take care" of the underlying storage unit (i.e. the storage will be deleted if the registry is set to `Removed`)

**Unmanaged**
Asks the operator to leave the underlying storage unit alone, in this case the storage won't  be deleted if the registry is set to `Removed`

If `Spec.Storage.ManagementState` is left empty it is set appropriately during the driver bootstrap, once it is set the operator never touches it again, the user can then change its value at any moment. Changes on it reflect on `Status.StorageManaged` property on the next sync execution. `Status.StorageManaged` is going to be deprecated but for the time being it is expected to be in sync with `Spec.Storage.ManagementState`.

Some rules are taken into account when the driver boostraps the underlying storage:

1. If the operator created the underlying storage unit `ManagementState` is set to `Managed`
2. If user provided an underlying storage unit configuration (i.e. bucket or container name) and the storage unit already exists `ManagementState` is set to `Unmanaged`
3. If user provided an underlying storage unit configuration (i.e. bucket or container name) and the storage unit **does not** exist and is created by the operator `ManagementState` is set to `Managed`
4. If it is already set (`Managed` or `Unmanaged`) the operator does not change it, regardless of if it created or not the underlying storage unit.

In an attempt to make review easier this PR is broken into multiple commits, one for each driver and two extras: one for bumping `openshift/api` dependency and the other to implement the migration `Spec.Storage.ManagementState <-> Status.StorageManaged`


ricardomaraschini:
/assign @dmage 

I think this is ready for a first review.

pkg/storage/emptydir/emptydir.go
```diff
@@ -73,6 +73,9 @@ func (d *driver) StorageChanged(cr *imageregistryv1.Config) bool {
 }
 
 func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
+	if cr.Spec.Storage.ManagementState == "" {
+		cr.Spec.Storage.ManagementState = imageregistryv1.StorageManagementStateUnmanaged
```
dmage, line 77: Maybe it's better to say that emptyDir is also managed by default?

ricardomaraschini: done.

PR #598, opened 2020-08-28
Bug 1857687: Using last finished job and disabling retries

ricardomaraschini:
This commit disables retries on pruning jobs, it also uses now the last
finished job status when checking for Condition (the operator becomes
deprecated only if the last finished job failed).

### DIFF
--- a/pkg/operator/controllerimagepruner.go
+++ b/pkg/operator/controllerimagepruner.go
@@ -225,8 +225,15 @@ func (c *ImagePrunerController) sync() error {
 
 	lastPrunerJobConditions := []batchv1.JobCondition{}
 	if len(prunerJobs) > 0 {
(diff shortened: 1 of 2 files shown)


dmage:
@ricardomaraschini why didn't you use the cherrypick robot to make this PR?

ricardomaraschini:
> @ricardomaraschini why didn't you use the cherrypick robot to make this PR?

I thought that it would only work in an Open PR. Do you want me to do that? I gonna have to close the BZ because I think it will duplicate the original one.

ricardomaraschini:
@dmage I think you are right, I will use the cherry-pick bot and close this BZ. This is going to become a mess if I don't do that, parts of this PR solves a different problem.

PR #603, opened 2020-09-01
Bug 1853734: Reusing Azure API key

ricardomaraschini:
Reusing the same primary access key, caching it for five minutes.

### DIFF
--- a/pkg/storage/azure/azure_test.go
+++ b/pkg/storage/azure/azure_test.go
@@ -722,6 +722,7 @@ func Test_assureContainer(t *testing.T) {
 			drv := NewDriver(context.Background(), storageConfig, listers)
 			drv.authorizer = autorest.NullAuthorizer{}
 			drv.sender = sender
+			primaryKey = cachedKey{}
 
 			var requestCounter int
 			drv.httpSender = pipeline.FactoryFunc(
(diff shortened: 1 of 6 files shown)


pkg/storage/azure/cached_key.go
```diff
@@ -0,0 +1,49 @@
...
+		metrics.AzureKeyCacheHit()
+		return k.value, nil
+	}
+
+	keysResponse, err := cli.ListKeys(ctx, resourceGroup, account)
+	if err != nil {
+		return "", err
+	}
+
+	metrics.AzureKeyCacheMiss()
```
dmage, line 43: it should be before ListKeys

pkg/storage/azure/azure_test.go
```diff
@@ -722,6 +722,7 @@ func Test_assureContainer(t *testing.T) {
 			drv := NewDriver(context.Background(), storageConfig, listers)
 			drv.authorizer = autorest.NullAuthorizer{}
 			drv.sender = sender
+			primaryKey = cachedKey{}
```
dmage, line 725: is it in order to clear cache after another test?

pkg/storage/azure/cached_key.go
ricardomaraschini: right.

pkg/storage/azure/azure_test.go
ricardomaraschini: It does not change much, with this in place we can avoid having different mocked http responses when the key is cached. It gives all tests on that test table the same environment.

PR #604, opened 2020-09-07
Bug 1857246: Retrying in case of conflict during Config update

ricardomaraschini:
We have multiple controllers concurrently writting the status of the
config, this commit introduces retries in case of any of them updated
the config resource before us. This is clearly a workaround, as the
main controller may create storage if we fail to update the
configuration here it may ended up creating storage again on the next
cycle, hence the need to make sure we succeed.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -15,6 +15,7 @@ import (
 	kubeclient "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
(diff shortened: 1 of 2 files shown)


dmage:
@ricardomaraschini can you create an e2e test that checks that after the bootstrap storage is created only once?

ricardomaraschini:
> @ricardomaraschini can you create an e2e test that checks that after the bootstrap storage is created only once?

As to have tests for every platform would be overkill I changed the error message and implemented a test to make sure it is not happening. PTAL.

test/e2e/bootstrap_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/openshift/cluster-image-registry-operator/test/framework"
+)
+
+func TestBootstrapFailToUpdateSpec(t *testing.T) {
+	te := framework.Setup(t)
```
dmage, line 11: Use SetupAvailableImageRegistry instead of this bunch of Deploy/Wait/Ensure calls.

dmage:
/hold
it doesn't work

pkg/operator/controller.go
```diff
@@ -247,14 +248,26 @@ func (c *Controller) sync() error {
...
-		)
-		if err != nil {
-			if !errors.IsConflict(err) {
-				klog.Errorf("unable to update %s: %s", utilObjectInfo(cr), err)
+		var updatedCR *imageregistryv1.Config
+		if err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			updatedCR, err = c.listers.RegistryConfigs.Get(defaults.ImageRegistryResourceName)
+			if err != nil {
+				return err
 			}
```
dmage, line 256: the object from the lister shouldn't be mutated, DeepCopy may help

ricardomaraschini: done.

PR #605, opened 2020-09-08
Bug 1876961: Fix "slice index out of range"

dmage:
The result of GetOperatorState is used by different controllers to
report about their conditions, sometimes they mutate the returned value.

Listers return pointer to the same value if Get is used multiple times
and the object didn't change between calls.

It means that the clients of GetOperatorState could mutate the same
object without any synchronization. In some cases it caused panics.


ricardomaraschini (approved):
/lgtm

PR #606, opened 2020-09-10
Bug 1824834: httpSecret should be generated when it is not set

dmage:
This PR have two purposes:

1. it reduces the risk of failing to bootstrap the image registry config
2. it ensures that httpSecret is set even if the config object is created by someone else

### DIFF
--- a/test/e2e/configuration_test.go
+++ b/test/e2e/configuration_test.go
@@ -29,6 +29,52 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/test/framework"
 )
 
+func TestHTTPSecretDefaulter(t *testing.T) {
+	te := framework.SetupAvailableImageRegistry(t, &imageregistryapiv1.ImageRegistrySpec{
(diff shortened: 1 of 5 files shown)


test/e2e/configuration_test.go
```diff
@@ -29,6 +29,26 @@ import (
...
+
+	cr, err := te.Client().Configs().Get(context.Background(), defaults.ImageRegistryResourceName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if cr.Spec.HTTPSecret == "" {
+		t.Errorf("got empty spec.httpSecrets, want random string")
+	}
+}
```
ricardomaraschini, line 50: Could you update this test and set HTTPSecret back to empty and check if it gets populated again?

ricardomaraschini (changes requested):

ricardomaraschini (approved):
/lgtm

PR #609, opened 2020-09-16
Bug 1879426: Remove unsafe type casting

dmage:
This type casting is not needed as it goes right after the previous one which is made in a proper way.


ricardomaraschini (approved):
/lgtm

PR #612, opened 2020-09-18
Bug 1879176: Alerting on failed image prune job

ricardomaraschini:
1. Do not degrade the operator on job failures
2. Uses alerting instead of degrading the operator


markmc:
(patch manager) fixes a significant user-impacting issue

PR #617, opened 2020-09-29
Bug 1878040: Bumping openshift/api

ricardomaraschini:
Bump to include default values for log level api field.


dmage:
Replaced by #618.

PR #622, opened 2020-10-08
Cleanup: Use constants if possible

Fedosin:


dmage:
I was thinking to nuke this defaults.go, but while we still have it, let's add one more variable there.

/lgtm

PR #624, opened 2020-10-09
Bug 1886849: panic during creation of Azure storage container

dmage:
When the operator fails to create an Azure storage container, it panics in attempt to format the error reason.


dmage:
/retitle Bug 1886849: panic during creation of Azure storage container

bparees:
overriding BZ req after discussion w/ @dmage.  This code is not in 4.6 and there's no reasonable way to confirm this doesn't happen in 4.6 otherwise given the racey nature of the issue.

ricardomaraschini (approved):
/lgtm

PR #625, opened 2020-10-09
Bug 1850465: Keep node-ca daemonset in-sync with reference version

dmage:
The PR makes the operator to detect when the daemonset is changed and stomp these changes.


ricardomaraschini (approved):
/lgtm

PR #626, opened 2020-10-13
Bug 1885856: Creating prometheus record rule for v1 imports

ricardomaraschini:
Creating a record rule keeping track of the number of repository imports
using registry v1 protocol.

### DIFF
--- a/manifests/09-prometheus-rules.yaml
+++ b/manifests/09-prometheus-rules.yaml
@@ -17,3 +17,5 @@ spec:
         message: |
           Image Registry Storage configuration has changed in the last 30
           minutes. This change may have caused data loss.
+    - expr: sum(apiserver_v1_image_imports_total)
+      record: :apiserver_v1_image_imports:sum


ricardomaraschini:
/hold

Let's get someone from monitoring to LGTM this one as well.

manifests/09-prometheus-rules.yaml
```diff
@@ -17,3 +17,7 @@ spec:
         message: |
           Image Registry Storage configuration has changed in the last 30
           minutes. This change may have caused data loss.
+  - name: image-registry.rules
+    rules:
+    - expr: sum(apiserver_v1_image_imports_total)
+      record: apiserver_v1_image_imports_total:sum
```
simonpasquier, line 23: To comply with [best practices](https://prometheus.io/docs/practices/rules/):

```suggestion
      record: :apiserver_v1_image_imports:sum
```

manifests/09-prometheus-rules.yaml
```diff
@@ -17,3 +17,7 @@ spec:
         message: |
           Image Registry Storage configuration has changed in the last 30
           minutes. This change may have caused data loss.
+  - name: image-registry.rules
```
simonpasquier, line 20: I'm not sure you need to create another rule group, the new rule can be added to the existing `ImageRegistryOperator`.

ricardomaraschini: I see no problem in merging them, I was trying to separated alerting from recording rules (not sure if that makes sense though).

When it comes to the naming I noticed that best practices state that it should be named as: "level:metric:operations", on this case we are omitting the "level" and naming it as ":metric:operations", is this intentional ?

simonpasquier: > we are omitting the "level" and naming it as ":metric:operations", is this intentional ?

yes because you're aggregating away all labels. If you were doing `sum by(foo) (apiserver_v1_image_imports_total)` then the resulting metric would be `foo:apiserver_v1_image_imports`

simonpasquier (commented):
/lgtm

PR #627, opened 2020-10-13
Bug 1887010: Make pruner always aware of registry name

dmage:
If the image pruner is not provided with a registry hostname, it tries to guess it from image streams. But they might be missing on fresh clusters without the samples operator. To avoid this problem, the operator should provide the pruner with the registry hostname.


wzheng1:
When image registry is in Managed state, pruner job pod has below options:
      --prune-registry=true
      --registry-url=https://image-registry.openshift-image-registry.svc:5000
If image registry is set to Removed, pruner job will be set as below:
      --prune-registry=false
So this bug can be set to be verified now.

ricardomaraschini (approved):
/lgtm

PR #414, opened 2019-12-11
Bug 1769691: Report Available when registry is explicitly Removed

ricardomaraschini:
[This](https://github.com/openshift/cluster-image-registry-operator/pull/387) PR was not back ported to `release-4.1` and we are not not able to proceed with cluster updates. As I could not cherry-pick directly from the PR(merge conflicts) I am submitting it as a new one.

PR #629, opened 2020-10-22
add current profile annotations to CVO manifests

guillaumerose:
This is matches openshift/enhancements#414 and doesn't change existing behavior


dmage:
/hold
until the enhancement proposal is merged.

deads2k:
The name has been explicitly merged in https://github.com/openshift/enhancements/pull/510 .  This is consistent with the current state of the enhancement.  #414 will help describe future CVO input, but the handling is already in place for this default profile.

/lgtm

PR #631, opened 2020-11-02
Bug 1892799: Operator should use entrypoint script

dmage:
In 4.6 we've lost entrypoint script. This PR restores it and adds tests to avoid future regressions.


ricardomaraschini (approved):
/lgtm

PR #634, opened 2020-11-04
Bug 1893956: Make /etc/pki/ca-trust/extracted writable

dmage:


bparees:
@dmage looks like a unit test needs to be updated to expect the new mount

dmage:
/hold
It seems it doesn't work

dmage:
`update-ca-trust` expects that `/etc/pki/ca-trust/extracted/pem` and other directories already exist.

bparees:
i guess you could make the init script create those dirs, in the registry image.

dmage:
/hold cancel
This one seems to work.

PR #637, opened 2020-11-11
Bug 1894677: Reduce default log level for pruner

dmage:
Log level 2 causes `oc adm prune images` to print stack traces for goroutines in case of an error. This makes the last chunk of logs meaningless, but only this chunk is available in the termination message.

### DIFF
--- a/pkg/resource/prunercronjob.go
+++ b/pkg/resource/prunercronjob.go
@@ -239,7 +239,17 @@ func (gcj *generatorPrunerCronJob) getKeepYoungerThan(cr *imageregistryapiv1.Ima
 }
 
 func (gcj *generatorPrunerCronJob) getLogLevel(cr *imageregistryapiv1.ImagePruner) int {
-	return loglevel.LogLevelToVerbosity(cr.Spec.LogLevel)
+	level := loglevel.LogLevelToVerbosity(cr.Spec.LogLevel)
+	if level == 2 {
+		// The Normal log level is 2, but it causes `oc` to print stack traces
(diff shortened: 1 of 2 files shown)


pkg/resource/prunercronjob.go
```diff
@@ -239,7 +239,17 @@ func (gcj *generatorPrunerCronJob) getKeepYoungerThan(cr *imageregistryapiv1.Ima
 }
 
 func (gcj *generatorPrunerCronJob) getLogLevel(cr *imageregistryapiv1.ImagePruner) int {
-	return loglevel.LogLevelToVerbosity(cr.Spec.LogLevel)
+	level := loglevel.LogLevelToVerbosity(cr.Spec.LogLevel)
```
ricardomaraschini, line 242: I think it makes it more assertive if we do something like this:

if cr.spec.loglevel == normal {
    return 1
}
return loglevel.LogLevelToVerbosity(cr.Spec.LogLevel)

no further comment would be needed as our normal is level 1. what do you think?

ricardomaraschini (changes requested):

pkg/resource/prunercronjob.go
dmage: Take a look at LogLevelToVerbosity, it returns 2 when it gets Normal or something unknown (like an empty string). The comment will be needed anyway because it's unclear why we cannot use cluster defaults.

ricardomaraschini (approved):
/lgtm

PR #638, opened 2020-11-11
CORS-1584: support custom CA bundle for AWS API

staebler:
When a cluster is installed in a AWS C2S region, access to the AWS API requires a custom CA bundle for trust. The custom CA bundle is read from the "ca-bundle.pem" key of the kube-cloud-config ConfigMap in the openshift-config-managed namespace.

https://issues.redhat.com/browse/CORS-1584

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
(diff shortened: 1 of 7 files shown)


pkg/storage/s3/s3.go
```diff
@@ -111,13 +113,26 @@ func GetConfig(listers *regopclient.Listers) (*S3, error) {
 		}
 	}
 
+	cloudConfig, err := listers.OpenShiftConfigManaged.Get(defaults.KubeCloudConfigName)
```
staebler, line 116: For OpenStack, the "ca-bundle.pem" is grabbed from the openshift-config/cloud-provider-config ConfigMap. I have opted to use what I think is the more appropriate openshift-config-managed/kube-cloud-config ConfigMap. In the long term, we should probably be consistent, however.

pkg/storage/s3/s3.go
```diff
@@ -111,13 +113,26 @@ func GetConfig(listers *regopclient.Listers) (*S3, error) {
...
+		if ok {
+			cfg.CABundle = caBundle
+		}
+	}
+
 	return cfg, nil
 }
 
-func (d *driver) updateConfigAndGetCredentials() (accessKey, secretKey string, err error) {
+func (d *driver) updateConfigAndGetCredentials() (*S3, error) {
```
dmage, line 132: This function shouldn't return `*S3`. It makes really tempting to use it instead of `d.Config`.
If you don't want it to be a part of `spec.storage.s3` nor a part of the configmap `image-registry-private-configuration-user`, then use shouldn't touch this function.

dmage:
/retitle CORS-1584: support custom CA bundle for AWS API

pkg/storage/s3/s3.go
dmage: Either the config should have the name of the configmap and this name should be used only in getS3Service, or getS3Service should use the well-known configmap.

staebler: Thanks. I have separated the code for fetching the custom CA bundle into its own function and removed it from `spec.storage.s3`.

PR #639, opened 2020-11-13
Bug 1897520: Use mkdir -p to create ca-trust directories

dmage:
When the registry is restarted, it already has content in the ca-trust extracted directory from the previous run. So `mkdir` shouldn't fail when directories already exist.


ricardomaraschini (approved):
/lgtm

PR #640, opened 2020-11-13
Bug 1921909: build and use AWS config file for operator and operand

joelddiaz:
In order to support alternative AWS authentication mechanisms beyond just access/secret key authentication:

- [x] Update the operator Deployment to mount in the projected ServiceAccount token (to a default-by-convention location of /var/run/secrets/openshift/serviceaccount).
- [x] Update the AWS S3 driver to respect the new CredentialsRequest Secret field ('credentials'), but keep supporting the old-style format.
- [x] Build and use an AWS config file when creating the AWS S3 client (whether new-style or old-style Secret data).
- [x] Store the AWS credentials config in the image-registry's image-registry-private-configuration Secret.
- [x] Mount the image-registry-private-configuration Secret into the image-registry Deployment PodSpec so that the image-registry can start consuming creds from the Secret instead of only using access/secret key data (requires separate code changes to image-registry).
- [x] Update the image-registry environment variables so that the operator sets the location of the fully-formed credentials file.
- [x] Modify the PodTemplate to always mount in the ServiceAccount projected volume (to a default-by-convention location of /var/run/secrets/openshift/serviceaccount).

With these changes (and the corresponding image-registry support), it should now be possible to use web identity token authentication. This requires:
- setting up an AWS OpenID identity provider that trusts the cluster's OIDC provider
- creating an AWS IAM Role with the permissions requested in this repo's CredentialsRequest
- granting both ServiceAccounts 'registry' and 'cluster-image-registry-operator' to be allowed to assume the AWS IAM Role
- setting up the Secret holding the AWS credentials (installer-cloud-credentials) so that the 'credentials' field contains a working AWS configuration:
```
[default]
role_arn = arn:aws:iam:ACCOUNT_NUM:role/ROLE_NAME
web_identity_token_file = /var/run/secrets/openshift/serviceaccount/token
```

There is still an open question about if/when we move to temporary credentials by default, how the e2e tests should work. They presently re-use the credentials found in the cluster, but when using web identity authentication, it is the Pod that is granted permissions and there are no credentials that can just be grabbed for the various e2e tests.

This is related to https://issues.redhat.com/browse/CO-1255

### DIFF
--- a/manifests/01-registry-credentials-request.yaml
+++ b/manifests/01-registry-credentials-request.yaml
@@ -31,10 +31,10 @@ spec:
       - s3:GetLifecycleConfiguration
       - s3:GetBucketLocation
       - s3:ListBucket
-      - s3:HeadBucket
       - s3:GetObject
       - s3:PutObject
       - s3:DeleteObject
       - s3:ListBucketMultipartUploads
       - s3:AbortMultipartUpload
+      - s3:ListMultipartUploadParts
       resource: "*"
--- a/manifests/07-operator.yaml
+++ b/manifests/07-operator.yaml
@@ -69,6 +69,9 @@ spec:
               mountPath: /var/run/configmaps/trusted-ca/
             - name: image-registry-operator-tls
               mountPath: /etc/secrets
+            - name: bound-sa-token
+              mountPath: /var/run/secrets/openshift/serviceaccount
+              readOnly: true
       volumes:
         - name: image-registry-operator-tls
           secret:
@@ -80,3 +83,12 @@ spec:
             items:
             - key: ca-bundle.crt
               path: tls-ca-bundle.pem
+        # This service account token can be used to provide identity outside the cluster.
+        # For example, this token can be used with AssumeRoleWithWebIdentity to authenticate
+        # with AWS using an IAM OIDC provider and STS.
+        - name: bound-sa-token
+          projected:
+            sources:
+            - serviceAccountToken:
+                path: token
+                audience: openshift
--- a/pkg/resource/deployment_test.go
+++ b/pkg/resource/deployment_test.go
@@ -104,8 +104,14 @@ func TestChecksum(t *testing.T) {
 				secretLister:    secretLister,
 			}
 
-			cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().ConfigMaps().Informer().HasSynced)
-			cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().Secrets().Informer().HasSynced)
+			waitCtx, waitCancel := context.WithTimeout(ctx, time.Minute)
+			if !cache.WaitForCacheSync(
+				waitCtx.Done(),
+				kubeInformer.Core().V1().ConfigMaps().Informer().HasSynced,
+				kubeInformer.Core().V1().Secrets().Informer().HasSynced) {
+				t.Fatal("caches not syncing")
+			}
+			defer waitCancel()
 
 			// Generate a base deployment
 			obj, err := gd.expected()
@@ -126,7 +132,6 @@ func TestChecksum(t *testing.T) {
 					t.Fatalf("failed to update secret: %s", err)
 				}
 
-				cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().Secrets().Informer().HasSynced)
 				// Fixme: why is this wait necessary?
 				if err := waitForUpdatedSecret(ctx, kubeClient, test.updatedSecret.Data); err != nil {
 					t.Fatalf("failed waiting for secret update to become visible: %s", err)
--- a/pkg/resource/deployment_test.go
+++ b/pkg/resource/deployment_test.go
@@ -1,0 +1,239 @@
+package resource
+
+import (
+	"context"
+	"reflect"
+	"testing"
+	"time"
+
+	appsapi "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	kubeinformers "k8s.io/client-go/informers"
+	kubeclient "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/cache"
+
+	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
+	fakeconfig "github.com/openshift/client-go/config/clientset/versioned/fake"
+	configinformers "github.com/openshift/client-go/config/informers/externalversions"
+
+	"github.com/openshift/cluster-image-registry-operator/pkg/defaults"
+	"github.com/openshift/cluster-image-registry-operator/pkg/envvar"
+)
+
+func TestChecksum(t *testing.T) {
+	tests := []struct {
+		name          string
+		origSecret    *corev1.Secret
+		updatedSecret *corev1.Secret
+		validate      func(*testing.T, string, *appsapi.Deployment)
+	}{
+		{
+			name: "no secret changes",
+			origSecret: testSecret(map[string][]byte{
+				"credentials": []byte("orig creds"),
+			}),
+			validate: func(t *testing.T, origHash string, dep *appsapi.Deployment) {
+				currentHash := dep.Annotations[defaults.ChecksumOperatorAnnotation]
+				if origHash != currentHash {
+					t.Errorf("Has unexpectedly changed from %s to %s", origHash, currentHash)
+				}
+			},
+		},
+		{
+			name: "secret changes",
+			origSecret: testSecret(map[string][]byte{
+				"credentials": []byte("orig creds"),
+			}),
+			updatedSecret: testSecret(map[string][]byte{
+				"credentials": []byte("updated creds"),
+			}),
+			validate: func(t *testing.T, origHash string, dep *appsapi.Deployment) {
+				currentHash := dep.Annotations[defaults.ChecksumOperatorAnnotation]
+				if origHash == currentHash {
+					t.Errorf("Hash unexpectedly didn't change from %s", origHash)
+				}
+			},
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			kubeClient := fake.NewSimpleClientset(test.origSecret)
+			kubeInformer := kubeinformers.NewSharedInformerFactory(kubeClient, 0)
+
+			configClient := fakeconfig.NewSimpleClientset()
+			configInformer := configinformers.NewSharedInformerFactory(configClient, 0)
+
+			cmLister := kubeInformer.Core().V1().ConfigMaps().Lister().ConfigMaps(defaults.ImageRegistryOperatorNamespace)
+			secretLister := kubeInformer.Core().V1().Secrets().Lister().Secrets(defaults.ImageRegistryOperatorNamespace)
+
+			proxyLister := configInformer.Config().V1().Proxies().Lister()
+
+			kubeInformer.Start(ctx.Done())
+			configInformer.Start(ctx.Done())
+
+			_, err := kubeClient.CoreV1().Namespaces().Create(context.TODO(), &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: defaults.ImageRegistryOperatorNamespace,
+					Annotations: map[string]string{
+						defaults.SupplementalGroupsAnnotation: "1/2",
+					},
+				},
+			}, metav1.CreateOptions{})
+			if err != nil {
+				t.Errorf("failed to create namespace for test: %s", err)
+				t.FailNow()
+			}
+
+			_, err = kubeClient.CoreV1().ConfigMaps(defaults.ImageRegistryOperatorNamespace).Create(context.TODO(), &corev1.ConfigMap{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      defaults.ImageRegistryCertificatesName,
+					Namespace: defaults.ImageRegistryOperatorNamespace,
+				},
+			}, metav1.CreateOptions{})
+			if err != nil {
+				t.Errorf("unable to create ConfigMap %s: %s", defaults.ImageRegistryCertificatesName, err)
+				t.FailNow()
+			}
+
+			driver := &testDriver{}
+
+			gd := &generatorDeployment{
+				driver:          driver,
+				coreClient:      kubeClient.CoreV1(),
+				proxyLister:     proxyLister,
+				cr:              &imageregistryv1.Config{},
+				configMapLister: cmLister,
+				secretLister:    secretLister,
+			}
+
+			cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().ConfigMaps().Informer().HasSynced)
+			cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().Secrets().Informer().HasSynced)
+
+			// Generate a base deployment
+			obj, err := gd.expected()
+			if err != nil {
+				t.Errorf("unexpected error: %s", err)
+				t.FailNow()
+			}
+
+			generatedDeployment, ok := obj.(*appsapi.Deployment)
+			if !ok {
+				t.Errorf("failed to convert to Deployment")
+				t.FailNow()
+			}
+
+			origHash := generatedDeployment.Annotations[defaults.ChecksumOperatorAnnotation]
+
+			if test.updatedSecret != nil {
+				_, err = kubeClient.CoreV1().Secrets(defaults.ImageRegistryOperatorNamespace).Update(context.TODO(), test.updatedSecret, metav1.UpdateOptions{})
+				if err != nil {
+					t.Errorf("failed to update secret: %s", err)
+					t.FailNow()
+				}
+
+				cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().Secrets().Informer().HasSynced)
+				// Fixme: why is this wait necessary?
+				if err := waitForUpdatedSecret(t, kubeClient, test.updatedSecret.Data); err != nil {
+					t.Errorf("failed waiting for secret update to become visible: %s", err)
+					t.FailNow()
+				}
+			}
+
+			// Check for expected changes in the deployment
+			obj, err = gd.expected()
+			if err != nil {
+				t.Errorf("unexpected error: %s", err)
+				t.FailNow()
+			}
+
+			regeneratedDeployment, ok := obj.(*appsapi.Deployment)
+			if !ok {
+				t.Errorf("failed to convert to Deployment")
+				t.FailNow()
+			}
+
+			test.validate(t, origHash, regeneratedDeployment)
+		})
+	}
+}
+
+func testSecret(sData map[string][]byte) *corev1.Secret {
+	return &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: defaults.ImageRegistryOperatorNamespace,
+			Name:      defaults.ImageRegistryPrivateConfiguration,
+		},
+		Data: sData,
+	}
+}
+
+func waitForUpdatedSecret(t *testing.T, kubeClient kubeclient.Interface, expectedData map[string][]byte) error {
+	err := wait.Poll(time.Second, time.Minute, func() (stop bool, err error) {
+		sec, err := kubeClient.CoreV1().Secrets(defaults.ImageRegistryOperatorNamespace).Get(context.TODO(), defaults.ImageRegistryPrivateConfiguration, metav1.GetOptions{})
+		if err != nil {
+			// Keep waiting
+			return false, nil
+		}
+		if reflect.DeepEqual(sec.Data, expectedData) {
+			return true, nil
+		}
+		// Keep trying
+		return false, nil
+	})
+	if err != nil {
+		t.Fatalf("failed to wait for the Secret to be updated: %s", err)
+	}
+	return err
+}
+
+type testDriver struct{}
+
+func (d *testDriver) ConfigEnv() (envvar.List, error) {
+	return nil, nil
+}
+
+func (d *testDriver) CreateStorage(*imageregistryv1.Config) error {
+	panic("CreateStorage not implemented")
+}
+
+func (d *testDriver) ID() string {
+	panic("ID not implemented")
+}
+
+func (d *testDriver) RemoveStorage(*imageregistryv1.Config) (bool, error) {
+	panic("RemoveStorage not implemented")
+}
+
+func (d *testDriver) StorageChanged(*imageregistryv1.Config) bool {
+	panic("StorageChanged not implemented")
+}
+
+func (d *testDriver) StorageExists(*imageregistryv1.Config) (bool, error) {
+	panic("StorageExists not implemented")
+}
+
+func (d *testDriver) VolumeSecrets() (map[string]string, error) {
+	panic("VolumeSecrets not implemented")
+}
+
+func (d *testDriver) Volumes() ([]corev1.Volume, []corev1.VolumeMount, error) {
+	volumes := []corev1.Volume{
+		{
+			Name: defaults.ImageRegistryPrivateConfiguration,
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: defaults.ImageRegistryPrivateConfiguration,
+				},
+			},
+		},
+	}
+	return volumes, []corev1.VolumeMount{}, nil
+}
--- a/pkg/resource/deployment_test.go
+++ b/pkg/resource/deployment_test.go
@@ -65,7 +65,23 @@ func TestChecksum(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 
-			kubeClient := fake.NewSimpleClientset(test.origSecret)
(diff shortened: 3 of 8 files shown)


joelddiaz:
/hold
this depends on https://github.com/openshift/image-registry/pull/257

pkg/storage/s3/s3_test.go
```diff
@@ -48,15 +50,19 @@ func TestGetConfig(t *testing.T) {
 	})
 	listers := testBuilder.BuildListers()
 
-	config, err := GetConfig(listers)
+	s3Driver := &driver{
+		Listers: listers,
```
dmage, line 54: `GetCloudCredentials` is not set, the test panics because of this

pkg/storage/s3/s3_test.go
```diff
@@ -157,18 +166,26 @@ func TestConfigEnv(t *testing.T) {
 	listers := testBuilder.BuildListers()
 
 	d := NewDriver(ctx, config, listers)
+	defer func() {
+		conf, err := d.GetConfig()
```
dmage, line 170: This will create a new config file, why do you need this?

pkg/storage/s3/s3.go
```diff
@@ -30,12 +33,16 @@ import (
...
 type S3 struct {
-	AccessKey      string
-	SecretKey      string
-	Bucket         string
-	Region         string
-	RegionEndpoint string
+	Bucket                string
+	Region                string
+	RegionEndpoint        string
+	SharedCredentialsFile string
```
dmage, line 45: Let's get rid of this type, it confuses people.

pkg/storage/s3/s3.go
```diff
@@ -77,47 +89,63 @@ func GetConfig(listers *regopclient.Listers) (*S3, error) {
...
 
-	return cfg, nil
+	return sharedCredsFile, nil
 }
 
-func (d *driver) updateConfigAndGetCredentials() (accessKey, secretKey string, err error) {
-	cfg, err := GetConfig(d.Listers)
+// updateConfigAndGetCredentialsLocation will return the location of an AWS
+// credentials config file that the caller is responsible for cleaning up.
+func (d *driver) updateConfigAndGetCredentialsLocation() (string, error) {
```
dmage, line 145: Let's split this function into two:

* updateConfig
* getCredentialsLocation

I think the only reason why we have it as a single function so that people don't forget to call `updateConfig` before using it. Perhaps it should be checked via tests.

That'll help us to get rid of the local type `S3`.

pkg/storage/s3/s3.go
```diff
@@ -238,36 +274,75 @@ func (d *driver) ConfigEnv() (envs envvar.List, err error) {
...
-	return nil, nil
+	// Return the same credentials data that the image-registry-operator is using
+	// so that it can be stored in the image-registry Pod's Secret.
+	cfg, err := d.GetConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to retrieve S3 driver config: %v", err)
+	}
+	defer os.Remove(cfg.SharedCredentialsFile)
+
+	data, err := ioutil.ReadFile(cfg.SharedCredentialsFile)
```
dmage, line 339: This is a strange exercise: create a file, read its content, and delete the file.
I don't have suggestions, but there is room for improvement.

pkg/storage/s3/s3.go
```diff
@@ -643,3 +718,58 @@ func (d *driver) RemoveStorage(cr *imageregistryv1.Config) (bool, error) {
...
+
+func saveSharedCredentialsFile(data []byte) (string, error) {
+	f, err := ioutil.TempFile("", "aws-shared-credentials")
+	if err != nil {
+		return "", fmt.Errorf("failed to create file for shared credentials: %v", err)
+	}
+	defer f.Close()
+
+	if _, err := f.Write(data); err != nil {
+		return "", fmt.Errorf("failed to write credentials to %s: %v", f.Name(), err)
```
dmage, line 758: You'll have an empty file in this case (or with partially written data).

pkg/storage/s3/s3.go
```diff
@@ -45,23 +52,28 @@ type driver struct {
 
 	// roundTripper is used only during tests.
 	roundTripper http.RoundTripper
+
+	// used during tests to insert custom credentials
+	GetCloudCredentials func(*regopclient.Listers) (string, error)
```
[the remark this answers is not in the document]
dmage: btw, do we need it? We mock listers to provide custom data.

pkg/storage/s3/s3.go
```diff
@@ -643,3 +718,58 @@ func (d *driver) RemoveStorage(cr *imageregistryv1.Config) (bool, error) {
...
+}
+
+// sharedCredentialsFileFromStaticCreds takes an access/secret key pair and creates an AWS
+// configuration file using the provided keys.
+// It returns the path to the file containing the config data.
+func sharedCredentialsFileFromStaticCreds(accessKey string, accessSecret string) (string, error) {
+	buf := &bytes.Buffer{}
+	fmt.Fprint(buf, "[default]\n")
+	fmt.Fprintf(buf, "aws_access_key_id = %s\n", accessKey)
+	fmt.Fprintf(buf, "aws_secret_access_key = %s\n", accessSecret)
```
dmage, line 770: Can it contain special symbols? Does the value need to be escaped?

dmage (changes requested):
Is there a way to be sure that the operator always deletes temporary files? By reviewing this code I found one place where the operator can create a file and keep it. I'm not sure that I found all such cases. If the operator doesn't delete files, eventually it may run out of inodes and stop working. If we are going to create files, we need somehow verify that the operator doesn't have leaked temporary files after e2e-aws and e2e-aws-operator tests.

joelddiaz:
> Is there a way to be sure that the operator always deletes temporary files? 

One thing we can possibly do is generate the temporary filename once per driver initialization (I think I saw that there would be at most two driver initializations) and keep overwriting the file when creating the AWS session/client.
That does sort of complicate the e2e tests as we will create one temp file for every driver initialization (so perhaps have the e2e tests reach into the driver/struct to clean up the created file...), but it simplifies the actual in-cluster runtime situation.
WDYT?

pkg/storage/s3/s3.go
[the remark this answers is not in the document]
joelddiaz: :+1:  It was a bit of a mess to use the listers because the in-cluster pieces still need working AWS configs (which look different than the out-of-cluster AWS clients), so I modified it so that the e2e builds it's own AWS client and doesn't rely on running the S3 driver locally to query the state of things.

pkg/storage/s3/s3_test.go
joelddiaz: This is some remnants from an earlier approach where the S3 driver would keep only one copy of the AWS config/creds file. Removing it now...

pkg/storage/s3/s3.go
joelddiaz: I can't find a definitive list of allowed characters for the secret key. I do see the occasional '+', '/', '[', ']' and parenthesis, but nothing that would require escaping. I'll update it to use a single Fprintf() with backticks just to be on the safe side.

joelddiaz:
@dmage i've tried to address your PR feedback. let me know (especially the re-working of the S3 driver to get rid of the S3 struct and the way the config state is maintained).

joelddiaz:
fixed merge conflicts from #638 having merged

ricardomaraschini (changes requested):
Looks good, I have left a few comments for clarification. PTAL

manifests/01-registry-credentials-request.yaml
```diff
@@ -31,10 +31,10 @@ spec:
       - s3:GetLifecycleConfiguration
       - s3:GetBucketLocation
       - s3:ListBucket
-      - s3:HeadBucket
```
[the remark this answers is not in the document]
joelddiaz: I only discovered this while digging into the issues I was having before I realized I needed to add the new s3:ListMultipartUploadParts permission when running with STS credentials ¯\_(ツ)_/¯

manifests/07-operator.yaml
```diff
@@ -80,3 +83,12 @@ spec:
...
               path: tls-ca-bundle.pem
+        # This service account token can be used to provide identity outside the cluster.
+        # For example, this token can be used with AssumeRoleWithWebIdentity to authenticate
+        # with AWS using an IAM OIDC provider and STS.
+        - name: bound-sa-token
+          projected:
+            sources:
+            - serviceAccountToken:
+                path: token
+                audience: openshift
```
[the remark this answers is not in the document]
joelddiaz: This is what makes the entire web of trust work between AWS and the Pod running inside the cluster. You configure an AWS IAM Identity Provider that trusts identities that the OpenShift cluster can validate/sign. So now the Pod can make an iam:STS API call and say "here is my token signed by an OpenID provider that you trust". Then AWS validates that the token is indeed signed by something that it trusts, and (assuming other things check out properly) AWS will hand back temporary credentials.

joelddiaz:
@ricardomaraschini thanks for the review. FYI, I've put in two new commits in the PR addressing your feedback

manifests/01-registry-credentials-request.yaml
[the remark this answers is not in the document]
ricardomaraschini: Could you confirm that this permission removal won't affect our calls to `HeadBucketWithContext` even when using access key and secret? Just trying to be on the safe side here :-)

manifests/07-operator.yaml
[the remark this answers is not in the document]
ricardomaraschini: Ah, I understand now. Thanks!

pkg/storage/s3/s3_test.go
```diff
@@ -100,16 +105,19 @@ func TestGetConfigCustomRegionEndpoint(t *testing.T) {
...
 
-	expected := &S3{
-		AccessKey:      "access",
-		SecretKey:      "secret",
-		Region:         "example",
-		RegionEndpoint: "https://s3.example.com",
+	expected := &imageregistryv1.ImageRegistryConfigStorageS3{
+		Region:             "example",
+		RegionEndpoint:     "https://s3.example.com",
+		VirtualHostedStyle: true,
```
ricardomaraschini, line 120: Was this test wrong before? I mean, why do we expect `VirtualHostedStyle` to be true now?

manifests/01-registry-credentials-request.yaml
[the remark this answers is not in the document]
joelddiaz: s3:HeadBucket is not a real permission. Manually adding "s3:HeadBucket" to an IAM policy (because you can't select it through the GUI) and viewing the policy in the GUI visual editor shows the following message:
```
 Unrecognized actions
IAM does not recognize one or more actions. The action name might include a typo or might be part of a previewed or custom service.
s3:HeadBucket
```

pkg/storage/s3/s3_test.go
joelddiaz: I hope I can explain this :)

The current test in the master branch calls and tests s3.GetConfig(), but the actual s3 package really uses s3.updateConfigAndGetCredentials() (which in turn does call s3.GetConfig()). And in s3.updateConfigAndGetCredentials() there is an extra step where the VirtualHostedStyle is set on the config depending on various values. https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/s3/s3.go#L124-L129

The restructured code now has the single s3.UpdateEffectiveConfig() call to do the same work that s3.updateConfigAndGetCredentials() used to do. And the new s3.UpdateEffectiveConfig() now does the setting of the VirtualHostedStyle that only s3.updateConfigAndGetCredentials() used to do.

So now the test case in this PR is updated to call/test s3.UpdateEffectiveConfig() and the VirtualHostedStyle value is being set. And that is why I had to update the test case.

I honestly don't know what this value of VirtualHostedStyle accomplishes, and I just tried to preserve the existing behavior. I spent a lot of time trying to figure out why it is necessary to set VirtualHostedStyle, but I don't believe I was able to fully understand it.

As I re-read the code just now to write up this reply, I see that the current s3.updateConfigAndGetCredentials() actually tests the condition for both Config.Region and Config.RegionEndpoint being unset before it goes down the code path to potentially set Config.VirtualHostedStyle, but my restructuring only checks against Region being unset. I *think* it doesn't affect the way things work as this code really only ever runs the first time we are populating the config, but I could change it to more faithfully reproduce what the current code does.

WDYT?

ricardomaraschini: To be honest I also don't fully understand the impacts of VirtualHostedStyle and when we should leverage it and when we don't (it seems like the user can activate it manually as well). I know it has something to do with the way we reach the bucket: with VirtualHostedStyle the bucket is accessible through `<bucketname>.s3.amazon.com` instead of `s3.amazonaws.com/<bucketname>`. That being said I think we can agree that the least we change the current behavior the better, specially because @dmage, that would be able to explain us exactly what that means and in what circumstances we need to use it, is off. BTW, I think that the access without VirtualHostedStyle (for new buckets) has been deprecated: https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

joelddiaz: :+1: i put back the condition to match the previous behavior so only if Region and RegionEndpoint were zero-length would it go down that codepath.

joelddiaz:
@dmage @ricardomaraschini can we get a re-review based on the last changes addressing PR feedback?

sdodson:
/retitle Bug 1921909: build and use AWS config file for operator and operand

manifests/07-operator.yaml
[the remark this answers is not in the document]
ricardomaraschini: @joelddiaz Isn't this the token for the `cluster-image-registry-operator` service account? If this is the case do we still need to explicitly mount it? As we are not explicitly opting out of `automountServiceAccountToken` the token is already mounted under `/var/run/secrets/kubernetes.io/serviceaccount`. Please clarify.

pkg/resource/deployment_test.go
```diff
@@ -0,0 +1,220 @@
...
+				driver:          &testDriver{},
+				coreClient:      kubeClient.CoreV1(),
+				proxyLister:     proxyLister,
+				cr:              &imageregistryv1.Config{},
+				configMapLister: cmLister,
+				secretLister:    secretLister,
+			}
+
+			cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().ConfigMaps().Informer().HasSynced)
+			cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().Secrets().Informer().HasSynced)
```
ricardomaraschini, line 108: Can we user a context with a timeout here? If these caches never get ready we won't know exactly what happen other the whole test timing out. Something like below would make it more clear (using a context with Timeout):

```
if !cache.WaitForCacheSync(
		ctx.Done(),
		kubeInformer.Core().V1().ConfigMaps().Informer().HasSynced,
		kubeInformer.Core().V1().Secrets().Informer().HasSynced,
	) {
		t.Fatal("caches not syncing")
	}
```

pkg/resource/deployment_test.go
```diff
@@ -0,0 +1,239 @@
...
+			origHash := generatedDeployment.Annotations[defaults.ChecksumOperatorAnnotation]
+
+			if test.updatedSecret != nil {
+				_, err = kubeClient.CoreV1().Secrets(defaults.ImageRegistryOperatorNamespace).Update(context.TODO(), test.updatedSecret, metav1.UpdateOptions{})
+				if err != nil {
+					t.Errorf("failed to update secret: %s", err)
+					t.FailNow()
+				}
+
+				cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().Secrets().Informer().HasSynced)
```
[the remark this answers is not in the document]
ricardomaraschini: Yeah, I think this line can be removed.

pkg/resource/deployment_test.go
```diff
@@ -0,0 +1,220 @@
...
+
+			if test.updatedSecret != nil {
+				_, err = kubeClient.CoreV1().Secrets(defaults.ImageRegistryOperatorNamespace).Update(ctx, test.updatedSecret, metav1.UpdateOptions{})
+				if err != nil {
+					t.Fatalf("failed to update secret: %s", err)
+				}
+
+				cache.WaitForCacheSync(ctx.Done(), kubeInformer.Core().V1().Secrets().Informer().HasSynced)
+				// Fixme: why is this wait necessary?
+				if err := waitForUpdatedSecret(ctx, kubeClient, test.updatedSecret.Data); err != nil {
```
ricardomaraschini, line 131: nit: this should not be necessary, does the test fail if you remove this wait?

pkg/resource/podtemplatespec.go
```diff
@@ -318,6 +329,32 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
...
+					{
+						ServiceAccountToken: &corev1.ServiceAccountTokenProjection{
+							Audience: "openshift",
+							Path:     "token",
+						},
+					},
+				},
+			},
+		},
+	}
```
ricardomaraschini, line 347: I guess that the answer at https://github.com/openshift/cluster-image-registry-operator/pull/640/files#r566756238 will also solve my question regarding this other service account token :-)

pkg/resource/deployment_test.go
joelddiaz: It's been a while since I touched this, but I believe I added it because I was running into issue otherwise.

manifests/07-operator.yaml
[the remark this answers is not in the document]
joelddiaz: The contents of the two ServiceAccounts are slightly different.

Possibly most importantly the keyId field in the kubernetes.io/serviceaccount location doesn't match what is uploaded to the OIDC S3 bucket:

Standard ServiceAccount token:
`sh-4.4$ cat /var/run/secrets/kubernetes.io/serviceaccount/token | awk -F. '{ print $1 }' | base64 -d
{"alg":"RS256","kid":"RjUCNiAxtfGZcTjdVWLvNyksvfhTuQ7dmqAxBrbZIYA"}base64: invalid input
`
Projected ServiceAccount token:
`sh-4.4$ cat /var/run/secrets/openshift/serviceaccount/token | awk -F. '{ print $1 }' | base64 -d
{"alg":"RS256","kid":"lQZjBZuL5Xzq0B6Ih3VQVWFi6zVPE8u7fq_yubth2GQ"}base64: invalid input
`

And looking at the OIDC keys.json, you see the missmatch:

```
[jdiaz@minigoomba ~]$ curl https://jdiaz-stsfull-installer.s3.amazonaws.com/keys.json | jq -r '.keys[].kid'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   917  100   917    0     0   2345      0 --:--:-- --:--:-- --:--:--  2351
lQZjBZuL5Xzq0B6Ih3VQVWFi6zVPE8u7fq_yubth2GQ
```

Looking at the second field of the token there are more differences:

Standard ServiceAccount token:
`sh-4.4$ cat /var/run/secrets/kubernetes.io/serviceaccount/token | awk -F. '{ print $2 }' | base64 -d
{"iss":"kubernetes/serviceaccount","kubernetes.io/serviceaccount/namespace":"openshift-image-registry","kubernetes.io/serviceaccount/secret.name":"cluster-image-registry-operator-token-jbtdj","kubernetes.io/serviceaccount/service-account.name":"cluster-image-registry-operator","kubernetes.io/serviceaccount/service-account.uid":"30278254-d1bc-4631-ad41-777cbef7bb39","sub":"system:serviceaccount:openshift-image-registry:cluster-image-registry-operator"}base64: invalid input`

Projected ServiceAccount:
`sh-4.4$ cat /var/run/secrets/openshift/serviceaccount/token | awk -F. '{ print $2 }' | base64 -d
{"aud":["openshift"],"exp":1611938123,"iat":1611934523,"iss":"https://s3.us-east-1.amazonaws.com/jdiaz-stsfull-installer","kubernetes.io":{"namespace":"openshift-image-registry","pod":{"name":"cluster-image-registry-operator-8cb64687b-bpdp9","uid":"5050f308-9c1d-4c3c-89a4-8c454d0e08f6"},"serviceaccount":{"name":"cluster-image-registry-operator","uid":"30278254-d1bc-4631-ad41-777cbef7bb39"}},"nbf":1611934523,"sub":"system:serviceaccount:openshift-image-registry:cluster-image-registry-operator"}`

Here the issuer field (iss) differs, there is a missing audience (aud), no expiration (exp) or issued-at (iat).

But I do agree that they are fundamentally the same as you can see the serviceaccount.uid match.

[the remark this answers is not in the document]
ricardomaraschini: Thanks for clarifying this aspect. Two distinct tokens indeed :-)

pkg/resource/deployment_test.go
ricardomaraschini: Well, this is not a no go condition, let's just move this forward.

ricardomaraschini:
/unhold

https://github.com/openshift/image-registry/pull/257 is already merged.

dmage:
This is a new feature and there is a strong desire to have it in 4.7.0. As it's a feature, it should be approved (i.e. it should have xxx-approved labels), so please take a look.

/assign @wzheng1
It should be possible to pull/push images into the registry when OpenShift is installed in a regular way. The registry pods should have AWS secret and access key in their `/var/run/secrets/cloud/credentials` file (this file didn't exist before this PR).
It should also be possible to pull/push images when OpenShift installed [with with an STS credential](https://github.com/openshift/enhancements/blob/master/enhancements/cloud-integration/aws/aws-sts-support.md). In this case the registry pods should mention `/var/run/secrets/openshift/serviceaccount/token` in their `/var/run/secrets/cloud/credentials` file.

/assign @bmcelvee
As I can see, we don't need to change anything regarding the registry, all changes are about the installer, cloud-credentials-operator, and CVO. But maybe STS credentials needs somehow be mentioned in the registry docs.

/assign @sferich888
Please take a look.

See also:

* https://issues.redhat.com/browse/CCO-52
* https://github.com/openshift/enhancements/blob/master/enhancements/cloud-integration/aws/aws-sts-support.md
* https://github.com/joelddiaz/enhancements/blob/sts-by-default/enhancements/cloud-integration/aws/aws-sts-by-default.md#installer-manifests

bparees:
@wzheng1 @bmcelvee @sferich888 as we need this prior to codefreeze for 4.7 (feb 5th), please prioritize this as much as possible.

joelddiaz:
> See also:
> 
>     * https://issues.redhat.com/browse/CCO-52
> 
>     * https://github.com/openshift/enhancements/blob/master/enhancements/cloud-integration/aws/aws-sts-support.md
> 
>     * https://github.com/joelddiaz/enhancements/blob/sts-by-default/enhancements/cloud-integration/aws/aws-sts-by-default.md#installer-manifests

I would refer to https://github.com/openshift/cloud-credential-operator/pull/290 for the instructions on how to install in STS-mode (not a generally supported feature FWIW, more of where we intend to take AWS credentials in the future).

For this PR, when a cluster is installed in traditional Mint mode (what CI e2e does), the difference is that the Secret containing the AWS credentials will show an AWS configuration that happens to include the aws_access_key_id and aws_secret_access_key fields.

```
$ oc get secrets -n openshift-image-registry installer-cloud-credentials -o json | jq -r .data.credentials | base64 -d
[default]
aws_access_key_id  = AKACCESSKEYID
aws_secret_access_key = SECRET_ACCESS_KEY
```

And in STS mode, it will reference the AWS IAM Role name to assume, and the location of the ServiceAccount token to AssumeRole() with.
```
$ oc get secrets -n openshift-image-registry installer-cloud-credentials -o json | jq -r .data.credentials | base64 -d
[default]
role_arn = arn:aws:iam::123456789012:role/jdiaz-ststest-image-registry
web_identity_token_file = /var/run/secrets/openshift/serviceaccount/token
```

wzheng1:
@dmage @bparees Get it, QE is working on the test now.

gregsheremeta:
ping 

@dmage can you review the TCs?

Anything else outstanding before this can merge?

manifests/07-operator.yaml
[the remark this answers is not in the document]
bmcelvee: Is this automatically set, or does the user need to do anything manually?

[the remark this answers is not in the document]
joelddiaz: The Deployment for the image-registry-operator is permanently changed to start projecting in the ServiceAccount to this /var/run/secrets/openshift/serviceaccount/token location. OpenShift handles signing the token and updating/rotating it as needed automatically. Whether or not we use the ServiceAccount token, it will be there going forward (even for non-AWS clusters).

dmage:
@wzheng1 tests look good, should I add `tc-approved` somewhere?

@gregsheremeta we need 3 approvals: qe-approved from @wzheng1, docs-approved from @bmcelvee, px-approved from @sferich888

test/e2e/aws_test.go
```diff
@@ -694,30 +703,140 @@ func TestAWSFinalizerDeleteS3Bucket(t *testing.T) {
 	if err != nil {
 		t.Errorf("unable to get image registry resource: %#v", err)
 	}
-	driver, err := storage.NewDriver(&cr.Spec.Storage, kcfg, mockLister)
+
+	// Create an AWS config using the in-cluster credentials so that we can watch the S3 bucket
+	imageRegistryPrivateConfiguration, err := te.Client().Secrets(defaults.ImageRegistryOperatorNamespace).Get(
```
dmage, line 708: @joelddiaz this test fails, you are trying to access this Secret after the operator's config is deleted (at this point the operator is finalizing the config resource and deleting dependant resources).

joelddiaz: @dmage IIUC, I suppose I can move grabbing the Secret's content above line 693 where the Configs().Delete() is called.

gregsheremeta:
ping @dmage can you re-review?

PR #643, opened 2020-11-19
Bug 1948714: IBM Cloud manifest profile patch

csrwng:
Part 2 of cluster profiles support for IBM Cloud.
Introduces a directory that will contain profile-specific patches.
Profile-specific manifests are generated with 'make profile-manifests'
and 'hack/verify-profile-manifests.sh' (called by 'make verify') verifies
that the manifests directory is up to date.

On IBM Cloud the Image Registry operator needs to run on worker nodes, thus the
master node selector needs to be removed from the operator deployment
manifest.


csrwng:
/hold
This should only merge when profiles support has been turned on in the CVO

hack/update-profile-manifests.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
```
dmage, line 1: Is this script specific for this repo of do we have it copied to multiple repos?

csrwng: Hi @dmage so far the thinking is that it would be copied to multiple repos, given that we don't have a good way of sharing scripts across multiple components. It could be placed in library-go, but not all components use library-go. Also not all components use the same directories for their manifests. I'm definitely open to ideas on how to better handle this.

dmage: @csrwng what about https://github.com/openshift/build-machinery-go ? `cluster-image-registry-operator` uses only one target from build-machinery, so it shouldn't be a problem to add a single target to other repositories.

See https://github.com/openshift/cluster-image-registry-operator/blob/e5877a88fa9d13a7b69075595c021fce5a884333/Makefile#L9-L13

You can add something similar:

```
$(call add-profile-manifests,./manifests/,./profile-patches/) 
```

csrwng: That sounds promising, will try it out. Thx!

csrwng: Submitted https://github.com/openshift/build-machinery-go/pull/26 - will update this PR when that merges

csrwng:
@dmage I've now updated this PR to bump to the latest build-machinery-go and I'm now using that add a patch for ibm-cloud-managed.

Makefile
```diff
@@ -6,12 +6,17 @@ GOLANGCI_LINT = _output/tools/golangci-lint
 GOLANGCI_LINT_CACHE = $(PWD)/_output/golangci-lint-cache
 GOLANGCI_LINT_VERSION = v1.24
 
+GO_REQUIRED_MIN_VERSION = 1.14
```
ricardomaraschini, line 9: Is there any special reason for not going with a newer version? I have noticed [here](https://github.com/openshift/build-machinery-go/blob/master/make/lib/golang.mk#L20-L23) that if it is not set we go for 1.15.2.

manifests/07-operator-ibm-cloud-managed.yaml
```diff
@@ -0,0 +1,86 @@
+apiVersion: apps/v1
```
ricardomaraschini, line 1: Would you mind explaining how this is going to work now? Seems like we have two distinct deployments now in `manifests` directory. Isn't CVO going to create two deployments? Please clarify.

vendor/github.com/openshift/build-machinery-go/make/lib/doc.go
```diff
@@ -0,0 +1,3 @@
+// required for gomod to pull in packages.
```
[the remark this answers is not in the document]
csrwng: @ricardomaraschini it *is* in 2 separate commits

manifests/07-operator-ibm-cloud-managed.yaml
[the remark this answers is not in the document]
csrwng: Basically for any given installation of openshift only one cluster profile is active. The CVO will only apply manifests that contain an annotation for the active cluster profile.
Ref: https://github.com/openshift/cluster-version-operator/pull/404
Enhancement: https://github.com/openshift/enhancements/pull/200

Makefile
csrwng: The reason I added this here, is that verify previously failed (looks like your CI test is using an older go version):
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cluster-image-registry-operator/643/pull-ci-openshift-cluster-image-registry-operator-master-verify/1376950809396252672
Note that this is the **minimum** version. It's fine to use a newer release.

ricardomaraschini (approved):
/lgtm

PR #644, opened 2020-11-24
Bug 1889921: Reporting degraded if not available

ricardomaraschini:
Report this operator as Degraded if there is no replica or less replicas
then specified on the configuration for more than one minute.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"fmt"
+	"time"
 
 	appsapi "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
@@ -10,6 +11,7 @@ import (
 
 	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
 	operatorapiv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	"github.com/openshift/cluster-image-registry-operator/pkg/defaults"
 	"github.com/openshift/cluster-image-registry-operator/pkg/metrics"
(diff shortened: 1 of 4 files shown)


pkg/operator/status.go
```diff
@@ -294,6 +295,13 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
 
 	updateCondition(cr, operatorapiv1.OperatorStatusTypeProgressing, operatorProgressing)
 
+	if isDeploymentStatusAvailable(deploy) {
```
wking, line 298: Can't we look at `operatorAvailable.Status` here, instead of calling out to `isDeploymentStatusAvailable` again?  And do we need `.deployUnavailableSince` when we should already have `LastTransitionTime` on the `Available` status (although maybe not on `operatorAvailable` until we use some tooling syncs it into the existing `Conditions` object).

pkg/operator/status.go
```diff
@@ -309,6 +317,24 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
+			operatorDegraded.Reason = "NoReplicasAvailable"
+		}
+	} else if !isDeploymentStatusComplete(deploy) {
+		if c.deployIncompleteSince.IsZero() {
+			c.deployIncompleteSince = time.Now()
+		}
+		if time.Since(c.deployIncompleteSince) > time.Minute {
+			operatorDegraded.Status = operatorapiv1.ConditionTrue
+			operatorDegraded.Message = "The registry has minimum availability"
+			operatorDegraded.Reason = "MinimumAvailability"
```
wking, line 336: This reason/message sounds positive.  If we are going to set Degraded=True, we should say something negative to motivate the unhappy status.  Something like `Registry deployment has timed out progressing`.  But... can we lean on the Deployment controller for this and look at `ProgressDeadlineExceeded`?  See my rotted-due-to-lack-of-interest openshift/library-go#575 for one possible implementation.

pkg/operator/status.go
ricardomaraschini: Yes, this makes total sense. I will try to adjust according to your input, thanks for that.

ricardomaraschini: Done, PTAL.

pkg/operator/status.go
ricardomaraschini: Done, PTAL.

pkg/operator/status.go
```diff
@@ -309,6 +311,30 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
 	} else if cr.Spec.ManagementState == operatorapiv1.Removed {
 		operatorDegraded.Message = "The registry is removed"
 		operatorDegraded.Reason = "Removed"
+	} else if deploy == nil {
+		operatorDegraded.Status = operatorapiv1.ConditionTrue
+		operatorDegraded.Message = "The deployment does not exist"
+		operatorDegraded.Reason = "DeploymentNotFound"
```
dmage, line 317: If we shouldn't immediately react on absence of the deployment. Most likely the operator will fix this problem in a second.

pkg/operator/status.go
```diff
@@ -309,6 +311,30 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
+			operatorDegraded.Status = operatorapiv1.ConditionTrue
+			operatorDegraded.Message = updatedAvailableCondition.Message
+			operatorDegraded.Reason = "Unavailable"
+		}
+	} else if !isDeploymentStatusComplete(deploy) {
+		for _, cond := range deploy.Status.Conditions {
+			if cond.Type != appsapi.DeploymentProgressing {
+				continue
+			}
+			if cond.Reason != "ProgressDeadlineExceeded" {
```
dmage, line 330: If we want to rely on `progressDeadlineSeconds` of the registry deployment, I'd say we should be more opinionated and set it to 1 minute.

pkg/operator/status.go
ricardomaraschini: Yes, you are absolutely right. I have adjusted it by means of removing it, now it will take one minute as well. PTAL.

pkg/operator/status.go
ricardomaraschini: Done, PTAL.

PR #650, opened 2021-01-11
Bug 1915041: add ListMultipartUploadParts permission

joelddiaz:
s3:ListMultipartUploadParts is needed to be able to check on status of uploads. The permission is implied when the IAM user/credentials that created the MultiPartUploads is the same as the user/credentials trying to list them (see https://docs.aws.amazon.com/AmazonS3/latest/dev/mpuAndPermissions.html ).

When users/credentials are rotated, there can be a disconnect between the user/creds performing the ListMultipartUploadParts and the user/creds performing the uploads.

Stop depending on the ListMultipartUploadParts permission being implicitly granted, and just explicitly list this permission as required to allow credentials rotation without a risk of losing the ability to s3:ListMultipartuploadParts.

### DIFF
--- a/manifests/01-registry-credentials-request.yaml
+++ b/manifests/01-registry-credentials-request.yaml
@@ -37,4 +37,5 @@ spec:
       - s3:DeleteObject
       - s3:ListBucketMultipartUploads
       - s3:AbortMultipartUpload
+      - s3:ListMultipartUploadParts
       resource: "*"


joelddiaz:
Rather than allow this new permission to slip into a z-stream release, introduce the explicit permission as part of of 4.7.0 to avoid introducing permissions changes in a z-stream update.

/cc @sdodson

dmage:
/hold

The tests are blocked until https://github.com/openshift/openshift-apiserver/pull/177 lands

PR #658, opened 2021-01-25
Bug 1914407: Explicitly set node-ca runAsUser and runAsGroup

ricardomaraschini:
This commit adds runAsUser and runAsGroup directives to the node-ca daemonset configuration.


ricardomaraschini:
@dmage I have added also `runAsGroup` and tested in a cluster. Everything seems fine. PTAL

PR #660, opened 2021-02-05
Bug 1902076: Checking for route status

ricardomaraschini:
This commits adds a check for route statuses during operator status sync
process. If any of the routes was not Admitted the operator will report
itself as Degraded.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -21,6 +21,7 @@ import (
 
 	imageregistryv1 "github.com/openshift/api/imageregistry/v1"
 	operatorv1 "github.com/openshift/api/operator/v1"
+	routev1 "github.com/openshift/api/route/v1"
 	configclient "github.com/openshift/client-go/config/clientset/versioned"
 	configinformers "github.com/openshift/client-go/config/informers/externalversions"
 	imageregistryclient "github.com/openshift/client-go/imageregistry/clientset/versioned"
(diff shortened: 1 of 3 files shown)


pkg/operator/controller.go
```diff
@@ -248,7 +271,11 @@ func (c *Controller) sync() error {
 		deploy = deploy.DeepCopy() // make sure we won't corrupt the cached vesrion
 	}
 
-	c.syncStatus(cr, deploy, applyError)
+	routes, err := c.getRoutes(cr)
+	if err != nil {
+		return err
```
dmage, line 276: Failure to access one of the operator's routes shouldn't prevent the operator from reporting its status. Routes are expected to be absent when the operator is Removed (this blocks tests).

dmage:
/bugzilla refresh
I'd prefer to be on the safe side and don't want to fix it just before code freeze.

pkg/operator/controller.go
```diff
@@ -211,6 +212,29 @@ func (c *Controller) createOrUpdateResources(cr *imageregistryv1.Config) error {
 	return nil
 }
 
+// getRoutes returns a list of all routes configured for the image registry, including
+// the default route if configured.
+func (c *Controller) getRoutes(cr *imageregistryv1.Config) ([]*routev1.Route, error) {
+	var routes []*routev1.Route
+	if cr.Spec.DefaultRoute {
+		if route, err := c.listers.Routes.Get(defaults.RouteName); err != nil {
+			klog.V(5).Infof("unable to get default route: %s", err)
```
dmage, line 221: V(5) won't be visible even if you enable Debug logging. Only Trace or TraceAll that also includes HTTP requests from client-go.

pkg/operator/controller.go
ricardomaraschini: done, ptal.

pkg/operator/controller.go
ricardomaraschini: done, ptal.

PR #662, opened 2021-02-08
Mark emptyDir as not Upgradeable configuration

dmage:
When the registry uses emptyDir, it cannot be upgraded without losing its data.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -347,4 +347,23 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
 	}
 
 	updateCondition(cr, defaults.OperatorStatusTypeRemoved, operatorRemoved)
+
+	operatorUpgradeable := operatorapiv1.OperatorCondition{
+		Status:  operatorapiv1.ConditionTrue,
+		Message: "",
+		Reason:  "AsExpected",
+	}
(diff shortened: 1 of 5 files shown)


pkg/operator/status.go
```diff
@@ -347,4 +347,22 @@ func (c *Controller) syncStatus(cr *imageregistryv1.Config, deploy *appsapi.Depl
...
+	}
+	if cr.Spec.ManagementState == operatorapiv1.Unmanaged {
+		operatorUpgradeable.Message = "The registry configuration is set to unmanaged mode"
+		operatorUpgradeable.Reason = "Unmanaged"
+	} else if cr.Spec.ManagementState == operatorapiv1.Removed {
+		operatorUpgradeable.Message = "The registry is removed"
+		operatorUpgradeable.Reason = "Removed"
+	} else if cr.Spec.Storage.EmptyDir != nil {
+		operatorUpgradeable.Message = "The registry is configured to use emptyDir. The registry will lose pushed images once a registry pod is restarted (for example, due to a configuration change, a node restart, or an upgrade)"
+		operatorUpgradeable.Reason = "EmptyDir"
```
bparees, line 364: shouldn't this be setting it to false?

dmage: >_<
it should

dmage:
I also want to have tests

ricardomaraschini (approved):
/lgtm

PR #674, opened 2021-03-24
Bug 1937464: Swift: Do not modify driver config

Fedosin:
Driver config must be used for user input only and the operator shouldn't modify it using values from clouds.yaml.

This commit ensures that driver config stays unchanged, and it's used just to update default values provided by clouds.yaml.

### DIFF
--- a/pkg/storage/swift/swift.go
+++ b/pkg/storage/swift/swift.go
@@ -166,23 +166,21 @@ func (d *driver) getSwiftClient() (*gophercloud.ServiceClient, error) {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


dmage:
/hold

I think now I understand why the current approach doesn't work. But this fix will give us another problem.

spec is intended to be editable only by customers if they want to provide their own storage, the operator shouldn't override them. The operator can default settings, but it shouldn't change them.

If we've found that these settings are dynamic, perhaps the operator should keep spec empty. This will backward incompatible change, but as the current approach doesn't work anyway, maybe it's not that bad. Let me think about it.

PR #681, opened 2021-04-29
Bug 1945387: Setting required pod anti-affinity rules

ricardomaraschini:
Instead of using Preferred we now use Required pod anti-affinity. This
way we can enforce image-registry pods to do not run on the same node
at the same time, this at a price of restricting the max number of pods
to to the total amount of worker nodes.

### DIFF
--- a/pkg/resource/deployment.go
+++ b/pkg/resource/deployment.go
@@ -8,6 +8,7 @@ import (
 	appsapi "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	appsset "k8s.io/client-go/kubernetes/typed/apps/v1"
 	coreset "k8s.io/client-go/kubernetes/typed/core/v1"
 	appslisters "k8s.io/client-go/listers/apps/v1"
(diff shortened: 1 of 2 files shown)


ricardomaraschini:
/assign @dmage 

Still waiting for the tests but this is ready for a first look.

pkg/resource/podtemplatespec.go
```diff
@@ -23,6 +23,43 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/storage"
 )
 
+var (
+	// requiredAntiAffinity enforces one image registry pod per node.
+	requiredAntiAffinity = &corev1.Affinity{
```
dmage, line 28: Why do you need these global variables?

pkg/resource/deployment.go
```diff
@@ -95,8 +96,17 @@ func (gd *generatorDeployment) expected() (runtime.Object, error) {
 
 	// Strategy defaults to RollingUpdate
 	deployStrategy := gd.cr.Spec.RolloutStrategy
+	var rollingUpdate *appsapi.RollingUpdateDeployment
 	if deployStrategy == "" {
 		deployStrategy = string(appsapi.RollingUpdateDeploymentStrategyType)
+		if gd.cr.Spec.Replicas == 2 {
```
dmage, line 102: Do you want to do it only when `rolloutStrategy ` is not set?

pkg/resource/podtemplatespec.go
ricardomaraschini: done.

pkg/resource/deployment.go
ricardomaraschini: That was my initial thought but yeah, I guess we can always set it. Changing it, PTAL.

pkg/resource/podtemplatespec.go
```diff
@@ -367,27 +367,43 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
...
+		if cr.Spec.Replicas == 2 {
+			affinity = &corev1.Affinity{
+				PodAntiAffinity: &corev1.PodAntiAffinity{
+					RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{
+						{
+							TopologyKey: "kubernetes.io/hostname",
+							Namespaces: []string{
+								defaults.ImageRegistryOperatorNamespace,
+							},
+							LabelSelector: &metav1.LabelSelector{
```
dmage, line 399: Why do you need the label selector only when you require anti-affinity? Shouldn't the PodAffinityTerm be the same in both cases?

ricardomaraschini: I have played a little bit with these Affinity settings and the outcome is that we also need to include this LabelSelector on both. Adjusted accordingly. I am not even sure what was being requested without the LabelSelector in the first place. PTAL

dmage:
@ricardomaraschini I think it's better to squash commits, otherwise LGTM

ricardomaraschini:
> @ricardomaraschini I think it's better to squash commits, otherwise LGTM

done

PR #685, opened 2021-05-07
Bug 1952610: Bump dependencies

dmage:
To support new AWS regions, we need to use a newer version of aws-sdk-go.


dmage:
/hold

Tests are failing, needs investigation.

ricardomaraschini (approved):
/lgtm

PR #694, opened 2021-06-16
Bug 1973318: Properly set custom tolerations

hasueki:
This fixes properly setting custom tolerations defined in imagepruner CR to the CronJob spec.


dmage:
/retitle Bug 1973318: Properly set custom tolerations
/assign @ricardomaraschini

ricardomaraschini (approved):
/lgtm

PR #703, opened 2021-06-30
Start using embed for assets

Fedosin:
In go 1.16 they introduced embed module that allows to inject assets in the binary automatically. With this module we don't need to generate code using go-bindata anymore, which means this library can be safely removed.


ricardomaraschini (approved):
/lgtm

PR #705, opened 2021-07-12
Bug 1958376: Disallow blob public access for Azure storage account and require TLS1.2+

dmage:


xiuwang:
Launch cluster from this pr.
After installation successfully, check the storage account created by image registry in azure webconsole. It has set disallow blob public access and MinimumTLSVersion to 1.2.

imageregistrywxjaz7flkqs 
Blob public access. Disabled
Minimum TLS version Version 1.2

/label qe-approved

ricardomaraschini (approved):
/lgtm

PR #709, opened 2021-08-09
Bug 1981639: Update rolling update parameters

dmage:


pkg/resource/deployment.go
```diff
@@ -102,6 +102,26 @@ func (gd *generatorDeployment) expected() (runtime.Object, error) {
...
+		// 2. starts migration to the new pod template according to rolling
+		//    update parameters.
+		//
+		// To scale up from 2 replicas (when the registry pods have hard
+		// anti-affinity rules) to 6 replicas on a minimal cluster with 2
+		// worker nodes the deployment should tolerate 5 unavailable replicas:
+		//
+		//  * 4 replicas out of 6 cannot fit onto 2 workers,
+		//  * 1 replica should be deleted before a new one can be created.
+		maxUnavailable := intstr.FromInt(int(gd.cr.Spec.Replicas) - 1)
```
ricardomaraschini, line 119: Interesting to see what happens when gd.cr.Spec.Replicas is 1. That would inform that we can have a max of zero unavailable replicas. Judging by your comment I guess it will scale up the "old pod template" to 2 before starting the migration to the "new pod template", is that it? Just checking.

dmage: If replicas is 1, then `maxSurge = ceil(1*0.25) = 1`, so during upgrades the deployment will be allowed to create one new replica and won't be allowed to delete the old replica until the new one is ready.

You mean replicas is 1 and we want to upgrade? It should work even on single node installations and will have zero downtime.
2 -> 1: I guess it'll be scaled down first, and then we'll have at least one node which can be used for the registry pods.
1 -> 2: First we'll have two pods with preferred anti-affinity rules, then one pod will be allowed to be unavailable, so there will be at least one node for a new registry pod.

ricardomaraschini (approved):
/lgtm

PR #711, opened 2021-08-20
e2e: Dump build information on failure

adambkaplan:
If the test build fails, dump relevant build information to the log.

### DIFF
--- a/test/e2e/removed_test.go
+++ b/test/e2e/removed_test.go
@@ -44,6 +44,18 @@ func TestImageRegistryRemovedWithImages(t *testing.T) {
 		}
 	}()
 
+	if buildName, err := runTestBuild(ctx, te, nsName); err != nil {
+		te.Error(err)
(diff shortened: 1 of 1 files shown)


adambkaplan:
/hold

I think the build pod itself needs to be dumped. We hit the failure but it is hard to tell why the build is not trusting the internal registry.

adambkaplan:
@dmage I see a lot of failures on the `e2e-aws-operator` job - is this test broken?

dmage:
@adambkaplan it was broken, but 4 days ago we merged a fix for it. TestImageRegistryRemovedWithImages is flaky and it's exactly the test you're adding debug info to.

adambkaplan (commented):
/hold cancel

Build pod info is now dumped on failure.

dmage:
No customer facing changes.
/label docs-approved
/label px-approved

It shouldn't change CI pass rate, so no actions from QE are required.
/label qe-approved

PR #715, opened 2021-09-02
Bug 1972827: Avoid disruptions

dmage:
The registry should give endpoints time to propagate through the system. Otherwise requests will go to the pod that is shutting down.

See also:
* https://learnk8s.io/graceful-shutdown
* https://github.com/openshift/console-operator/commit/8010dd8093f8eb120fce1208fdd55eab69891558


ricardomaraschini (approved):
/lgtm

PR #721, opened 2021-09-22
Bug 1907433: Decrease severity for event queue messages

dmage:
These messages are not meaningful for customers and rarely used even for debugging.


ricardomaraschini (approved):
/lgtm

PR #722, opened 2021-09-24
Fix race in TestAWSFinalizerDeleteS3Bucket

dmage:
Once the registry config object is deleted, the testing framework cannot
predictably stop operator and delete its objects the operator will be in
the bootstrapping stage. Waiting until bootstrapping is finished will
prevent from concurrent deletion and creation of objects.


dmage:
/retitle Bug 2008096: Fix race in TestAWSFinalizerDeleteS3Bucket

dmage:
/override ci/prow/e2e-aws-operator
Cluster cannot recover for must-gather: https://bugzilla.redhat.com/show_bug.cgi?id=1981549

PR #728, opened 2021-11-10
IR-224: Bump openshift/api

flavianmissi:
also bumps a few related packages as well to fix compilation errors
in some of the vendored packages.

The only changes needed to support the new versions were adding
context to some calls. Used `context.TODO()` for that.

Closes https://issues.redhat.com/browse/IR-224

PR #730, opened 2021-11-12
IR-195: set pod topology spread constraint

flavianmissi:
Defaults to:

```
  topologySpreadConstraints:
  - maxSkew: 1 
    topologyKey: topology.kubernetes.io/zone 
    whenUnsatisfiable: DoNotSchedule
    labelSelector:
      matchLabels:
        docker-registry: default
  - maxSkew: 1 
    topologyKey: kubernetes.io/hostname 
    whenUnsatisfiable: DoNotSchedule
    labelSelector:
      matchLabels:
        docker-registry: default
  - maxSkew: 1 
    labelSelector:
      matchLabels:
        docker-registry: default
    topologyKey: node-role.kubernetes.io/worker
    whenUnsatisfiable: DoNotSchedule
```

Users are able to override it via ImageRegistry Spec. In that case, only the user provided constraints will apply (the default constraints are removed).

~With the default constraints, it is _not_ possible to schedule pods on master nodes, regardless of tolerations or node selectors added.~
When node selectors are specified, no default constraints are used. We do this to maintain current behaviour where it's possible to deploy the image registry to specific nodes that may otherwise be excluded by the topology constraints.

Relates to [IR-195](https://issues.redhat.com/browse/IR-195)
Depends on https://github.com/openshift/cluster-image-registry-operator/pull/728 and https://github.com/openshift/api/pull/1052

~NOTE: this branch has `flavianmissi:upgrade-openshift-api` rebased into it. Please ignore the vendored changes, they should be removed when https://github.com/openshift/cluster-image-registry-operator/pull/728 lands and we update openshift/api to a commit including https://github.com/openshift/api/pull/1052~


dmage:
/lgtm

/hold
until `replace` is removed after approvals

/assign @xiuwang @bmcelvee @sferich888 
for approvals

dmage:
/assign @bmcelvee @sferich888 
for Docs and Px approvals

pkg/resource/podtemplatespec.go
```diff
@@ -378,9 +378,14 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 		resources = *cr.Spec.Resources
 	}
 
+	nodes, err := coreClient.Nodes().List(context.TODO(), metav1.ListOptions{LabelSelector: "topology.kubernetes.io/zone"})
```
dmage, line 381: err is not checked

flavianmissi: fixed

pkg/resource/podtemplatespec.go
```diff
@@ -378,45 +378,83 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
...
+	if err != nil {
+		return corev1.PodTemplateSpec{}, deps, fmt.Errorf("Could not check nodes for zone failure domain: %s", err)
```
dmage, line 383:
```suggestion
		return corev1.PodTemplateSpec{}, deps, fmt.Errorf("could not check nodes for zone failure domain: %s", err)
```

KeenonLee:
Test again, nodeSelector and tolerations works well, thanks
/label qe-approved

dmage:
/label px-approved
IR-228 isn't marked as TE Needed

PR #734, opened 2021-11-25
Use go 1.17

dmage:
Replaces #729.


tools.go
```diff
@@ -1,3 +1,4 @@
+//go:build tools
```
dmage, line 1: This line is generated automatically by `gofmt`.

PR #736, opened 2021-12-09
Bug 2034484: Library go bump

eggfoobar:
Bumping the library-go dependency to latest, this should take advantage of leader election changes for SNO clusters proposed in this [library-go PR](https://github.com/openshift/library-go/pull/1242) and performance improvements in this [PR](https://github.com/openshift/library-go/pull/1235).

Changes:
- updated library-go to use latest master branch
- updated handler structs to use the new explicit handler structs


dmage:
Is it part of a Jira story or a BZ?

dmage:
TestLeaderElection needs to be adjusted, apparently the new leader election is slower and the test needs to wait until these pods join the election process.

eggfoobar:
> Is it part of a Jira story or a BZ?

The Jira story this is tied to is [CNF-3684](https://issues.redhat.com/browse/CNF-3684)

> TestLeaderElection needs to be adjusted, apparently the new leader election is slower and the test needs to wait until these pods join the election process.

That's interesting, for non-SNO clusters the timing of 137/107/26 should not have been changed. Looking into it

PR #739, opened 2022-01-04
IR-227: Remove fallback from platformStatus to installer config

dmage:
platformStatus should be available since 4.2, but due to bugs we
couldn't remove it in 4.2. The last known bug [1] was fixed in 4.7, so
it should be safe to remove the legacy path.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1890038


dmage:
/assign @ricardomaraschini 

No user visible changes.
/label docs-approved
/label px-approved

/assign @xiuwang 
@xiuwang do you know if we still have cases when platformStatus is not available on born-in-4.1 clusters?

xiuwang:
@dmage We have upgrade ci to trigger upgrade from 4.1 to 4.10 all the path. I will check this pr today.

xiuwang:
I installed a 4.1 aws cluster, could upgrade to 4.10.0-0.ci.test-2022-01-06-020610-ci-ln-013rftk-latest(pr build version) on full the path.

/lgtm
/label qe-approved

PR #741, opened 2022-01-14
no CredentialsRequests in ibm-cloud-managed

joelddiaz:
The ibm-cloud-managed profile doesn't use the cloud-credential-operator.
The CredentialsRequest CRs should not be installed in that environment.

Unmark the existing CredentialsRequests so they are no longer installed
in ibm-cloud-managed.

Create a list of previously-installed CredentialsRequests so that they
are cleaned up by the CVO only on the ibm-cloud-managed profile.

xref: https://issues.redhat.com/browse/CCO-177


dmage:
@joelddiaz to merge this, you need either a valid bug OR 3 approvals (from Docs, QE, and Px).

PR #751, opened 2022-02-03
Bug 2048214: Alibaba: adding permissions for using KMS encryption

kwoodson:
Adding necessary permissions to enable KMS key encryption in the registry.


dmage:
Thank you!
/lgtm

PR #760, opened 2022-03-22
IR-198: Add default storage type for Nutanix platform

thunderboltsid:
Set default storage type for Nutanix platform to removed.


pkg/storage/storage.go
```diff
@@ -183,6 +183,11 @@ func GetPlatformStorage(listers *regopclient.Listers) (imageregistryv1.ImageRegi
 	case configapiv1.AlibabaCloudPlatformType:
 		cfg.OSS = &imageregistryv1.ImageRegistryConfigStorageAlibabaOSS{}
 		replicas = 2
+	case configapiv1.NutanixPlatformType:
+		cfg.PVC = &imageregistryv1.ImageRegistryConfigStoragePVC{
```
dmage, line 187: We don't recommend to use PVC on new platforms. It's better to have the image registry removed by default (like we have for bare metal) or to use object storage.

thunderboltsid: Thanks for the review @dmage. Can you explain the rationale for the recommendation?

dmage:
/assign @bmcelvee @xiuwang @sferich888 
for approvals

xiuwang:
Can't install nutanix cluster so far, I will keep close with installer qe.

xiuwang:
Verified the image registry operator default to Removed with pr build image.
/label qe-approved

bmcelvee:
Note: no new content needs to be developed, but the _registry-removed_ module should be included in the installation guide. 

/label docs-approved

PR #762, opened 2022-03-30
Bug 2067995: Use OnRootMismatch for fsGroupChangePolicy

dmage:
This should help on installations that have millions of files on the registry volume.

PR #763, opened 2022-03-31
Bug 2067995: Deployment annotations, runtimeClassName override and fs policy change

ricardomaraschini:
We are using config.spec.unsupportedConfigOverrides to allow users to override image registry runtimeClass and annotations. This commit also uses OnRootMismatch for fsGroupChangePolicy in the registry deployment.

### DIFF
--- a/pkg/resource/configoverrides.go
+++ b/pkg/resource/configoverrides.go
@@ -1,0 +1,13 @@
+package resource
+
+// ConfigOverrides holds data users can set to override default object configurations created
+// by this operator. This is stored in the registry Config.Spec.UnsupportedConfigOverrides.
+type ConfigOverrides struct {
+	Deployment *DeploymentOverrides `json:"deployment,omitempty"`
+}
+
+// DeploymentOverrides holds items that can be overwriten in the image registry deployment.
+type DeploymentOverrides struct {
+	Annotations      map[string]string `json:"annotations,omitempty"`
+	RuntimeClassName *string           `json:"runtimeClassName,omitempty"`
+}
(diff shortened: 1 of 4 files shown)


pkg/resource/configoverrides.go
```diff
@@ -0,0 +1,13 @@
+package resource
+
+// ConfigOverrides holds data users can set to override default object configurations created
+// by this operator. This is stored in the registry Config.Spec.UnsupportedConfigOverrides.
+type ConfigOverrides struct {
+	Deploy *DeploymentOverrides `json:"deploy,omitempty"`
```
dmage, line 6: It's better to name it as `deployment`, don't use abbreviations in API

pkg/resource/configoverrides.go
```diff
@@ -0,0 +1,13 @@
...
+}
+
+// DeploymentOverrides holds items that can be overwriten in the image registry deployment.
+type DeploymentOverrides struct {
+	Annotations  map[string]string `json:"annotations,omitempty"`
+	RuntimeClass *string           `json:"runtimeClass,omitempty"`
```
dmage, line 12: runtimeClassName

pkg/resource/configoverrides.go
ricardomaraschini: fixed.

pkg/resource/configoverrides.go
ricardomaraschini: fixed

ricardomaraschini:
@dmage I have incorporated changes from https://github.com/openshift/cluster-image-registry-operator/pull/762 in here

PR #769, opened 2022-04-27
Bug 2009352: IR-253: add prometheus rules for image registry operations metrics

flavianmissi:
image

### DIFF
--- a/docs/metrics.md
+++ b/docs/metrics.md
@@ -1,0 +1,15 @@
+# Exported metrics
+
+## `imageregistry:operations_count:sum`
+
+| Operation | Resource type      |
+| --------- | ------------------ |
+| `get`     | `blob`, `manifest` |
+| `create`  | `blob`, `manifest` |
+
+## `imageregistry:imagestreamtags_count:sum`
+
+| Source     | Location             | Description                                                   |
+| ---------- | -------------------- | ------------------------------------------------------------- |
+| `imported` | `openshift`, `other` | Image Stream Tags imported in 'openshift' or other namespaces |
+| `pushed`   | `openshift`, `other` | Image Stream Tags pushed to 'openshift' or other namespaces   |
--- a/manifests/09-prometheus-rules-registry-operations.yaml
+++ b/manifests/09-prometheus-rules-registry-operations.yaml
@@ -1,0 +1,44 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: image-registry-rules
+  namespace: openshift-image-registry
+  annotations:
+    include.release.openshift.io/ibm-cloud-managed: "true"
+    include.release.openshift.io/self-managed-high-availability: "true"
+    include.release.openshift.io/single-node-developer: "true"
+spec:
+  groups:
+  - name: imageregistry.operations.rules
+    rules:
+    - expr: |
+        label_replace(
+          label_replace(
+            sum by (operation) (imageregistry_request_duration_seconds_count{operation="BlobStore.ServeBlob"}), "operation", "get", "operation", "(.+)"
+          ), "resource_type", "blob", "resource_type", ""
+        )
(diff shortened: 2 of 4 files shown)


manifests/09-prometheus-rules-registry-operations.yaml
```diff
@@ -0,0 +1,40 @@
...
+spec:
+  groups:
+  - name: imageregistry.operations.rules
+    rules:
+    - expr: |
+        label_replace(
+          label_replace(
+            sum by (operation) (imageregistry_request_duration_seconds_count{operation="BlobStore.ServeBlob"}), "operation", "get", "operation", "(.+)"
+          ), "resource_type", "blob", "resource_type", ""
+        )
```
flavianmissi, line 15: this feels kinda ugly but it's the only way I found of adding a new label (the outermost `label_replace`) and changing the value of a label (the innermost `label_replace`)

manifests/09-prometheus-rules-registry-operations.yaml
```diff
@@ -0,0 +1,40 @@
...
+      record: imageregistry:operations_sum:count
```
dmage, line 16: I think it should be `imageregistry:operations_count:sum`

flavianmissi: darn you're right, how did it end up like this

flavianmissi: fixed

manifests/09-prometheus-rules-registry-operations.yaml
ricardomaraschini: Why do we need to rename the labels ? If it is necessary then why not use a simple label_replace:

```
label_replace(
     sum by (operation) (imageregistry_request_duration_seconds_count{operation="ManifestService.Put"}), "operation", "manifestCreate", "operation", "(.+)"
)
```

flavianmissi: I found that separating resource from operation (i.e manifest / get) improved the overall metric readability.

flavianmissi:
I'm on it

docs/metrics.md
```diff
@@ -0,0 +1,15 @@
...
+| Operation | Resource type  |
+| --------- | -------------- |
+| `get`     | blob\|manifest |
+| `create`  | blob\|manifest |
+
+## `imageregistry:imagestreamtags_count:sum`
+
+| Source     | Location         | Description                                                     |
+| ---------- | ---------------- | --------------------------------------------------------------- |
+| `imported` | openshift\|other | Image Stream Tags imported in 'openshift' or 'other' namespaces |
```
xiuwang, line 14: The quotation marks for `other` may be misleading to think it named `other` namespace.

flavianmissi: Good point - removed.

dmage: I would keep quotation in the second column, but remove them from the description.

flavianmissi: hmm not sure I follow - do you mean the second column should look like:
```
'openshift'|'other'
```
or maybe did you mean we should use backticks, like the first column:
```
`openshift`|`other`
```
?

dmage: | Source     | Location         | Description                                                     |
| ---------- | ---------------- | --------------------------------------------------------------- |
| `imported` | `openshift`, `other` | Image Stream Tags imported in 'openshift' or other namespaces |

```
| Source     | Location         | Description                                                     |
| ---------- | ---------------- | --------------------------------------------------------------- |
| `imported` | `openshift`, `other` | Image Stream Tags imported in 'openshift' or other namespaces |
```

wdyt?

flavianmissi:
I think https://github.com/openshift/cluster-image-registry-operator/pull/775 has to merge before we can create a backport for this one

flavianmissi:
I mean https://github.com/openshift/cluster-image-registry-operator/pull/777 not 775

PR #770, opened 2022-04-28
Bug 2007611: Merge S3 CA bundle with system CA bundle

dmage:


pkg/storage/s3/s3.go
```diff
@@ -236,27 +238,48 @@ func (d *driver) getS3Service() (*s3.S3, error) {
...
+			RootCAs: rootCAs,
+		},
+		MaxIdleConns:          100,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   10 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+		ForceAttemptHTTP2:     true,
+	}
+
+	err = http2.ConfigureTransport(tr)
```
flavianmissi, line 271: We need to call this in the swift integration code as well[1].
[1] https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/swift/swift.go?plain=1#L225

PR #772, opened 2022-05-02
Bug 2009352: IR-120: Add prometheus rules for image stream tags rules

ricardomaraschini:
Create a record rule to collect the amount of image stream tags present
in the cluster.

### DIFF
--- a/manifests/09-prometheus-rules-imagestreams.yaml
+++ b/manifests/09-prometheus-rules-imagestreams.yaml
@@ -1,0 +1,11 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: imagestreams-rules
+  namespace: openshift-image-registry
+spec:
+  groups:
+  - name: imagestreams.rules
+    rules:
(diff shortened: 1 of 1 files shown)


manifests/09-prometheus-rules-imagestreams.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: PrometheusRule
+metadata:
+  name: imagestreams-rules
+  namespace: openshift-image-registry
+spec:
+  groups:
+  - name: imagestreams.rules
+    rules:
+    - expr: sum by (location, source) (image_registry_image_stream_tags_total)
+      record: imagestreamtags:count:sum
```
flavianmissi, line 11: does it make sense to prepend this with `imageregistry:`? i.e `imageregistry:imagestreamtags:count:sum`
additionally the other rule we're adding uses `_count` instead of `:count`. which one do we prefer?

ricardomaraschini: I prefer your approach. Have updated, PTAL.

PR #773, opened 2022-05-02
Fix storageClass name when creating PVC

MaysaMacedo:
Recently the CSI StorageClass named as standard-csi
was made the default in 4.11, consequently the PVC
creation would fail with storageclass.storage.k8s.io
standard not found. This commit fixes the issue by
ensuring the pvc uses the correct storageClass name.


pkg/operator/bootstrap.go
```diff
@@ -108,8 +108,8 @@ func (c *Controller) createPVC(accessMode corev1.PersistentVolumeAccessMode, cla
 		return err
 	}
 
-	// "standard" is the default StorageClass name, that was provisioned by the cloud provider
-	storageClassName := "standard"
+	// "standard-csi" is the default StorageClass name, that was provisioned by the cloud provider
```
jsafrane, line 111: very minor nit: I would mention that `standard-csi` is for 4.11 and newer

dmage:
/label qe-approved
CI results are sufficient to merge this.

/label docs-approved
/label px-approved
No user visible changes.

pkg/operator/bootstrap.go
MaysaMacedo: done

jsafrane:
lgtm from storage POV

mandre:
This issue causes all jobs running on vh-mecha to fail.
/lgtm

PR #784, opened 2022-06-02
Bug 2083466: Disable dualstack when it is not available

dmage:
In some AWS regions S3 does not have a dual-stack endpoint. In these regions the operator should fallback to the standard endpoint.

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -83,6 +83,22 @@ func (er *endpointsResolver) EndpointFor(service, region string, opts ...func(*e
 	return endpoints.DefaultResolver().EndpointFor(service, region, opts...)
 }
 
+func isUnknownEndpointError(err error) bool {
+	_, ok := err.(endpoints.UnknownEndpointError)
+	return ok
+}
+
+func regionHasDualStackS3(region string) (bool, error) {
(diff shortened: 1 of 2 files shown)


pkg/storage/s3/s3.go
```diff
@@ -224,6 +240,18 @@ func (d *driver) getCABundle() (string, error) {
...
 }
 
+// useDualStack returns true if the driver should use dual-stack endpoints
+func (d *driver) useDualStack() (bool, error) {
+	if d.Config.RegionEndpoint != "" {
+		return true, nil
+	}
+	ok, err := regionHasDualStackS3(d.Config.Region)
+	if err != nil {
+		return true, fmt.Errorf("failed to determine if region %s has dual stack S3: %w", d.Config.Region, err)
```
flavianmissi, line 250: probably a nit: maybe return `false` here with the error? although that value isn't used by the caller on errors, it would be more inline with what `regionHasDualStackS3` returns with an error.

flavianmissi:
left a comment but lgtm as is
/lgtm

pkg/storage/s3/s3.go
dmage: I thought it's safer to assume that is supports IPv6. But as it won't be used, it's better to return false there, otherwise it's just confusing.

I'll fix it, I need to fix linter errors too.

xiuwang:
/label qe-approved
Installed AWS, C2S and SC2S cluster successfully

PR #787, opened 2022-07-01
WIP: use readyz endpoint for readiness probes

rphillips:


flavianmissi:
any specific reason to target `release-4.11` instead of targeting `master` then back-porting it?

bparees:
i think we were investigating a regression in 4.11 when this PR was created so we were focused on that.  The regression has since been resolved, but i think this change is still a net positive and worth considering (in master.  not something we should backport at all, at this point, imho)

PR #790, opened 2022-07-25
Bug 2093440: Use actualDaemonSet for SetDaemonSetGeneration

neisw:
Continue to see failures in Managed cluster should start all core operator due to NodeCADaemonControllerDegraded.  Adding logging to investigate further

https://sippy.dptools.openshift.org/sippy-ng/jobs/4.12/runs?filters=%7B%22items%22%3A%5B%7B%22columnField%22%3A%22failed_test_names%22%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22%5Bsig-arch%5D%5BEarly%5D%20Managed%20cluster%20should%20start%20all%20core%20operators%20%5BSkipped%3ADisconnected%5D%20%5BSuite%3Aopenshift%2Fconformance%2Fparallel%5D%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22never-stable%22%7D%2C%7B%22columnField%22%3A%22variants%22%2C%22not%22%3Atrue%2C%22operatorValue%22%3A%22contains%22%2C%22value%22%3A%22techpreview%22%7D%5D%2C%22linkOperator%22%3A%22and%22%7D&sortField=timestamp&sort=desc

### DIFF
--- a/pkg/resource/nodecadaemon.go
+++ b/pkg/resource/nodecadaemon.go
@@ -18,7 +18,7 @@ import (
 	"github.com/openshift/library-go/pkg/operator/resource/resourceread"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
-	"github.com/openshift/cluster-image-registry-operator/bindata"
+	assets "github.com/openshift/cluster-image-registry-operator/bindata"
 	"github.com/openshift/cluster-image-registry-operator/pkg/defaults"
 )
 
@@ -70,19 +70,18 @@ func (ds *generatorNodeCADaemonSet) Create() (runtime.Object, error) {
 }
 
 func (ds *generatorNodeCADaemonSet) Update(o runtime.Object) (runtime.Object, bool, error) {
-	daemonSet := ds.expected()
+	desiredDaemonSet := ds.expected()
 
 	_, opStatus, _, err := ds.operatorClient.GetOperatorState()
 	if err != nil {
 		return nil, false, err
 	}
-
-	dep, updated, err := resourceapply.ApplyDaemonSet(
+	actualDaemonSet, updated, err := resourceapply.ApplyDaemonSet(
 		context.TODO(),
 		ds.client,
 		ds.recorder,
-		daemonSet,
-		resourcemerge.ExpectedDaemonSetGeneration(daemonSet, opStatus.Generations),
+		desiredDaemonSet,
+		resourcemerge.ExpectedDaemonSetGeneration(desiredDaemonSet, opStatus.Generations),
 	)
 	if err != nil {
 		return o, updated, err
@@ -90,7 +89,7 @@ func (ds *generatorNodeCADaemonSet) Update(o runtime.Object) (runtime.Object, bo
 
 	if updated {
 		updateStatusFn := func(newStatus *operatorv1.OperatorStatus) error {
-			resourcemerge.SetDaemonSetGeneration(&newStatus.Generations, daemonSet)
+			resourcemerge.SetDaemonSetGeneration(&newStatus.Generations, actualDaemonSet)
 			return nil
 		}
 
@@ -100,11 +99,11 @@ func (ds *generatorNodeCADaemonSet) Update(o runtime.Object) (runtime.Object, bo
 			updateStatusFn,
 		)
 		if err != nil {
-			return dep, updated, err
+			return actualDaemonSet, updated, err
 		}
 	}
 
-	return dep, updated, nil
+	return actualDaemonSet, updated, nil
 }
 
 func (ds *generatorNodeCADaemonSet) Delete(opts metav1.DeleteOptions) error {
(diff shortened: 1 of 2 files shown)


dmage:
Most failures are upgrade related, to fix them we need to backport NodeCADaemonProgressing to 4.11 (so that the installer waits until node-ca is fully deployed and become available).

neisw:
Thanks Oleg,  We wanted to see if we could determine what causes the caches not synced error to begin with as well as why it seems to take 10 minutes to recover.  Just wanting to try to get more insight into the root issues.

E0602 14:33:53.462338       1 nodecadaemon.go:86] NodeCADaemonController: unable to sync: failed to update object *v1.DaemonSet, Namespace=openshift-image-registry, Name=node-ca: daemonsets.apps "node-ca" is forbidden: caches not synchronized, requeuing
I0602 14:33:53.557931       1 recorder_logging.go:44] &Event{ObjectMeta:{dummy.16f4d45710871f42  dummy    0 0001-01-01 00:00:00 +0000 UTC <nil> <nil> map[] map[] [] []   
is encountered 

I0602 14:43:43.107649       1 generator.go:60] object *v1.DaemonSet, Namespace=openshift-image-registry, Name=node-ca updated: 
I0602 14:43:43.127904       1 generator.go:60] object *v1.ClusterOperator, Name=image-registry updated: removed:apiVersion="config.openshift.io/v1", removed:kind="ClusterOperator", changed:metadata.resourceVersion={"25916" -> "31395"}, changed:status.conditions.2.lastTransitionTime={"2022-06-02T14:33:53Z" -> "2022-06-02T14:43:43Z"}, removed:status.conditions.2.message="NodeCADaemonControllerDegraded: failed to update object *v1.DaemonSet, Namespace=openshift-image-registry, Name=node-ca: daemonsets.apps \"node-ca\" is forbidden: caches not synchronized", changed:status.conditions.2.reason={"NodeCADaemonControllerError" -> "AsExpected"}, changed:status.conditions.2.status={"True" -> "False"}

dmage:
You may need to dive into openshift-apiserver. I suspect that the reason is that the image registry operator publishes the registry name, the name gets observed by the openshift-apiserver operator, and it causes redeploy of openshift-apiserver. New openshift-apiserver instances need some time to fill their caches. Exponential backoff on the registry operator side can also cause additional delay in recovering after openshift-apiserver redeploy.

dmage:
err: daemon sets are upstream objects, so kube-apiserver and cluster-kube-apiserver-operator, not openshift-apiserver.

neisw:
Hi Oleg,  I have created a [gist](https://gist.github.com/neisw/19c6cf1f1243946ebd7f861f964ceb5e) with some information on what I have found.  Mostly it looks like the expectedGeneration passed in during the [Update](https://github.com/openshift/cluster-image-registry-operator/blob/30f80bdfda183b007f5a073d7f16d889e7f83bd7/pkg/resource/nodecadaemon.go#L85) call is always 0 so it never matches the existingCopy value and the apply happens over and over.  There are multiple updates right at the time of the failure, if we had the right generation value we would most likely not be making an update at the time we encounter that failure.

Still scratching my head on why the retry doesn't seem to kick in and we have the 10 minute delay.  It would still be possible to hit this error even with the correct expectedGeneration and it would be nice to recover quickly if we could.  But the pattern of multiple updates every 10 minutes or so seems to be from the UpdateFunc getting called 3 times on that interval && the issue above.

I have another run going to see if we can get any more info on why we don't seem to be updating the correct generation value during [SetDaemonSetGeneration](https://github.com/openshift/cluster-image-registry-operator/blob/30f80bdfda183b007f5a073d7f16d889e7f83bd7/pkg/resource/nodecadaemon.go#L93).

I hope this helps some.  If you have any thoughts let me know.

neisw:
@dmage with some help from @deads2k found the issue with the '0' expectedGeneration values was the desired dameonset was being used to set the generation value and not the actual one returned from the apply.  There is a little extra logging thrown in as I'm still curious why we didn't appear to be retrying (but in my testing that code path seemed to be working)  Would like to get your thoughts and see if the fix looks correct.

pkg/resource/nodecadaemon.go
```diff
@@ -70,27 +70,26 @@ func (ds *generatorNodeCADaemonSet) Create() (runtime.Object, error) {
...
+		resourcemerge.ExpectedDaemonSetGeneration(desiredDaemonSet, opStatus.Generations),
 	)
 	if err != nil {
 		return o, updated, err
 	}
 
 	if updated {
 		updateStatusFn := func(newStatus *operatorv1.OperatorStatus) error {
-			resourcemerge.SetDaemonSetGeneration(&newStatus.Generations, daemonSet)
+			resourcemerge.SetDaemonSetGeneration(&newStatus.Generations, actualDaemonSet)
```
deads2k, line 92: good fix

PR #798, opened 2022-08-25
Bug OCPBUGS-716: Add support for kube event recorder

neisw:
While investigating caches not synched error it was noted that the code was using a logging event recorder.  A request was made at that time to update the event recorder to increase observability and CI signal as we likely would have caught the frequent updating vi our CI testing.

### DIFF
--- a/pkg/operator/nodecadaemon.go
+++ b/pkg/operator/nodecadaemon.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/klog/v2"
 
 	operatorv1 "github.com/openshift/api/operator/v1"
+	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 
 	"github.com/openshift/cluster-image-registry-operator/pkg/defaults"
@@ -27,6 +28,7 @@ import (
 )
 
 type NodeCADaemonController struct {
+	eventRecorder   events.Recorder
 	appsClient      appsv1client.AppsV1Interface
 	operatorClient  v1helpers.OperatorClient
 	daemonSetLister appsv1listers.DaemonSetNamespaceLister
@@ -37,12 +39,14 @@ type NodeCADaemonController struct {
 }
 
 func NewNodeCADaemonController(
+	eventRecorder events.Recorder,
 	appsClient appsv1client.AppsV1Interface,
 	operatorClient v1helpers.OperatorClient,
 	daemonSetInformer appsv1informers.DaemonSetInformer,
 	serviceInformer corev1informers.ServiceInformer,
 ) *NodeCADaemonController {
 	c := &NodeCADaemonController{
+		eventRecorder:   eventRecorder,
 		appsClient:      appsClient,
 		operatorClient:  operatorClient,
 		daemonSetLister: daemonSetInformer.Lister().DaemonSets(defaults.ImageRegistryOperatorNamespace),
@@ -102,7 +106,7 @@ func (c *NodeCADaemonController) processNextWorkItem() bool {
 
 func (c *NodeCADaemonController) sync() error {
 	ctx := context.TODO()
-	gen := resource.NewGeneratorNodeCADaemonSet(c.daemonSetLister, c.serviceLister, c.appsClient, c.operatorClient)
+	gen := resource.NewGeneratorNodeCADaemonSet(c.eventRecorder, c.daemonSetLister, c.serviceLister, c.appsClient, c.operatorClient)
 
 	availableCondition := operatorv1.OperatorCondition{
 		Type:   "NodeCADaemonAvailable",
--- a/pkg/operator/starter.go
+++ b/pkg/operator/starter.go
@@ -6,6 +6,7 @@ import (
 	kubeinformers "k8s.io/client-go/informers"
 	kubeclient "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
+	"k8s.io/klog/v2"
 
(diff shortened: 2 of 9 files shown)


pkg/operator/nodecadaemon.go
```diff
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	appsv1informers "k8s.io/client-go/informers/apps/v1"
 	corev1informers "k8s.io/client-go/informers/core/v1"
+	kubeclient "k8s.io/client-go/kubernetes"
```
deads2k, line 15: the preference here is to use the smaller group client types.  See kubernetes/types/apps/v1 below.

pkg/operator/starter.go
```diff
@@ -23,6 +25,10 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/defaults"
 )
 
+const (
+	componentName = "image-registry"
```
deads2k, line 29: needs godoc.  Component used for what, seen by who?

pkg/operator/nodecadaemon.go
```diff
@@ -37,12 +39,14 @@ type NodeCADaemonController struct {
 }
 
 func NewNodeCADaemonController(
+	kubeClient *kubeclient.Clientset,
```
deads2k, line 42: can we pass an eventrecorder instead?  like https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/pkg/operator/starter.go#L114

pkg/operator/starter.go
```diff
@@ -126,9 +133,17 @@ func RunOperator(ctx context.Context, kubeconfig *restclient.Config) error {
 		configInformers.Config().V1().Images(),
 	)
 
+	var recorder events.Recorder
+	controllerRef, err := events.GetControllerReferenceForCurrentPod(context.TODO(), kubeClient, defaults.ImageRegistryOperatorNamespace, nil)
```
deads2k, line 137: check to be sure you have the pod name envvar set: Something like https://github.com/openshift/cluster-kube-apiserver-operator/blob/master/manifests/0000_20_kube-apiserver-operator_06_deployment.yaml#L70-L73

pkg/operator/starter.go
```diff
@@ -126,9 +133,17 @@ func RunOperator(ctx context.Context, kubeconfig *restclient.Config) error {
...
+	if err != nil {
+		klog.Warningf("Cannot create controller reference event recorder. Using logging event recorder: %v", err)
```
deads2k, line 139: This seems exit worthy.  Check to see what cluster-kube-apiserver-operator does.

pkg/operator/starter.go
```diff
@@ -112,6 +118,7 @@ func RunOperator(ctx context.Context, kubeconfig *restclient.Config) error {
 	)
 
 	nodeCADaemonController := NewNodeCADaemonController(
```
deads2k, line 120: if you pass in an eventRecorder you will be able to set up your unit tests with https://github.com/openshift/library-go/blob/master/pkg/operator/events/recorder_in_memory.go#L32-L34

pkg/resource/deployment.go
```diff
@@ -6,6 +6,13 @@ import (
 	"fmt"
 	"os"
 
+	corev1 "k8s.io/api/core/v1"
```
deads2k, line 9: nit tidy up imports

pkg/resource/deployment.go
```diff
@@ -41,9 +48,23 @@ type generatorDeployment struct {
 	cr              *imageregistryv1.Config
 }
 
-func newGeneratorDeployment(lister appslisters.DeploymentNamespaceLister, configMapLister corelisters.ConfigMapNamespaceLister, secretLister corelisters.SecretNamespaceLister, proxyLister configlisters.ProxyLister, coreClient coreset.CoreV1Interface, client appsset.AppsV1Interface, driver storage.Driver, cr *imageregistryv1.Config) *generatorDeployment {
+func newGeneratorDeployment(kubeconfig *rest.Config, lister appslisters.DeploymentNamespaceLister, configMapLister corelisters.ConfigMapNamespaceLister, secretLister corelisters.SecretNamespaceLister, proxyLister configlisters.ProxyLister, coreClient coreset.CoreV1Interface, client appsset.AppsV1Interface, driver storage.Driver, cr *imageregistryv1.Config) *generatorDeployment {
```
deads2k, line 51: we want to reuse the same eventrecorder here.

pkg/operator/nodecadaemon.go
neisw: removed with the change to pass in the eventRecorder

pkg/operator/nodecadaemon.go
neisw: yep

pkg/operator/starter.go
neisw: https://github.com/openshift/library-go/blob/4362aa519714a4b62b00ab8318197ba2bba51cb7/pkg/controller/controllercmd/builder.go#L230 seemed to allow for this without exiting, modeled it off of that.

deads2k:
infra
/retest-required

This change looks nice to me.  After this, I suggest wiring to other aspects well.

/assign @dmage

dmage:
/lgtm

to merge it, you need either a valid Jira bug, or a valid Jira story that is marked as no-doc, no-qe, and no-px (or get approvals from docs/qe/px).

dmage:
i.e. for a Jira story you need to get qe-approved, docs-approved, and px-approved labels

PR #806, opened 2022-10-18
WIP: Add capacity for qe testing only

shellyyang1989:


flavianmissi:
hey @shellyyang1989, looks like the `manifests/07-operator-ibm-cloud-managed.yaml` was manually changed, but this file is auto-generated - that's why the verify tests are failing. It's generated based on `manifests/07-operator.yaml` which you already changed.

To fix the verify test I think you need to run `make update` and push whatever gets regenerated.

shellyyang1989:
Hi @flavianmissi, thanks for spotting it! It's a PR for my test only which is done. Let me close it!

PR #818, opened 2022-11-25
hack/test-go.sh: generate coverage reports

flavianmissi:


dmage:
/lgtm

marking it as approved as it doesn't affect users
/label docs-approved
/label px-approved
/label qe-approved

PR #819, opened 2022-11-25
OCPBUGS-4090: swift: Retry connecting to OpenStack

pierreprinetti:
The function IsSwiftEnabled probes a connection to Swift to determine whether CIRO should fall back to using a persistent volume claim as a storage backend. This function runs on the OpenStack platform to provide an alternative solution for when Swift is not available.

Before this patch, any failure to communicate with Swift would make CIRO fall back to using a PVC. However, a failure to probing Swift could come from a temporary failure to reach the OpenStack API.

With this patch, when an issue with the connection to the OpenStack identity service is not allowing CIRO to probe Swift, the probing reports an error instead of directly falling back to Swift.

As a consequence, after this patch, temporary failures in early steps of the connection to Swift, do not make CIRO fall back to using persistent volume claims when object storage is available in OpenStack.

### DIFF
--- a/pkg/storage/swift/swift.go
+++ b/pkg/storage/swift/swift.go
@@ -3,6 +3,7 @@ package swift
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"errors"
 	"fmt"
 	"net/http"
 	"reflect"
@@ -19,7 +20,7 @@ import (
 	yamlv2 "gopkg.in/yaml.v2"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apimachineryerrors "k8s.io/apimachinery/pkg/api/errors"
 	k8sutilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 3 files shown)


pierreprinetti:
/hold
just to have at least another openstack review. Thanks

pierreprinetti:
Cool, thanks. I'll just wait for `e2e-openstack` to complete successfully before removing the hold.

mdbooth (changes requested):
This doesn't look right to me for a couple of reasons. Firstly implementing a backoff loop inside a reconciler generally isn't the right thing to do because it will (depending on the implementation, at least) prevent the reconciler from doing anything else. The idiomatic thing to do is to return an error and wait to be called again.

Secondly, we have enough context in [getSwitchClient](https://github.com/openshift/cluster-image-registry-operator/blob/d975b6524f7957e8a7332650b85e541d59a246b0/pkg/storage/swift/swift.go#L280-L286) to know if the error is that swift isn't installed in the cluster, in which case we should fallback immediately, or that swift just isn't working right now, in which case we should never fallback.

flavianmissi:
@pierreprinetti please address @mdbooth's feedback - I think it makes sense. I'll hold approval for now.
/approve cancel

pierreprinetti:
@flavianmissi sure, I will address Matt's comment. Note that this bug is important for us, but not urgent; I had to put it aside for a moment.

pierreprinetti:
> The idiomatic thing to do is to return an error and wait to be called again.

Thank you. The function now errors.

> Secondly, we have enough context in [getSwitchClient](https://github.com/openshift/cluster-image-registry-operator/blob/d975b6524f7957e8a7332650b85e541d59a246b0/pkg/storage/swift/swift.go#L280-L286) to know if the error is that swift isn't installed in the cloud, in which case we should fallback immediately, or that swift just isn't working right now, in which case we should never fallback.

The current logic also falls back to Cinder in case of insufficient permissions (for example, the user does not have the Swift role). Unless I am missing something, this means that we have to probe directly acting on Swift.

PTAL!

flavianmissi:
Leaving LGTM to OpenStack folks.
/approve

pkg/storage/swift/swift.go
```diff
@@ -241,24 +258,42 @@ func (d *driver) getSwiftClient() (*gophercloud.ServiceClient, error) {
...
-	client, err = openstack.NewContainerV1(provider, endpointOpts)
-	if _, ok := err.(*gophercloud.ErrEndpointNotFound); ok {
-		endpointOpts.Type = "object-store"
-		client, err = openstack.NewContainerV1(provider, endpointOpts)
-		if err != nil {
-			return nil, err
+	client, err := openstack.NewContainerV1(provider, endpointOpts)
+	if err != nil {
+		if _, ok := err.(*gophercloud.ErrEndpointNotFound); ok {
+			// In gophercloud the default endpoint type for
```
mandre, line 272: Thanks for adding the documentation to the code, to spare some git spelunking for the next person trying to understand what this code is about.

mandre (commented):
/lgtm

PR #820, opened 2022-11-25
Add config for golangci-lint and fix errors

dmage:


.golangci.yaml
```diff
@@ -0,0 +1,11 @@
...
+  enable:
+  - errcheck
+  - exhaustive
+  - gosimple
+  - govet
+  - ineffassign
+  - misspell
+  - staticcheck
+  - typecheck
+  - unused
```
dmage, line 11: @flavianmissi do you want to add something from https://golangci-lint.run/usage/linters/ ?

dmage: let's enable it too

dmage:
The code essentially hasn't changed, so marking this PR as approved.
/label qe-approved
/label docs-approved
/label px-approved

PR #824, opened 2022-12-11
Bug 2065166: Remove roles/iam.serviceAccountUser role

dmage:


dmage:
It seems it works without this role.
/assign @flavianmissi

PR #825, opened 2022-12-19
IR-308: Add support for External platform

dmage:
The registry shouldn't be installed by default on the External platform.


dmage:
/assign @stevsmit 
/assign @xiuwang 
/assign @sferich888 
for docs, qe, px approvals

PR #829, opened 2023-01-10
[CFE-678] Add UserTags while creating Azure Storage Account

chiragkyal:
Related to : https://github.com/openshift/enhancements/pull/1191
API Changes : https://github.com/openshift/api/pull/1280

Add UserTags while creating Azure Storage Account.


pkg/storage/azure/azure.go
```diff
@@ -516,13 +517,33 @@ func (d *driver) assureStorageAccount(cfg *Azure, infra *configv1.Infrastructure
...
+
+	tagset := map[string]*string{
+		fmt.Sprintf("kubernetes.io_cluster.%s", infra.Status.InfrastructureName): to.StringPtr("owned"),
+	}
+
+	// at this stage we are not keeping user tags in sync. as per enhancement proposal
+	// we only set user provided tags when we created the bucket.
+	hasAzureStatus := infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.Azure != nil && infra.Status.PlatformStatus.Azure.ResourceTags != nil
+	if hasAzureStatus {
+		klog.Infof("user has provided %d tags", len(infra.Status.PlatformStatus.Azure.ResourceTags))
```
bharath-b-rh, line 532:
```suggestion
		klog.V(5).Infof("user has provided %d tags", len(infra.Status.PlatformStatus.Azure.ResourceTags))
```

pkg/storage/azure/azure.go
```diff
@@ -516,13 +517,33 @@ func (d *driver) assureStorageAccount(cfg *Azure, infra *configv1.Infrastructure
 		return "", false, fmt.Errorf("create storage account failed, name not available")
 	}
 
+	// Tag the storage account with the openshiftClusterID
+	// along with any user defined tags from the cluster configuration
+	klog.Info("setting azure storage account tags")
```
bharath-b-rh, line 522:
```suggestion
	klog.V(2).Info("setting azure storage account tags")
```

pkg/storage/azure/azure.go
```diff
@@ -516,13 +517,33 @@ func (d *driver) assureStorageAccount(cfg *Azure, infra *configv1.Infrastructure
...
+	// we only set user provided tags when we created the bucket.
+	hasAzureStatus := infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.Azure != nil && infra.Status.PlatformStatus.Azure.ResourceTags != nil
+	if hasAzureStatus {
+		klog.Infof("user has provided %d tags", len(infra.Status.PlatformStatus.Azure.ResourceTags))
+		for _, tag := range infra.Status.PlatformStatus.Azure.ResourceTags {
+			klog.Infof("user has provided storage account tag: %s: %s", tag.Key, tag.Value)
+			tagset[tag.Key] = to.StringPtr(tag.Value)
+		}
+	}
+	klog.V(5).Infof("tagging storage accoun with tags: %+v", tagset)
```
bharath-b-rh, line 538:
```suggestion
	klog.V(5).Infof("tagging storage account with tags: %+v", tagset)
```

pkg/storage/azure/azure_test.go
```diff
@@ -329,6 +333,132 @@ func TestConfigEnvWithUserKey(t *testing.T) {
...
+
+			for _, resp := range sender.response {
+				if resp != nil && resp.Request != nil && resp.Request.Body != nil {
+
+					reqBody := make(map[string]interface{})
+					if err := json.NewDecoder(resp.Request.Body).Decode(&reqBody); err != nil {
+						t.Fatalf("error decoding request: %q", err)
+					}
+
+					// ignore reqest without tags
```
bharath-b-rh, line 435:
```suggestion
					// ignore request without tags
```

pkg/storage/azure/azure_test.go
```diff
@@ -329,6 +333,132 @@ func TestConfigEnvWithUserKey(t *testing.T) {
...
+					if _, ok := reqBody["tags"]; ok {
+						if tags, ok := reqBody["tags"].(map[string]interface{}); ok {
+
+							// convert into correct type
+							receivedTags := make(map[string]string)
+							for k, v := range tags {
+								receivedTags[k] = fmt.Sprintf("%+v", v)
```
bharath-b-rh, line 442:
```suggestion
								receivedTags[k] = v.(string)
```

pkg/storage/azure/azure_test.go
```diff
@@ -329,6 +333,132 @@ func TestConfigEnvWithUserKey(t *testing.T) {
...
+			for _, resp := range sender.response {
+				if resp != nil && resp.Request != nil && resp.Request.Body != nil {
+
+					reqBody := make(map[string]interface{})
+					if err := json.NewDecoder(resp.Request.Body).Decode(&reqBody); err != nil {
+						t.Fatalf("error decoding request: %q", err)
+					}
+
+					// ignore reqest without tags
+					if _, ok := reqBody["tags"]; ok {
```
bharath-b-rh, line 436: What if none of the response.Request.Body contain tags?

pkg/storage/azure/azure.go
chiragkyal: Done

pkg/storage/azure/azure.go
chiragkyal: Done

pkg/storage/azure/azure.go
```diff
@@ -516,13 +517,33 @@ func (d *driver) assureStorageAccount(cfg *Azure, infra *configv1.Infrastructure
...
+		fmt.Sprintf("kubernetes.io_cluster.%s", infra.Status.InfrastructureName): to.StringPtr("owned"),
+	}
+
+	// at this stage we are not keeping user tags in sync. as per enhancement proposal
+	// we only set user provided tags when we created the bucket.
+	hasAzureStatus := infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.Azure != nil && infra.Status.PlatformStatus.Azure.ResourceTags != nil
+	if hasAzureStatus {
+		klog.Infof("user has provided %d tags", len(infra.Status.PlatformStatus.Azure.ResourceTags))
+		for _, tag := range infra.Status.PlatformStatus.Azure.ResourceTags {
+			klog.Infof("user has provided storage account tag: %s: %s", tag.Key, tag.Value)
```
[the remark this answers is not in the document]
chiragkyal: Done

pkg/storage/azure/azure.go
chiragkyal: Thanks for noting the typo

pkg/storage/azure/azure_test.go
chiragkyal: Thanks!

pkg/storage/azure/azure_test.go
chiragkyal: Yes, we can write like this also,  but as we don't know the exact type of `v`, doing type assertion could lead to `panic` if `v` is not a string. That is rare, but to be in safe side I have written like this. 

if we choose to use type assertion, then I think it should be 

```go
if val, ok := v.(string); ok{
	receivedTags[k] = val
}
```
I'm fine with either way.

pkg/storage/azure/azure_test.go
chiragkyal: That's a good point. I have added a check for this. Thanks for noting.

pkg/storage/azure/azure_test.go
bharath-b-rh: Yeah, existing should be fine.

pkg/storage/azure/azure_test.go
```diff
@@ -329,6 +333,138 @@ func TestConfigEnvWithUserKey(t *testing.T) {
...
+
+							// compare the tags
+							if !reflect.DeepEqual(tt.expectedTags, receivedTags) {
+								t.Fatalf(
+									"expected tags %+v, received tags %+v",
+									tt.expectedTags, receivedTags,
+								)
+							}
+						} else {
+							t.Fatal("unable to type asset tags field")
```
flavianmissi, line 456: s/asset/assert

pkg/storage/azure/azure_test.go
```diff
@@ -329,6 +333,138 @@ func TestConfigEnvWithUserKey(t *testing.T) {
...
+							// convert into correct type
+							receivedTags := make(map[string]string)
+							for k, v := range tags {
+								receivedTags[k] = fmt.Sprintf("%+v", v)
+							}
+
+							// compare the tags
+							if !reflect.DeepEqual(tt.expectedTags, receivedTags) {
+								t.Fatalf(
+									"expected tags %+v, received tags %+v",
```
flavianmissi, line 451: There's a handy lib called `cmp` that will output diffs between two values. Since this isn't a simple string I think it's worth using it to output the diff here.
Example: https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/s3/s3_test.go#L158-L160

chiragkyal: oh, nice. I was not aware of this lib. Sure, this can be changed to 

```
t.Fatalf(
		"unexpected tags: %s",
		cmp.Diff(tt.expectedTags, receivedTags),
	)
```

pkg/storage/azure/azure_test.go
```diff
@@ -329,6 +333,138 @@ func TestConfigEnvWithUserKey(t *testing.T) {
...
+
+					reqBody := make(map[string]interface{})
+					if err := json.NewDecoder(resp.Request.Body).Decode(&reqBody); err != nil {
+						t.Fatalf("error decoding request: %q", err)
+					}
+
+					// ignore request without tags
+					if _, ok := reqBody["tags"]; ok {
+						foundTags = true
+						if tags, ok := reqBody["tags"].(map[string]interface{}); ok {
```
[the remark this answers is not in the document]
chiragkyal: yup, good point. The `else` part can be removed by failing the test early. 

However, I would choose to keep the other conditions (line 438 and 430), without `continue` for better readability may be. Thoughts ?

pkg/storage/azure/azure_test.go
chiragkyal: typo: changed to `assert`

pkg/storage/azure/azure_test.go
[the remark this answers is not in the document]
flavianmissi: The less indentation the better readability. But it's good enough as is now.

jinyunma:
pre-merge testing is passed, storage account for image registry has expected user defined tags, details can be found in https://issues.redhat.com/browse/CORS-2507
/label qe-approved

PR #783, opened 2022-05-19
Bump golangci-lint

dmage:
The current version doesn't work with go1.18. This fix is needed for #782.

PR #832, opened 2023-01-18
[release-4.10] OCPBUGS-5974: swift: Retry connecting to OpenStack

pierreprinetti:
This is a cherry-pick of https://github.com/openshift/cluster-image-registry-operator/pull/830
This PR also cherry-picks https://github.com/openshift/cluster-image-registry-operator/pull/783 because older versions of `gosimple` trigger errors on my code, requiring it to be less explicit.

---

The function IsSwiftEnabled probes a connection to Swift to determine whether CIRO should fall back to using a persistent volume claim as a storage backend. This function runs on the OpenStack platform to provide an alternative solution for when Swift is not available.

Before this patch, any failure to communicate with Swift would make CIRO fall back to using a PVC. However, a failure to probing Swift could come from a temporary failure to reach the OpenStack API.

With this patch, when an issue with the connection to the OpenStack identity service is not allowing CIRO to probe Swift, the probing reports an error instead of directly falling back to Swift.

As a consequence, after this patch, temporary failures in early steps of the connection to Swift do not make CIRO fall back to using persistent volume claims when object storage might be available in OpenStack.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ PROG  := cluster-image-registry-operator
 
 GOLANGCI_LINT = _output/tools/golangci-lint
 GOLANGCI_LINT_CACHE = $(PWD)/_output/golangci-lint-cache
-GOLANGCI_LINT_VERSION = v1.24.0
+GOLANGCI_LINT_VERSION = v1.46.2
 
 GO_REQUIRED_MIN_VERSION = 1.16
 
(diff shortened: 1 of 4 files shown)


pierreprinetti:
e2e-openstack depends on https://github.com/openshift/release/pull/35530

mandre (commented):
/lgtm

PR #842, opened 2023-02-16
Power VS: Use IBM COS as storage backend

mjturek:

PR #843, opened 2023-02-17
MULTIARCH-3212: Use IBM COS as storage backend for PowerVS

dharaneeshvrd:
Fixes https://issues.redhat.com/browse/MULTIARCH-3212


flavianmissi:
could you please sync with the author of https://github.com/openshift/cluster-image-registry-operator/pull/842 ?

dharaneeshvrd:
> could you please sync with the author of #842 ?

Yes, we did, decided to keep this PR. Will close the other one.

manifests/01-registry-credentials-request-powervs.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: openshift-image-registry-ibmcos
```
flavianmissi, line 6: `manifests/01-registry-credentials-request-ibmcos.yaml` already uses that name.

pkg/storage/ibmcos/powervs_regions.go
```diff
@@ -0,0 +1,61 @@
...
+
+// powerVSRegion describes respective IBM Cloud region and PowerVS zones associated with a region in PowerVS.
+// which helps to decide COS location from PowerVS region with the intention to keep COS in same region as PowerVS.
+type powerVSRegion struct {
+	description string
+	cosRegion   string
+}
+
+// powerVSRegions provides a mapping between Power VS and IBM Cloud regions.
+var powerVSRegions = map[string]powerVSRegion{
```
flavianmissi, line 15: Is there any way this map can be built dynamically? What happens when IBM adds a new region?

pkg/storage/ibmcos/powervs_regions.go
```diff
@@ -0,0 +1,61 @@
+package ibmcos
+
+import (
+	"fmt"
+)
+
+// powerVSRegion describes respective IBM Cloud region and PowerVS zones associated with a region in PowerVS.
+// which helps to decide COS location from PowerVS region with the intention to keep COS in same region as PowerVS.
```
flavianmissi, line 8: The mix of terms here is confusing. You're using:
* IBM cloud region
* PowerVS zone
* PowerVS region
* COS location

Could you rephrase this comment so that it's easier to understand for someone who isn't familiar with ibm cloud?

flavianmissi (commented):
Looking good overall. Left a few comments.

pkg/storage/ibmcos/powervs_regions.go
clnperez: no, unfortunately. we explored that when we created it for IPI. we have to keep up with this ourselves. i submitted a feature request to have some of this made available via an API but it's not been taken up.

clnperez (commented):
LGTM aside from the name de-dup change requested

manifests/01-registry-credentials-request-powervs.yaml
dharaneeshvrd: renamed.

pkg/storage/ibmcos/powervs_regions.go
dharaneeshvrd: Tried rephrase the comment to make it clear on why this is required, ptal!

pkg/storage/ibmcos/powervs_regions.go
flavianmissi: Thanks for expanding on this. 
With this code, will cluster install fail on a region that is not listed in this map?
We have this situation on aws (whenever they add a new region we have to upgrade their sdk or else install fails on that region), I'm trying to avoid having the same thing happen for IBM COS. Is there any fallback we could choose? Or would it just not work (in which case it is better to fail).

pkg/storage/ibmcos/powervs_regions.go
flavianmissi: Rephrased comment is much more clear - thank you!

pkg/storage/ibmcos/powervs_regions.go
dharaneeshvrd: > With this code, will cluster install fail on a region that is not listed in this map?
Yes, we would want cluster installation to get fail if an unknown region is passed.

> whenever they add a new region we have to upgrade their sdk or else install fails on that region
That's true, we are planning to keep this in common pkg like library-go and use it here in future(planning to do it after 4.13). So no need to update in this repo whenever new region is added.

flavianmissi: okay, so no fallback then.
it would indeed be a lot better to import this from library-go or some other shared package 👍🏼

flavianmissi:
@clnperez has lgtm.
/approve
/retest

dharaneeshvrd:
@juliemathew

Could you please add `/label qe-approved` here?

dharaneeshvrd:
@alishaIBM 
Could you please add `/label docs-approved` `/label px-approved` labels?

PR #850, opened 2023-03-16
OCPBUGS-9962: Swift: Support openstack project domain

mandre:
CIRO previously had no logic to handle openstack project domain if it was set in the `clouds.yaml` file.
This commit removes the custom building of gophercloud AuthOptions struct and instead rely on the `AuthOptions()` function from gophercloud-utils that accounts for project domain.

Also pass the project domain name and id to the image registry via the `REGISTRY_STORAGE_SWIFT_TENANTDOMAIN` and
`REGISTRY_STORAGE_SWIFT_TENANTDOMAINID` environment variables.

### DIFF
--- a/pkg/storage/swift/swift.go
+++ b/pkg/storage/swift/swift.go
@@ -39,6 +39,8 @@ type Swift struct {
 	Password                    string
 	Tenant                      string
 	TenantID                    string
+	TenantDomain                string
+	TenantDomainID              string
 	Domain                      string
 	DomainID                    string
 	RegionName                  string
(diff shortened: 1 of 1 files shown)


mandre:
/hold
Still testing in a real environment.

mandre:
Note to reviewers:
- Unit tests are failing in my local branch, but it doesn't seem related to my patch as I get the same failure from master branch. It might have to do with how I run the tests. The unit job failed for a different reason.
- I wasn't sure if I needed to add `TenantDomain` and `TenantDomain` to the [`ImageRegistryConfigStorageSwift` type in openshift/api](https://github.com/openshift/api/blob/5e04e73/imageregistry/v1/types.go#L244-L273). Is it necessary? It would make things harder to backport.

mandre:
Right, so following up on [my previous comment](https://github.com/openshift/cluster-image-registry-operator/pull/850#issuecomment-1472164544) I did a few tests and effectively, the feature to set user-defined configuration via the `image-registry-private-configuration-user` secret is unusable.

With an empty secret, it complains about missing values:
```
$ oc create secret -n openshift-image-registry generic image-registry-private-configuration-user
secret/image-registry-private-configuration-user created
$ oc get co image-registry
NAME             VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
image-registry   4.12.5    True        True          False      11h     Progressing: Unable to apply resources: unable to sync storage configuration: secret "openshift-image-registry/image-registry-private-configuration-user" does not contain required key "REGISTRY_STORAGE_SWIFT_USERNAME"...
```

With a secret containing mandatory keys, it doesn't know how to reach the endpoints:
```
$ oc create secret -n openshift-image-registry generic image-registry-private-configuration-user \
    --from-literal=REGISTRY_STORAGE_SWIFT_USERNAME=foo \
    --from-literal=REGISTRY_STORAGE_SWIFT_PASSWORD=foo \
    --from-literal=REGISTRY_STORAGE_SWIFT_APPLICATIONCREDENTIALID=foo \
    --from-literal=REGISTRY_STORAGE_SWIFT_APPLICATIONCREDENTIALNAME=foo \
    --from-literal=REGISTRY_STORAGE_SWIFT_APPLICATIONCREDENTIALSECRET=foo                                                                                                                                                                                                                                                                                    
secret/image-registry-private-configuration-user created
$ oc get co image-registry
NAME             VERSION   AVAILABLE   PROGRESSING   DEGRADED   SINCE   MESSAGE
image-registry   4.12.5    True        True          False      11h     Progressing: Unable to apply resources: unable to sync storage configuration: failed to authenticate against OpenStack: Get "/": unsupported protocol scheme ""...
```

flavianmissi:
Hi @mandre, could this be a documentation bug? i.e https://issues.redhat.com/browse/OCPBUGS-2889

mandre:
> ~Hi @mandre, could this be a documentation bug? i.e https://issues.redhat.com/browse/OCPBUGS-2889~ nevermind, this seems unrelated!

It's kind of related to what I'm reporting here though, the REGISTRY_STORAGE_SWIFT_APPLICATIONCREDENTIAL* keys should be optional. Same for REGISTRY_STORAGE_SWIFT_PASSWORD if you use application credential. And it's missing other required variables to set the auth URL for example.

flavianmissi:
> This means that when the image-registry-private-configuration-user secret exists, the AuthURL and other required auth parameters are always empty.
> 
> I'm tempted to remove this code to prevent users from accidentally breaking their cluster.

I think we'd need to take the long way here, regardless of which path we chose. If you look for [`image-registry-private-configuration-user` in our docs](https://docs.openshift.com/container-platform/4.12/registry/configuring_registry_storage/configuring-registry-storage-openstack-user-infrastructure.html#registry-operator-config-resources-secret-openstack_configuring-registry-storage-openstack-user-infrastructure), you'll see that we do document how to use that (even if it doesn't work). So whether we chose to deprecate this feature or to fix it, we'll need to do it one step at a time.

Additionally, the `image-registry-private-configuration-user` feature is available on other clouds too (see docs for reference) - if we decide to deprecate it, it should affect more than just Swift.

I guess this means doing that api change in openshift/api and dealing with however that would affect backporting :/

wewang58:
@flavianmissi  could you help to rebase the pr, so i can launch a cluster with it, thanks

flavianmissi:
Hi @wewang58, we have some pending changes on this PR so I think you can put it lower on your list of priorities for now.
[OCPBUGS-2889](https://issues.redhat.com/browse/OCPBUGS-2889) is related - I think we need to focus on it before we can move forward with the changes on this PR.

flavianmissi:
Hi @mandre, got an update about the  `Get "/": unsupported protocol scheme ""` error you were seeing.
The docs aren't very clear, but you need to configure both the `image-registry-private-configuration-user` secret as well as the operator config - the [docs](https://docs.openshift.com/container-platform/4.13/registry/configuring_registry_storage/configuring-registry-storage-openstack-user-infrastructure.html#registry-configuring-storage-openstack-user-infra_configuring-registry-storage-openstack-user-infrastructure) state:
```
Fill in the storage configuration in configs.imageregistry.operator.openshift.io/cluster:
oc edit configs.imageregistry.operator.openshift.io/cluster
```
but they don't specify what needs to be configured. The [next section in the same page](https://docs.openshift.com/container-platform/4.13/registry/configuring_registry_storage/configuring-registry-storage-openstack-user-infrastructure.html#registry-operator-configuration-resource-overview-openstack-swift_configuring-registry-storage-openstack-user-infrastructure) lists what is supported for swift. Note the `authURL` is listed, which is what is missing from your config. I think you might also need to set a couple of other keys, if I remember correctly you also need `domain` (or `domainID`) and `regionName`.

PR #854, opened 2023-04-26
OCPBUGS-12132: Updating ose-cluster-image-registry-operator images to be consistent with ART

flavianmissi:
Also bump golangci-lint, and fix a deprecation error.

* golangci-lint bug that required upgrade: https://github.com/golangci/golangci-lint/issues/3565
* deprecation error and the choice to simply removed the call: https://pkg.go.dev/math/rand#Seed

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,4 +1,4 @@
 build_root_image:
   name: release
   namespace: openshift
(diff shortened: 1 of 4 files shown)


flavianmissi:
/assign @dmage 

let's hold this for a bit.
/hold

PR #859, opened 2023-06-21
wip: attempt to guess at the new azure permissions credentials request api

flavianmissi:


abutcher:
The deprovision didn't remove the original installation resource group since the install didn't start due to capacity issues. `ccoctl azure create-all` can't proceed because the resource group exists in wrong region.

flavianmissi:
replaced by https://github.com/openshift/cluster-image-registry-operator/pull/890
/close

PR #863, opened 2023-06-29
IR-390: Expose registry CAs as one to MCO

flavianmissi:


flavianmissi:
/hold 

This depends on the MCO work in https://issues.redhat.com/browse/MCO-499, and _**cannot**_ land before it, or things _**will**_ break

pkg/resource/imageregistryca.go
```diff
@@ -0,0 +1,187 @@
...
+		return cm, err
+	}
+	if driver != nil {
+		storageCABundle, _, err := driver.CABundle()
+		if err != nil {
+			return cm, err
+		}
+		if storageCABundle != "" {
+			klog.V(4).Infof("using storage ca bundle (%d bytes)", len(storageCABundle))
+			cm.Data["storage-ca-bundle.pem"] = storageCABundle
```
dmage, line 146: afaik we need it only for image-registry so that it always trusts the storage certs; clients don't need `storage-ca-bundle.pem`

flavianmissi: makes sense - will remove.

PR #867, opened 2023-07-03
IR-373: remove node-ca daemon

flavianmissi:
removes the node-ca daemon from fresh installs, as well as existing (upgraded) clusters.


pkg/operator/nodecadaemon.go
```diff
@@ -122,65 +121,38 @@ func (c *NodeCADaemonController) sync() error {
...
-			availableCondition.Status = operatorv1.ConditionFalse
-			availableCondition.Reason = "NoAvailableReplicas"
-			availableCondition.Message = "The daemon set node-ca does not have available replicas"
+		gracePeriod := int64(0)
+		propagationPolicy := metav1.DeletePropagationForeground
+		opts := metav1.DeleteOptions{
+			GracePeriodSeconds: &gracePeriod,
+			PropagationPolicy:  &propagationPolicy,
 		}
+		err := gen.Delete(opts)
```
flavianmissi, line 146: What happens if you send the delete request twice?
You can see this:
1. add your own finalizer in the node-ca daemon object
2. run the reconcile loop
3. see what happens when the delete request gets called multiple times

flavianmissi: After adding a custom finalizer there were no errors in the operator logs. 
After a minute I removed the custom finalizer, and the logs first showed a transition to `NodeCADaemonAvailable: Unable to delete daemon set: daemonsets.apps \"node-ca\" not found`, then immediately transitioned to `NodeCADaemonProgressing: The daemon set node-ca is removed`.
The clusteroperator looks good too:
```
Status:
  Conditions:
    Last Transition Time:  2023-08-11T09:49:22Z
    Message:               Available: The registry is ready
NodeCADaemonAvailable: The daemon set node-ca is removed
ImagePrunerAvailable: Pruner CronJob has been created
    Reason:                Ready
    Status:                True
    Type:                  Available
    Last Transition Time:  2023-08-11T09:49:22Z
    Message:               Progressing: The registry is ready
NodeCADaemonProgressing: The daemon set node-ca is removed
    Reason:                Ready
    Status:                False
    Type:                  Progressing
    Last Transition Time:  2023-08-08T14:50:18Z
    Reason:                AsExpected
    Status:                False
    Type:                  Degraded
```

I wonder if we should change the `NodeCADaemonAvailable` to `NodeCADaemonRemoved` though, but I'm not sure 🤔

dmage: yep, looks like a race condition - Get returned the object which was deleted before the Delete request

flavianmissi: I have fixed the race condition. Marking this comment as resolved.

flavianmissi:
/unhold
this is ready for review

xiuwang:
/label qe-approved

Do upgrade test and regression test for #867  and https://github.com/openshift/machine-config-operator/pull/3854 . No new issue found

PR #873, opened 2023-07-05
CFE-846: Add user defined tags to the GCP buckets created

bharath-b-rh:
PR has the changes for https://github.com/openshift/enhancements/pull/1217 proposed to support GCP tags in OCP, which requires cluster-image-registry-operator to add gcp userTags available in the status sub resource of infrastructure CR, to the gcp buckets created by the operator.


go.mod
```diff
@@ -140,34 +151,37 @@ require (
...
+	k8s.io/kube-openapi v0.0.0-20230501164219-8b0f38b5fd1f // indirect
+	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.1.2 // indirect
 	sigs.k8s.io/json v0.0.0-20221116044647-bc3834ca7abd // indirect
 	sigs.k8s.io/kube-storage-version-migrator v0.0.4 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace google.golang.org/grpc => google.golang.org/grpc v1.40.0
+replace google.golang.org/grpc => google.golang.org/grpc v1.55.0
```
bharath-b-rh, line 185: Updated the package version to the same as found in required block.

pkg/storage/gcs/gcs.go
```diff
@@ -304,6 +304,10 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 				cr.Spec.Storage.GCS = d.Config.DeepCopy()
 			}
 		}
+
+		if err := addUserTagsToStorageBucket(d.Context, d.Listers, d.Config.Bucket, d.Config.Region); err != nil {
+			return err
```
bharath-b-rh, line 309: Create event instead of error.

pkg/resource/generator.go
```diff
@@ -147,14 +147,19 @@ func (g *Generator) syncStorage(cr *imageregistryv1.Config) error {
...
+		}
 	}
 
 	if runCreate {
-		reconf := g.storageReconfigured(cr, g.kubeconfig, g.listers)
 		if err := driver.CreateStorage(cr); err != nil {
 			return err
 		}
-		if reconf {
+		if g.storageReconfigured(cr, g.kubeconfig, g.listers) {
```
flavianmissi, line 162: is there a specific reason for this move?

bharath-b-rh: While making the code changes, IDE suggested the change :)

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +79,220 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+
+	return filteredTags
+}
+
+// getHTTPCodeInGAPIError return the HTTP status code in the Google API
+// defined error.
+func getHTTPCodeInGAPIError(e error) int64 {
+	return reflect.Indirect(
+		reflect.ValueOf(e).Elem().FieldByName(googleAPIHTTPErrFieldName),
+	).FieldByName(googleAPIHTTPCodeFieldName).Int()
```
flavianmissi, line 176: is reflection really the only way to do this? I can live with it if it is but if there is another way we should use it.

flavianmissi (commented):
looking good overall, left some comments

pkg/resource/generator.go
bharath-b-rh: suggestion incorporated.

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +79,220 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+	for _, k := range tagList {
+		dupTags[k] = false
+	}
+
+	listBindingsReq := &rscmgrpb.ListEffectiveTagsRequest{
+		Parent: parent,
+	}
+	bindings := client.ListEffectiveTags(ctx, listBindingsReq)
+	// a resource can have a maximum of 50 tags attached to it. Will iterate
+	// for 50 times in the worst case scenario of none of the break conditions are met.
```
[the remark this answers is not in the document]
bharath-b-rh: suggestion incorporated.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: In `gax-go/v2 v2.8.0` those fields weren't exported and had to make use of reflection. But from `v2.9.0`, `HTTPCode()` method is available for the same. I have updated the code with the changes

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +79,220 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+	for i := 0; i < 50; i++ {
```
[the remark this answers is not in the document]
bharath-b-rh: I am little hesitant to use unconditional loops in case the break conditions aren't met and it goes into infinite loop.
My idea here was to reduce the number of tag write API calls by looking for existing tags and skipping them, since it has per minute quota restriction. 

If the max tags number is increased, we would be able skip those in the first 50(which will not meet the purpose of the code here) but the logic [here](https://github.com/openshift/cluster-image-registry-operator/pull/873/files/33417f06aa5f83823ec346cf4b1aaaa1ed060cd5..6f8f74e620813410862c9fcc292fde8d9aede33c#diff-6e6f5be6182d62cd1ad9867497387a3c45de233050881e4479396cea0652d4a5R279) will handle existing tags error which weren't skipped. And this would result in additional write calls for already existing tags, but still work as expected, I think. 

Please let me know your thoughts.

xiuwang:
Hi @wewang58, could you help test this code?
Please collaborate with @jianli-wei on how we test it if you need help.
Thank you both!

pkg/resource/generator.go
flavianmissi: did you forget to push @bharath-b-rh ? code looks the same

pkg/storage/gcs/gcp_labels_tags.go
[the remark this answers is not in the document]
flavianmissi: Okay so you're saying that even if the max number of tags per resource increase and this code can no longer get all of the tags because of that - it would be okay because the filtering done in this function is an optimization to reduce the number of api calls when trying to tag resources. Got it.

In that case, let's just add this explanation as a comment to the code. WDYT?

pkg/resource/generator.go
bharath-b-rh: Sorry, yes I had missed committing changes and commented as done. It's updated now.

pkg/storage/gcs/gcp_labels_tags.go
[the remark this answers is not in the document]
bharath-b-rh: suggestion incorporated, updated the comment.

pkg/resource/generator.go
```diff
@@ -147,6 +147,12 @@ func (g *Generator) syncStorage(cr *imageregistryv1.Config) error {
 		if !exists {
 			runCreate = true
 		}
+
+		// GCP: If during previous reconcile error was set, failing to add required
+		// tags, setting runCreate to true  to retry adding the tags.
+		if cr.Spec.Storage.GCS != nil {
+			runCreate = true
+		}
```
flavianmissi, line 155: This feels hacky. Could you trigger recreation using the GCS driver's `StorageChanged` method? It's still kinda hacky but at least this generator wouldn't have to know about it.

pkg/storage/gcs/gcs.go
```diff
@@ -212,6 +212,13 @@ func (d *driver) StorageChanged(cr *imageregistryv1.Config) bool {
 }
 
 func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
+	// Previous status condition was set to failed because, of tag operation failure,
+	// means bucket creation was successful and hence will only try adding tags.
+	if cond := util.FetchCondition(cr, defaults.StorageTagged); cond.Status == operatorapi.ConditionFalse &&
+		cond.Reason == gcpTagsFailedStatusReason {
```
flavianmissi, line 218: let's change this condition around so that we check for anything other than success, instead of checking for the specific error? i.e
```
if cond := util.FetchCondition(cr, defaults.StorageTagged); cond.Status != operatorapi.ConditionTrue {
```

flavianmissi (commented):
Left some comments. Looks like I missed some stuff on the first round - sorry about that!

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +74,216 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+	return list
+}
+
+// getInfraResourceTagsList returns the user-defined tags present in the
+// status sub-resource of Infrastructure.
+func getInfraResourceTagsList(platformStatus *configv1.PlatformStatus) []configv1.GCPResourceTag {
+	if platformStatus != nil && platformStatus.GCP != nil && platformStatus.GCP.ResourceTags != nil {
+		return platformStatus.GCP.ResourceTags
+	}
+	klog.V(1).Infof("getInfraResourceTagsList: user-defined tag list is empty")
```
dmage, line 112: is not provided?

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +74,216 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+	// increased in future, it should not create an issue, since this is an optimization
+	// attempt to reduce the number the tag write calls by skipping already existing tags,
+	// since it has a quota restriction.
+	for i := 0; i < gcpMaxTagsPerResource; i++ {
+		binding, err := bindings.Next()
+		if errors.Is(err, iterator.Done) {
+			break
+		}
+		if err != nil || binding == nil {
+			klog.V(1).Infof("failed to list effective tags on the %s bucket: %v: %v", parent, binding, err)
```
dmage, line 150: Why don't return this error to the caller so that it can retry the operation?

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +74,216 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+		}),
+	}
+}
+
+// getTagBindingsClient returns the client to be used for creating tag bindings to
+// the resources.
+func getTagBindingsClient(ctx context.Context, listers *regopclient.StorageListers, location string) (*rscmgr.TagBindingsClient, error) {
+	cfg, err := GetConfig(listers)
+	if err != nil {
+		klog.Errorf("getTagBindingsClient: failed to read gcp config: %w", err)
```
dmage, line 190: Errors shouldn't be logged and returned at the same time. If the error is returned, it should be handled by the caller, often the caller will log it with additional context. If this log message is important for debugging, use `V(4).Infof`.

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +74,216 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+// to the passed GCP bucket resource. It's wrapper around addUserTagsToStorageBucket()
+// additionally updates status condition.
+func addTagsToStorageBucket(ctx context.Context, cr *imageregistryv1.Config, listers *regopclient.StorageListers, bucketName, region string) error {
+	if err := addUserTagsToStorageBucket(ctx, listers, bucketName, region); err != nil {
+		util.UpdateCondition(cr, defaults.StorageTagged, operatorapi.ConditionFalse,
+			gcpTagsFailedStatusReason, err.Error())
+		return err
+	}
+	util.UpdateCondition(cr, defaults.StorageTagged, operatorapi.ConditionTrue,
+		"Tagging Bucket Successful",
```
dmage, line 213: Either both calls should use named constants, or literal strings.

dmage (commented):
Will there be tests for this feature?

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: yes, if user hasn't configured tags.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: We are trying to save the API calls by finding the already existing tags, I feel retry would not be required, because the main objective would be achieved even without this. Please let me what do you think?

pkg/resource/generator.go
bharath-b-rh: suggestion incorporated, moved the check to `StorageChanged` method.

pkg/storage/gcs/gcs.go
bharath-b-rh: suggestion incorporated. Condition updated and moved to end of the function.

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -1,18 +1,57 @@
...
 )
 
 const (
 	// ocpDefaultLabelFmt is the format string for the default label
 	// added to the OpenShift created GCP resources.
 	ocpDefaultLabelFmt = "kubernetes-io-cluster-%s"
+
+	// gcpTagsFailedStatusReason is the operator condition status reason
+	// for failed tag operations.
+	gcpTagsFailedStatusReason = "Error Tagging Bucket"
```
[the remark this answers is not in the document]
bharath-b-rh: suggestion incorporated.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: suggestion incorporated, log removed and err wrapped with additional details.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: suggestion incorporated, success reason defined as a constant.

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +74,216 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+
+	infra, err := util.GetInfrastructure(listers.Infrastructures)
+	if err != nil {
+		klog.Errorf("addUserTagsToStorageBucket: failed to read infrastructure/cluster resource: %w", err)
+		return err
+	}
+
+	client, err := getTagBindingsClient(ctx, listers, region)
+	if err != nil || client == nil {
+		return fmt.Errorf("failed to create tag binding client for adding tags to %s bucket: %v",
```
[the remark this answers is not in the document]
bharath-b-rh: Yeah, while implementing had observed underneath APIs returned error doesn't implement `Error()` method and wrap formatting was failing. So just at this place used `%v`.

bharath-b-rh:
> Will there be tests for this feature?

Yeah, QE is working on the pre-merge tests and have asked for GCP regression tests too.

pkg/storage/gcs/gcp_labels_tags.go
flavianmissi: Perhaps this should be logged with `klog.V(4).Infof` instead then, since it seems to be more of a debug log?

bharath-b-rh: Yeah, but since we are not failing the operation and if it's sporadic/isolated failure, this log would be helpful I think. Could we use `V(1)`, since its logged only during error scenario?

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: suggestion incorporated. Log string updated to `getInfraResourceTagsList: user-defined tag list is not provided`

pkg/storage/gcs/gcp_labels_tags.go
flavianmissi: We probably could. I'm concerned about this happening and customers seeing it in their logs and mistaking it for something more serious than it is, that's why I suggested we change the level.
Do you think we absolutely need this on `V(1)`? If so, we probably want to rephrase the message so that there would be no confusion.

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -35,3 +77,213 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
...
+			break
+		}
+		if err != nil || binding == nil {
+			klog.V(1).Infof("failed to list effective tags on the %s bucket: %v: %v", parent, binding, err)
+			break
+		}
+		tag := binding.GetNamespacedTagValue()
+		if _, exist := dupTags[tag]; exist {
+			dupTags[tag] = true
+			klog.V(1).Infof("filterTagList: skipping tag %s already exists on the %s bucket", tag, parent)
```
flavianmissi, line 159: Same here, this looks like a debug log to me, so consider using `V(4)`

pkg/storage/gcs/gcp_labels_tags.go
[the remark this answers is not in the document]
flavianmissi: That's weird but okay. Marking this comment as resolved in that case.

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -1,25 +1,67 @@
...
+
+	"k8s.io/klog/v2"
 )
 
 const (
 	// ocpDefaultLabelFmt is the format string for the default label
 	// added to the OpenShift created GCP resources.
 	ocpDefaultLabelFmt = "kubernetes-io-cluster-%s"
+
+	// gcpTagsFailedStatusReason is the operator condition status reason
```
flavianmissi, line 35: s/gcpTagsFailedStatusReason/gcpTagsSuccessStatusReason

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -1,25 +1,67 @@
...
+	// for successful tag operations.
+	gcpTagsSuccessStatusReason = "SuccessTaggingBucket"
+
+	// gcpTagsFailedStatusReason is the operator condition status reason
+	// for failed tag operations.
+	gcpTagsFailedStatusReason = "ErrorTaggingBucket"
+
+	// gcpTagsFailedStatusReason is the maximum number of tags that can
```
flavianmissi, line 43: s/gcpTagsFailedStatusReason/gcpMaxTagsPerResource

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: I think `"filterTagList: listing effective tags on the %s bucket ended: %v: %v"` would be helpful for us to identifying if the scenario happened. Please let me WDYT.

wewang58:
HI, If i launch  a cluster with cluster-bot, which installer pr or api pr, also needed?

pkg/storage/gcs/gcp_labels_tags.go
flavianmissi: Alright I won't keep pushing, let's leave it as is

flavianmissi:
@bharath-b-rh are there other PRs needed to test this?
@wewang58 do you still need an answer?

flavianmissi:
test failures seem flake related
/retest

bharath-b-rh:
@flavianmissi @wewang58 : Yeah we need below installer PRs to test the feature.
https://github.com/openshift/installer/pull/7126
https://github.com/openshift/installer/pull/7201
https://github.com/openshift/installer/pull/7138
https://github.com/openshift/installer/pull/7279

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: suggestion incorporated.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: suggestion incorporated.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: suggestion incorporated.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: suggestion incorporated.

flavianmissi:
@bharath-b-rh do we need to merge the installer PRs and this one together in that case? sounds like this one won't work standalone.

bharath-b-rh:
> @bharath-b-rh do we need to merge the installer PRs and this one together in that case? sounds like this one won't work standalone.

Yeah, I am working with installer team to get the PRs merged. If we merge this without installer code, user cannot define tags and the registry operator won't add any tags. But the operator as such can be deployed.

flavianmissi:
e2e failure seems unrelated
/retest
/lgtm

PR #912, opened 2023-09-06
OCPBUGS-18469: increase storage account key cache expiration

flavianmissi:
make it 20 minutes instead of 5.

Azure has an api limit for storage account list (which apparently listing keys uses behind the scenes) of 100 calls / 5 minutes. if a project has a lot of clusters, this limit might be hit.


flavianmissi:
FTR I don't think we need this in 4.14.0 (we're still only merging critical fixes to master) so I'll wait with setting it until branching.

flavianmissi:
e2e failures seem unrelated
/retest

PR #924, opened 2023-09-25
OCPBUGS-2889: accept user/pass OR application credentials on Swift UPI secret

flavianmissi:
Previously the validation required both user/pass AND application credentials to be provided.
This allows users to use one OR the other.


flavianmissi:
failures seem unrelated
/retest

PR #959, opened 2023-11-13
OCPVE-790: annotate credentials request manifests

qJkee:
Add capability annotations to cloud credentials requests


flavianmissi:
this is not user-facing.
/label docs-approved
/label px-approved

PR #962, opened 2023-11-20
WIP IR-418: use installer bootstrapped storage config for Azure

flavianmissi:


flavianmissi:
I'm going to close this for now to not cause confusion - I'm not sure if we will pursue this. For now we'll ask customers to create the registry config manifest manually.
/close

PR #976, opened 2023-12-11
OCPBUGS-23953: avoid creation of multiple private endpoints on Azure

flavianmissi:
when a reconciliation fails for any reason, the operator will loose track of resources that were created during said reconciliation. this happens because the operator was designed to only write to its configuration spec after a successful reconciliation. to work around this, we tag the private endpoint using a well known tag, then check for a private endpoint with this tag before attempting to create one.


flavianmissi:
FTR: I've confirmed with installer folks that it's safe to use the infrastructure name for this purpose.

flavianmissi:
I'm going to ask QE to pre-merge this one to be on the safe side.
/hold

flavianmissi:
I'm not really confident about these changes... I'll close this PR until I have time to revisit it.

PR #981, opened 2023-12-18
IR-440: fix upstream Dockerfile

flavianmissi:


flavianmissi:
this is not user facing.
/label docs-approved
/label px-approved

PR #993, opened 2024-01-26
OCPBUGS-28225: pkg/storage/s3: enable bucket key on encryption settings

flavianmissi:
using this setting enables a key to be used for the entire bucket instead of AWS creating one key per object.
for more info see
https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-key.html.

---

### setting up your environment

* create a cluster (don't use cluster-bot unless you can access the AWS console for it)
* create a KMS key (use the defaults)
* copy the key id (ARN)
* add the registry IAM users to the list of users in your KMS key (if you skip this step you will get 500 Internal Server Error from the registry when you try to push images!)
* edit the operator config and set .spec.storage.s3.keyID to your key id,
  mine was "arn:aws:kms:ap-northeast-1:942491940283:key/da77a03c-692e-497d-bef6-d5da91e50f58", yours will be different.

### verify it works

* verify that the bucket "Bucket Key" is "Enabled" by going to the bucket on AWS console
* under the "Properties" tab, look for "Default encryption"
* the last field under "Default encryption" is "Bucket Key" - make sure it says "Enabled" under it

this PR enables bucket key for all types of encryption (SSE:KMS and SSE:S3) so
please also verify that only `encrypt: true` works (without specifying a `keyID`, this will use SSE:S3)

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -794,12 +794,14 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 			encryptionType = s3.ServerSideEncryptionAes256
 		}
 
+		enableBucketKey := true
 		_, err = svc.PutBucketEncryptionWithContext(d.Context, &s3.PutBucketEncryptionInput{
 			Bucket: aws.String(d.Config.Bucket),
 			ServerSideEncryptionConfiguration: &s3.ServerSideEncryptionConfiguration{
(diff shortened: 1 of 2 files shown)


flavianmissi:
this is for ROSA - we want to backport it all the way to 4.13.
we'll start with 4.15.z.
/cherry-pick release-4.15

PR #996, opened 2024-02-09
NO-JIRA: Add hack/local-dev.sh

dmage:


flavianmissi:
this is for development environments only and should not affect the actual product or users.
/label docs-approved
/label px-approved
/label qe-approved

PR #998, opened 2024-02-09
OCPBUGS-29003: move azure storage blobs from `docker` back into `/docker`

flavianmissi:
This undoes the effects caused by bug OCPBUGS-29003.
It needs to land in tandem with https://github.com/openshift/image-registry/pull/392.

-----

### Testing
**On a 4.13.z cluster**, push a few images to the internal registry.
(Optional) Start some deployments from the pushed images.
**Upgrade to 4.14.z** (NOTE! use a 4.14.z that **does not contain** the bug fix).
Pulling will no longer work for images pushed on 4.13.z.
Push new images. To test more parts of the code, ensure that you push both images that were not pushed to 4.13.z as well as SOME images that were already pushed to 4.13.z (NOTE! make sure you still have images pushed to 4.13.z that cannot be pulled at this point).
**Update the cluster to a 4.14.z version that contains the fix.** I think if you use ClusterBot to create a release build, then you can upgrade to that build using oc adm upgrade --to-image=<custom-built-release-image> (but I have not tried this).

While the migration job is running, the operator will change to "Progressing: true". Once the migration job successfully finishes running, the operator should become available again.

**Test the above steps, but using with a cluster wide proxy.**
If you know other configurations for the registry that could affect the migration pod job, test them too.

**Test deploying a cluster containing the fix to a different cloud.** The migration job azure-path-fix should not be created and the operator should become available as it normally would.


flavianmissi:
azure tests are failing due to:
```
AzurePathFixControllerDegraded: Migration failed: /bin/sh: line 1: /usr/bin/move-blobs: No such file or directory
```
I got this error during my tests and the reason was that the image used by the controller to launch the job container was not the same as the image I had built. I need to investigate and see if the same is happening on the job.

flavianmissi:
I don't know why yet, but the image used by the job pod is (and it's fetched from an env var `IMAGE`):
```
registry.build05.ci.openshift.org/ci-op-w54twd3x/stable@sha256:ee79443ab0a07572145d49943a1d5c594cf33af3f9e4470397a7611a4784f54b
```
but the image used by the operator pod is:
```
registry.build05.ci.openshift.org/ci-op-w54twd3x/stable@sha256:a33923b7600aa3f970c05293a54ad0ff8956b2bef92982cb8204422fa38e4633
```
we want them to match.

flavianmissi:
Looks like the IMAGE env var is the image-registry image. I'm looking into the best way to use the operator image for the job instead.

PR #1006, opened 2024-02-26
OCPBUGS-29932: cmd/move-blobs: log and exit 1 on error instead of panic

flavianmissi:


deads2k:
structurally in master, I would suggest following the `oc`/`kubectl` pattern of main calling to a single function that returns errors

/lgtm

flavianmissi:
> structurally in master, I would suggest following the oc/kubectl pattern of main calling to a single function that returns errors

Thanks, I'll keep that in mind. This command was written specifically to migrate blobs affected by a bug in 4.14 so I'm not planning on keeping it long-term.

/cherry-pick release-4.15

PR #1013, opened 2024-03-13
NO-JIRA: bump golangci-lint to v1.56.2

flavianmissi:
also run gofumpt.


flavianmissi:
@wewang58 this does not affect the product in any way, it's just bumping a linter. can you qe approve?
/assign @wewang58

PR #1027, opened 2024-04-23
CFE-962: Refactor code to mock GCP tag service for UTs

bharath-b-rh:
PR contains below changes:
- GCP Resource tags specific code refactoring to mock services to add unit test cases.
- Added permissions required for fetching tags and creating tag bindings.
- Removed checks for unsupported regions `us-east2` and `us-east3`, as these are not officially regions and restriction has also been removed from bucket doc [here](https://cloud.google.com/storage/docs/tags-and-labels#tags)

### DIFF
--- a/pkg/storage/gcs/gcp_labels_tags.go
+++ b/pkg/storage/gcs/gcp_labels_tags.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
-	"strings"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
@@ -42,6 +41,7 @@ const (
 
 	// gcpMaxTagsPerResource is the maximum number of tags that can
 	// be attached to a resource.
+	// https://cloud.google.com/resource-manager/docs/limits#tag-limits
 	gcpMaxTagsPerResource = 50
 
 	// gcpTagsRequestRateLimit is the tag request rate limit per second.
@@ -58,6 +58,36 @@ const (
 	bucketParentPathFmt = "//storage.googleapis.com/projects/_/buckets/%s"
 )
 
+// UserTagsNotDefined is returned when user defined tags is empty; used for updating
+// status condition.
+var UserTagsNotDefined = errors.New("user did not define any tags")
+
+// TagService is the interface that wraps methods for resource tag operations.
+type TagService interface {
+	AddTagsToStorageBucket(context.Context, *imageregistryv1.Config) error
+	Close()
+}
+
+// TagBindingsService is the interface that wraps methods for resource tag binding operations.
+type TagBindingsService interface {
+	DeduplicateTags(context.Context, string, []string) []string
+	CreateTagBindings(context.Context, string, []string) error
+	Close()
+}
+
+// tagServiceManager handles resource tagging.
+type tagServiceManager struct {
+	Listers           *regopclient.StorageListers
+	tagBindingsClient TagBindingsService
+}
+
+// tagBindingsClient handles resource tag bindings.
+type tagBindingsClient struct {
+	*rscmgr.TagBindingsClient
+}
+
+// getUserLabels returns the user defined labels in status subresource of
+// infrastructure/cluster resource, along with the default labels defined in OCP.
 func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]string, error) {
 	infra, err := util.GetInfrastructure(infraLister)
 	if err != nil {
@@ -78,10 +108,10 @@ func getUserLabels(infraLister configlisters.InfrastructureLister) (map[string]s
 	return labels, nil
 }
 
-// newLimiter returns token bucket based request rate limiter after initializing
+// newRequestLimiter returns token bucket based request rate limiter after initializing
 // the passed values for limit, burst(or token bucket) size. If opted for emptyBucket
 // all initial tokens are reserved for the first burst.
-func newLimiter(limit, burst int, emptyBucket bool) *rate.Limiter {
+func newRequestLimiter(limit, burst int, emptyBucket bool) *rate.Limiter {
 	limiter := rate.NewLimiter(rate.Every(time.Second/time.Duration(limit)), burst)
 
 	if emptyBucket {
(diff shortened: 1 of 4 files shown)


pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -170,120 +187,195 @@ func filterTagList(ctx context.Context, client *rscmgr.TagBindingsClient, parent
...
+// to the passed GCP bucket resource. It's wrapper around addUserTagsToStorageBucket()
+// and additionally updates status condition in image registry Config resource.
+func (t *tagServiceManager) AddTagsToStorageBucket(ctx context.Context, cr *imageregistryv1.Config) error {
+	if err := t.addTagsToStorageBucket(ctx, cr); err != nil {
+		if errors.Is(err, NoTagsDefined) {
+			util.UpdateCondition(cr, defaults.StorageTagged, operatorapi.ConditionFalse,
+				gcpTagsSuccessStatusReason, NoTagsDefined.Error())
+			return nil
+		}
+		util.UpdateCondition(cr, defaults.StorageTagged, operatorapi.ConditionTrue,
```
flavianmissi, line 372: should the condition really be true here, i.e `operatorapi.ConditionTrue`? 
since there is an error, I'd expect the condition to be `operatorapi.ConditionFalse`. 
WDYT?

flavianmissi (commented):
In this PR's description, you list 4 things that changed here. Could you separate your commits to also be broken down into those 4 things you listed? The refactoring (which I appreciate) is very hard to review when it's lumped together with other changes.
I left a few comments but have not been able to go through the whole thing yet.

bharath-b-rh:
> In this PR's description, you list 4 things that changed here. Could you separate your commits to also be broken down into those 4 things you listed? The refactoring (which I appreciate) is very hard to review when it's lumped together with other changes. I left a few comments but have not been able to go through the whole thing yet.

The PR description had below conflicting changes listed and I have updated it (by removing (1)).

> (1)Code changes to update status condition when adding tags was skipped due to buckets created in tags unsupported region
(2)Removed checks for unsupported regions us-east2 and us-east3, as these are not officially regions and restriction has also been removed from bucket doc [here](https://cloud.google.com/storage/docs/tags-and-labels#tags)

I have also tried adding different commits for the specific changes. Please let me know WDYT.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: I am using different reasons for success and failure which are `SuccessTaggingBucket` and `ErrorTaggingBucket`. And when there is an error, `ErrorTaggingBucket` reason is used and condition is set to true. Would it better if I just use `SuccessTaggingBucket` and set condition to `false` and message with `error string` when an error happens?

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -170,120 +183,190 @@ func filterTagList(ctx context.Context, client *rscmgr.TagBindingsClient, parent
...
 			errFlag = true
-			klog.Errorf("failed to add %s tag to %s bucket: %v", value, bucketName, err)
+			klog.Errorf("failed to add %s tag to %s bucket: %v", tag, parent, err)
 		}
-		klog.V(1).Infof("binding tag %s to %s bucket successful", value, bucketName)
+		klog.Infof("successfully added %s tag to %s bucket", tag, parent)
 	}
 	if errFlag {
-		return fmt.Errorf("failed to add tag(s) to %s bucket", bucketName)
+		return fmt.Errorf("failed to add tag(s) to %s bucket", parent)
```
flavianmissi, line 281: what's the motivation to rename `bucketName` to `parent`? I thought `bucketName` was more clear 🤔

pkg/storage/gcs/gcp_labels_tags.go
flavianmissi: Hmm I'm not sure I follow, maybe I'm misunderstanding things.
Doesn't `defaults.StorageTagged` condition set to `operatorapi.ConditionTrue` read as "the storage was tagged"? This is my confusion, please help me understand 😅

flavianmissi (commented):
Haven't gone through all commits yet - leaving some questions now to get the conversation going.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: gapi's refer the resource to which tag will be attached as parent, so thought to use the same. But yeah since currently it's just the bucket which is tagged and logs also mention has bucket, bucketName seems better. I will change it to bucketName and use parent in the wrapper method for gapi.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: Yeah, this will cause confusion. I was just reading reason and status and thought it to be apt and forgot about condition. Setting status to false during error scenario is the correct thing. I will update the code :)

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: I updated it to resourceName to keep it generic. Please let me know WDYT

flavianmissi (commented):
Very nice work with the unit tests @bharath-b-rh, thank you for putting the time to write them well!

While I continue the work on reviewing this, could you provide some test instructions for QE? This way we're ready for the next step (QE testing) when the time comes.
You will want to describe how to exercise the changes in this PR, and what to expect from the changes. It's probably best to do this in the reference Jira ticket, sometimes just a list of acceptance criteria might be enough to get QE started.

Thanks!

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -170,120 +184,198 @@ func filterTagList(ctx context.Context, client *rscmgr.TagBindingsClient, parent
...
+func (t *tagServiceManager) Close() {
+	t.tagBindingsClient.Close()
+}
+
+// AddTagsToStorageBucket adds the user-defined tags in the Infrastructure resource
+// to the passed GCP bucket resource. It's wrapper around addUserTagsToStorageBucket()
+// and additionally updates status condition in image registry Config resource.
+func (t *tagServiceManager) AddTagsToStorageBucket(ctx context.Context, cr *imageregistryv1.Config) error {
+	if err := t.addTagsToStorageBucket(ctx, cr); err != nil {
+		util.UpdateCondition(cr, defaults.StorageTagged, operatorapi.ConditionFalse,
```
flavianmissi, line 381: What do you think of moving the `util.UpdateCondition` calls from here into the caller (`CreateStorage` in gcs.go)? 
I ask because when reading the `CreateStorage`, when I see a call to `AddTagsToStorageBucket` I wouldn't expect this call to update operator conditions, so I think it would be more straight forward if this call returned an appropriate error, and the caller used this error to update the conditions accordingly.

flavianmissi (commented):
We're close! Left a comment, but other than that everything looks pretty neat.

pkg/storage/gcs/gcp_labels_tags.go
bharath-b-rh: Yeah, `CreateStorage` is having calls to `UpdateCondition` and it would be uniform.
Reason I added here was because, `AddTagsToStorageBucket` is called twice in `CreateStorage` and `StorageTagged` condition is [updated](https://github.com/openshift/cluster-image-registry-operator/pull/1027/commits/f4a77b05243f1b0b715d8f19eb6e66c9e226feb1#diff-6e6f5be6182d62cd1ad9867497387a3c45de233050881e4479396cea0652d4a5R374-R390) for three different reasons and thought to handle it in `AddTagsToStorageBucket`.

I could define a new method `UpdateTagCondition` in `gcs` package to avoid code repetition or add a comment at `AddTagsToStorageBucket` invoked places saying `UpdateCondition` is handled by the method. Please let me know WDYT.

bharath-b-rh:
> Very nice work with the unit tests @bharath-b-rh, thank you for putting the time to write them well!
> 
> While I continue the work on reviewing this, could you provide some test instructions for QE? This way we're ready for the next step (QE testing) when the time comes. You will want to describe how to exercise the changes in this PR, and what to expect from the changes. It's probably best to do this in the reference Jira ticket, sometimes just a list of acceptance criteria might be enough to get QE started.
> 
> Thanks!

I had reached out to QE about the changes, and pre-merge verification was completed and `qe-approved` label is added. QE also has updated JIRA with the test results. 

Also reached out to docs team to update below change. 

> Removed checks for unsupported regions us-east2 and us-east3, as these are not officially regions and restriction has also been removed from bucket doc [here](https://cloud.google.com/storage/docs/tags-and-labels#tags)

pkg/storage/gcs/gcp_labels_tags.go
flavianmissi: I like the idea of creating an `UpdateTagCondition` function in the `gcs` package. Since it'll only be used in the `gcs` package we should make internal, so it'd be `updateTagCondition` instead. Then you can call that from `CreateStorage`.
How does that sound?

bharath-b-rh: Yeah sounds good, I will make the required changes.

pkg/storage/gcs/gcs.go
```diff
@@ -320,7 +326,8 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 				cr.Spec.Storage.GCS = d.Config.DeepCopy()
 			}
 		}
-		return addTagsToStorageBucket(d.Context, cr, d.Listers, d.Config.Bucket, d.Config.Region)
+		// add user-defined tags to the created storage bucket and update `StorageTagged` condition.
+		return updateTagCondition(cr, tagMgr.AddTagsToStorageBucket(d.Context, cr))
```
flavianmissi, line 330: What do you think of separating this into two calls? I think it would be easier to read. i.e

```suggestion
		err := tagMgr.AddTagsToStorageBucket(d.Context, cr)
		return updateTagCondition(cr, err)
```

pkg/storage/gcs/gcs.go
```diff
@@ -331,8 +338,8 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 
 	// Previous status condition was set to failed because, of tag operation failure,
 	// means bucket creation was successful and hence will only try adding tags.
-	if cond := util.FetchCondition(cr, defaults.StorageTagged); cond.Status != operatorapi.ConditionTrue {
-		return addTagsToStorageBucket(d.Context, cr, d.Listers, d.Config.Bucket, d.Config.Region)
+	if cond := util.FetchCondition(cr, defaults.StorageTagged); cond.Reason == gcpTagsFailedStatusReason && cond.Status == operatorapi.ConditionTrue {
+		return updateTagCondition(cr, tagMgr.AddTagsToStorageBucket(d.Context, cr))
```
flavianmissi, line 342: Same suggestion as above applies here.

bharath-b-rh:
tests are failing with errors unrelated to changes and hypershift CI tests are for other PRs as well.

PR #1028, opened 2024-04-23
OCPBUGS-32710: pkg/storage/s3: use force path style in favour of virtual hosted style config

flavianmissi:
the latest upgrade of our openshift/docker-distribution fork brought in a patch introducing `forcepathstyle` configuration, which was mistakenly merged the openshift only `virtualhostedstyle` configuration.

to workaround the issue we simply drop the use of `virtualhostedstyle` in favour of `forcepathstyle`.

we will eventually remove our custom `virtualhostedstyle` path from our docker-distribution fork, but using
`.spec.storage.s3.virtualHostedStyle` should work as expected regardless.

### DIFF
--- a/pkg/resource/podtemplatespec_test.go
+++ b/pkg/resource/podtemplatespec_test.go
@@ -521,7 +521,7 @@ func TestMakePodTemplateSpecS3CloudFront(t *testing.T) {
 		"REGISTRY_STORAGE_S3_BUCKET":                {Value: "bucket"},
 		"REGISTRY_STORAGE_S3_REGION":                {Value: "region"},
 		"REGISTRY_STORAGE_S3_ENCRYPT":               {Value: "true"},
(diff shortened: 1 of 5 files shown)


flavianmissi:
let's hold for pre-merge testing
/hold

flavianmissi:
FTR this needs to be tested in both AWS S3 as well as IBM COS (apparently the virtual hosted style was hard coded there to false)

xiuwang:
@flavianmissi Hi, could I understand if REGISTRY_STORAGE_S3_FORCEPATHSTYLE=true, our retrieve blob address should be like https://s3-endpoint.etc/bucket-name/docker/registry/v2/blobs/sha256/<>/data ?

flavianmissi:
Hi @xiuwang - yes that is my understanding as well!

flavianmissi:
We're seeing failures on a test in vsphere in another PR - I'll try to run that same test against this PR to see if I can reproduce the issue (and confirm it's unrelated to that PR).
/test e2e-vsphere-operator

PR #1055, opened 2024-06-07
TRT-1576: Ensure affinity rules are applied for >=2 replicas

DennisPeriquet:
In [this PR in release repo](https://github.com/openshift/release/pull/52731/commits/caaa8089880f747e6775ac086ca96f7083928afe), we tried to set image-registry replicas to 3 but the anti-affinity rules did not get applied so we hard coded them there.  The [passing rehearse job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/52731/rehearse-52731-pull-ci-openshift-origin-release-4.17-e2e-gcp-ovn-techpreview-serial/1798677279266574336) on that release PR shows that the [image-registry deployment](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/52731/rehearse-52731-pull-ci-openshift-origin-release-4.17-e2e-gcp-ovn-techpreview-serial/1798677279266574336/artifacts/e2e-gcp-ovn-techpreview-serial/gather-extra/artifacts/deployments.json) (search for `"uid": "df518b63-2a03-4b12-9614-39c40646dc2b"`) has the anti-affinity rules applied.

The operator seems like the right place to apply those rules so that it can support any number of replicas beyond 1.

### DIFF
--- a/pkg/resource/podtemplatespec.go
+++ b/pkg/resource/podtemplatespec.go
@@ -440,9 +440,9 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 
 	// if user has provided an affinity through config spec we use it here, if not
 	// then we fallback to a preferred affinity configuration. we only require a
-	// certain affinity during schedule if the number of replicas is defined to two.
+	// certain affinity during schedule if the number of replicas is 2 or more.
 	affinity := cr.Spec.Affinity
-	if affinity == nil && cr.Spec.Replicas == 2 {
+	if affinity == nil && cr.Spec.Replicas >= 2 {
 		affinity = &corev1.Affinity{
 			PodAntiAffinity: &corev1.PodAntiAffinity{
 				RequiredDuringSchedulingIgnoredDuringExecution: []corev1.PodAffinityTerm{


flavianmissi:
I think there is a concern about replicas == 2 on SNO clusters, but I can't say I fully understand it. I pinged @dmage, hoping he can provide more context before we can proceed with this change.
For reference, this is the commit that introduced the check: https://github.com/openshift/cluster-image-registry-operator/blame/8302fdf8ea856afa92513dae6b14c05e71e2a914/pkg/resource/podtemplatespec.go

flavianmissi:
@DennisPeriquet the reason why we only apply affinity rules when replicas == 2 is because if we have 3 nodes and 3 replicas (and affinity rules) we have run into scheduling issues during upgrades where new replicas cannot be scheduled because of the affinity rules.
I think we might need to find a different way to solve the situation described in https://issues.redhat.com/browse/TRT-1576 (pod disruption budget?).
Let me know your thoughts.

DennisPeriquet:
Hi @flavianmissi and thanks for the comment.  We also tried this [PR in release repo](https://github.com/openshift/release/pull/52731/files) which sets the replica count to 3 at installation and adds the same affinity rules as used by the operator. The rehearsal [job](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/52731/rehearse-52731-pull-ci-openshift-origin-release-4.17-e2e-gcp-ovn-techpreview-serial/1798677279266574336) that uses the change shows that it seems to be ok with the 3 replicas and the tainting of 2 worker nodes.

Would it make more sense to make the change in the release repo (i.e., CI) vs. in the operator itself? This way, it will be CI specific but not a general change.

flavianmissi:
Hi @DennisPeriquet, have you reheased any upgrades? I think that an upgrade from a version x.y with affinit, 3 replicas and 3 worker nodes into version x.y+1 (also with affinity and 3 replicas) will cause availability disruptions because of scheduling + affinity. 
Do you know if there is a way we can test this with payload jobs? I don't mean to rain on your parade, I'm just looking into avoiding issues.

DennisPeriquet:
Hi @flavianmissi -- no worries if this doesn't work out :-) I appreciate your feedback!

we can do `/payload-job <a supported prow job>`. Do you have one in mind?

DennisPeriquet:
If we want to test `replicas: 3` with anti-affinity rules, we can use the [release PR](https://github.com/openshift/release/pull/52731) and run a job there.  I started a `/pj-rehearse periodic-ci-openshift-release-master-ci-4.17-e2e-azure-ovn-upgrade`; it seems the last time I tried that it didn't start; hopefully this time it will run.

DennisPeriquet:
[this slack thread](https://redhat-internal.slack.com/archives/C015H2WDJRY/p1718225955928429) shows that we'll have problems with metal and vsphere so we'll hold off until we can figure out a good solution.

PR #1061, opened 2024-06-12
IR-477: pkg/operator: deactivate azure path fix job

flavianmissi:
removes the job if it exists, and does nothing when it doesn't. we added this job to help users work around a breaking change in the uptsream distribution azure storage driver that changed the root storage path.
this is no longer needed at this point, so we remove it.

/assign @dmage

### DIFF
--- a/pkg/operator/azurepathfixcontroller.go
+++ b/pkg/operator/azurepathfixcontroller.go
@@ -1,17 +1,12 @@
 package operator
 
 import (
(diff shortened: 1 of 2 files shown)


flavianmissi:
e2e are flaking
/retest

flavianmissi:
this is not user facing.
/label  docs-approved
/label px-approved

PR #1074, opened 2024-07-12
IR-471: Adding additional validation

deepsm007:
/cc @flavianmissi

### DIFF
--- a/pkg/storage/s3/s3.go
+++ b/pkg/storage/s3/s3.go
@@ -419,7 +419,7 @@ func (d *driver) ConfigEnv() (envs envvar.List, err error) {
 		return
 	}
 
-	if currentFeatureGates.Enabled(util.ChunkSizeMiBFeatureGateName) && d.Config.ChunkSizeMiB > 0 {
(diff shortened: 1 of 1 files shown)


pkg/storage/s3/s3.go
```diff
@@ -419,7 +419,7 @@ func (d *driver) ConfigEnv() (envs envvar.List, err error) {
 		return
 	}
 
-	if currentFeatureGates.Enabled(util.ChunkSizeMiBFeatureGateName) && d.Config.ChunkSizeMiB > 0 {
+	if currentFeatureGates.Enabled(util.ChunkSizeMiBFeatureGateName) && d.Config.ChunkSizeMiB > 4 && d.Config.ChunkSizeMiB < 5121 {
```
flavianmissi, line 422: Is this range (4 and 5121) defined in the API validation as well? Just curious.

deepsm007: Yes, minimum=5 and maximum 5120 in kubebuilder validation

flavianmissi:
this is a validation change that matches the one implemented in the API level, so it's not really customer facing.
/label px-approved
/label docs-approved

PR #1076, opened 2024-07-16
OCPBUGS-37207: Power VS: Check endpoints against lower case strings

mjturek:
Power VS endpoint validation in the API only allows for lower case characters. However, the endpoint struct we check against is hardcoded to be PascalCase. This patch checks against the lower case version of the string to ensure it's recognized.

### DIFF
--- a/pkg/storage/ibmcos/ibmcos.go
+++ b/pkg/storage/ibmcos/ibmcos.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"reflect"
+	"strings"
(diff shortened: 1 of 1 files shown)


flavianmissi:
/retest
/lgtm
@mjturek / @juliemathew what do you think of testing this pre-merge? I'll put it on hold so we can discuss.
/hold

flavianmissi:
Thank you for looking into this one @juliemathew! Also note that the command you're looking for is `/label qe-approved` 😄 
/hold cancel

PR #1020, opened 2024-04-15
IR-467: Enable Azure MSI authentication

bryan-cox:
This PR enables Azure MSI authentication, which will be utilized in the managed service - ARO HCP.

Fixes [IR-467](https://issues.redhat.com/browse/IR-467)

PR #1079, opened 2024-08-01
Revert "IR-467: Enable Azure MSI authentication"

bryan-cox:
Reverts openshift/cluster-image-registry-operator#1020


bryan-cox:
@flavianmissi - while working on enabling MSI for another operator, Cesar and I ran into an issue where we are going to need to come up with another way that is consistent across all the operators that need this enabled. It's probably better that we don't mess with the creds so I opened this revert PR.

I'm going to put together an enhancement so we can all agree on the path forward for this (image registry, storage, ingress, and cloud network config).

PR #1082, opened 2024-08-05
IR-467: Enable MSI override for ARO HCP

bryan-cox:
ARO HCP needs to force the Azure authentication type to MSI for deployments of the cluster-image-registry in ARO HCP control planes. See https://docs.google.com/document/d/1Z7N2LAnRlgSgrFjjl2absOnkGFsI2TMcbwaW_CA1qek/edit#heading=h.bupciudrwmna for more details.

### DIFF
--- a/pkg/storage/azure/azure.go
+++ b/pkg/storage/azure/azure.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os"
 	"reflect"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


flavianmissi:
Thanks for putting this together @bryan-cox ! 
Other the changes you already made, we also need to make sure the client id is exposed as an environment variable to the image registry pods. You can use what's already been done for the workload identities effort as inspiration: https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/azure/azure.go?plain=1#L433-L445

bryan-cox:
> Thanks for putting this together @bryan-cox ! Other the changes you already made, we also need to make sure the client id is exposed as an environment variable to the image registry pods. You can use what's already been done for the workload identities effort as inspiration: https://github.com/openshift/cluster-image-registry-operator/blob/master/pkg/storage/azure/azure.go?plain=1#L433-L445

Hey @flavianmissi - we actually want those other pods to use workload identity and not the MSI identity. We only want MSI for the cluster-image-registry-operator. Microsoft is building us some additional sidecar containers that will take the MSI authentication call and approve it. I linked the SDR around this in the PR description for further details on this.

bryan-cox:
> Thanks for expanding @bryan-cox! 
> 
> So is the federated token also available when MSI is configured? 

@flavianmissi it should be, that is my understanding at least from the SDR.

flavianmissi:
/lgtm

This change is part of an OCP wide effort to support MSI. To the best of my knowledge, PX and docs should be handled at by ARO, not at the component level. Thus I'll provide the PX and docs approval myself. @bryan-cox please let me know if that is not the case.

/label docs-approved
/label px-approved

xiuwang:
/label qe-approved
@bryan-cox  I only do regression for image-registry on azure aks hypershift with this pr built nightly payload.
I will check again when ARO HCP supported MSI.

PR #1092, opened 2024-08-19
OCPBUGS-37543: pkg/storage/azure: do not retry remove storage operation

flavianmissi:
Previous to commit 472515d, `RemoveStorage` never triggered a retry. After removing storage account key usage for workload identity enabled clusters (OCPBUGS-37543), a failed container deletion (caused by a 404) would trigger a retry (which is not a full reconcile in this case, since the operator CR does not get updated), and would cause a permanent `AccountNotFound` error.

/assign @rajdeepc2792 

Holding until QE has verified this fix works.
/hold


flavianmissi:
Closing this - we'll revert the original change instead and get a fix in a follow-up PR.

PR #1105, opened 2024-08-28
[release-4.15] OCPBUGS-38895: pkg/resource: invoke update-ca-trust extract with --output

flavianmissi:
manual cherry pick to include both required commits.

overrides https://github.com/openshift/cluster-image-registry-operator/pull/1099

### DIFF
--- a/pkg/resource/azurepathfixjob.go
+++ b/pkg/resource/azurepathfixjob.go
@@ -255,7 +255,7 @@ func (gapfj *generatorAzurePathFixJob) expected() (runtime.Object, error) {
 							Command: []string{"/bin/sh"},
 							Args: []string{
 								"-c",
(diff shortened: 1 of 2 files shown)


flavianmissi:
For the record: we've seen this working on other aws builds (i.e `e2e-aws-ovn`), so if `e2e-aws-operator ` keep failing we should be good to override it.

flavianmissi:
Launched a cluster with this PR on Azure using cluster-bot: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/release-openshift-origin-installer-launch-azure-modern/1828816393814937600
If the cluster becomes available then we'll know this will have worked.

sdodson:
cluster is alive and well

PR #1106, opened 2024-08-28
[release-4.15] OCPBUGS-39099: Avoid Shared Access Key usage for Azure Storage Account when using Managed Identity based auth

flavianmissi:
note that other than running `go mod vendor` (second commit in this PR), I also needed to bump `go.opentelemetry.io/otel/exporters/otlp/otlptrace` and `go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc` to `v1.19.0` as per https://github.com/open-telemetry/opentelemetry-go/issues/4586 to fix an issue when running `go mod tidy`, this is why the vendored changes are bigger in this PR than in comparison with its sibling backports.

-----

**Note for QE**
Because 4.15 is GA, we cannot backport the "containers/delete" permission. @rajdeepc2792 has wrote the code in a way that storage account deletion should still proceed [when setting managementState: Removed] even after container deletion fails due to lack of permission, so please double check that is the case. If you find that not to be the case, we will still go ahead with this backport but we will create a pre-emptive bug describing the issue with a documented workaround (which we will need to figure out). Though I'm hoping it won't come to that 😅


rajdeepc2792:
@flavianmissi , I didn't face the problem when running `go mod tidy` after just performing the cherry-pick and running `go mod vendor` on my machine.
Although when upgraded the `go get go.opentelemetry.io/otel@v1.24.0`, I faced the issues similar to https://github.com/open-telemetry/opentelemetry-go/issues/4586, but the resolution to update `go.opentelemetry.io/otel/exporters/otlp/otlptrace` and `go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc` to v1.19.0 doesn't work for me.

Although the changes after the vendoring in this particular PR looks correct.

wewang58:
Sorry @flavianmissi  I removed the label, can you see [my comments](https://issues.redhat.com/browse/OCPBUGS-39099?focusedId=25461791&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25461791) in the bug, not sure if it's an effect scenario? thanks

flavianmissi:
failures look unrelated to changes
/retest

flavianmissi:
Closing in favor of https://github.com/openshift/cluster-image-registry-operator/pull/1111, which does not include a bump to otel.
/close

PR #1107, opened 2024-08-28
[release-4.14] OCPBUGS-39100: Avoid Shared Access Key usage for Azure Storage Account when using Managed Identity based auth

flavianmissi:
Note that other than running go mod vendor (second commit in this PR), I also needed to make two manual changes to this PR:
1. bump go.opentelemetry.io/otel/exporters/otlp/otlptrace and go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc to v1.19.0 as per https://github.com/open-telemetry/opentelemetry-go/issues/4586 to fix an issue when running go mod tidy, this is why the vendored changes are bigger in this PR than in comparison with its sibling backports. This was also done on the PR against 4.15
2. Bring the `pkg/storage/azure/azureclient` over to 4.14. This package was created as part of an operator feature available on OCP >= 4.15, and so it doesn't exist on the 4.14 branch and had to be added manually. Go.mod and vendor files also needed to be updated to support this new package. The commits on this PR should clearly reflect what was needed, please let me know otherwise.

----

**Note for QE**
Because 4.14 is GA, we cannot backport the "containers/delete" permission. @rajdeepc2792 wrote the code in a way that storage account deletion should still proceed [when setting managementState: Removed] even after container deletion fails due to lack of permission, so please double check that is the case. If you find that not to be the case, we will still go ahead with this backport but we will create a pre-emptive bug describing the issue with a documented workaround (which we will need to figure out). Though I'm hoping it won't come to that 😅

PR #1112, opened 2024-09-03
[release-4.14] OCPBUGS-39100: Avoid Shared Access Key usage for Azure Storage Account when using Managed Identity based auth

rajdeepc2792:
Alternate PR to https://github.com/openshift/cluster-image-registry-operator/pull/1107
Note that other than running go mod vendor (second commit in this PR), I also needed to make one manual change to this PR:

Bring the pkg/storage/azure/azureclient over to 4.14. This package was created as part of an operator feature available on OCP >= 4.15, and so it doesn't exist on the 4.14 branch and had to be added manually. Go.mod and vendor files also needed to be updated to support this new package.

No need to bump go.opentelemetry.io/otel/exporters/otlp/otlptrace and go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc to v1.19.0 as per https://github.com/open-telemetry/opentelemetry-go/issues/4586. `go mod vendor` and `go mod tidy` were successful without it.

Note for QE
Because 4.14 is GA, we cannot backport the "containers/delete" permission. @rajdeepc2792 wrote the code in a way that storage account deletion should still proceed [when setting managementState: Removed] even after container deletion fails due to lack of permission, so please double check that is the case. If you find that not to be the case, we will still go ahead with this backport but we will create a pre-emptive bug describing the issue with a documented workaround (which we will need to figure out). Though I'm hoping it won't come to that 😅


flavianmissi:
Closed the other backport PR in favor of this one.
/retitle [release-4.14] OCPBUGS-39100: Avoid Shared Access Key usage for Azure Storage Account when using Managed Identity based auth

flavianmissi:
I may have found the source of the missing authentication header. The operator uses https://github.com/jongio/azidext as an adaptor so that it can use recent auth types with our older azure sdk version.
`release-4.14` vendors version `0.4.0` of that library, while `release-4.15` onwards use version `0.5.0`. In the `0.5.0` release of azidext I found this bug reported: https://github.com/jongio/azidext/pull/57. The fix for this bug is the only thing new no `0.5.0`.
I'll try to clone this PR, bump that lib and see where that takes us.

flavianmissi:
The fix on https://github.com/openshift/cluster-image-registry-operator/pull/1114 solved the missing auth header issue 🎉 
/close

PR #1114, opened 2024-09-04
[release-4.14] OCPBUGS-39100: Avoid Shared Access Key usage for Azure Storage Account when using Managed Identity based auth

flavianmissi:
Alternate PR to https://github.com/openshift/cluster-image-registry-operator/pull/1112

Bring the pkg/storage/azure/azureclient over to 4.14. This package was created as part of an operator feature available on OCP >= 4.15, and so it doesn't exist on the 4.14 branch and had to be added manually. Go.mod and vendor files also needed to be updated to support this new package.

No need to bump go.opentelemetry.io/otel/exporters/otlp/otlptrace and go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc to v1.19.0 as per https://github.com/open-telemetry/opentelemetry-go/issues/4586. go mod vendor and go mod tidy were successful without it.

A bump to `azidext` v0.5.0 was required to fix auth header missing errors (caused by the bump of `azcore` to > 1.6.0), see https://github.com/jongio/azidext/pull/57 for details.

----

**Note for QE**
Because 4.14 is GA, we cannot backport the "containers/delete" permission. @rajdeepc2792 wrote the code in a way that storage account deletion should still proceed [when setting managementState: Removed] even after container deletion fails due to lack of permission, so please double check that is the case. If you find that not to be the case, we will still go ahead with this backport but we will create a pre-emptive bug describing the issue with a documented workaround (which we will need to figure out). Though I'm hoping it won't come to that 😅


flavianmissi:
`e2e-aws-operator` failed with: 
```
could not initialize namespace: could not set up namespace for test: Internal error occurred: resource quota evaluation timed out 
```
I'll give it sometime before retesting, we're mostly interested in azure e2e anyways.

flavianmissi:
`e2e-azure-ovn` passed so it looks like bumping azidext was the missing piece of the puzzle!
/retitle [release-4.14] OCPBUGS-39100: Avoid Shared Access Key usage for Azure Storage Account when using Managed Identity based auth

flavianmissi:
/payload-job periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-serial periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-azure-ovn-upgrade periodic-ci-openshift-release-master-ci-4.14-e2e-azure-ovn-upgrade

PR #1120, opened 2024-09-19
OCPBUGS-42196: pkg/storage/azure: use cluster-api tag key to discover vnet

flavianmissi:
the previous tag key seems to no longer be added to clusters created using the cluster-api since 4.17.

---

Making it WIP because we're not sure if this is the right way to go, or if the installer should be backwards compatible and continue tagging the vnet using the previous format.

### DIFF
--- a/pkg/storage/azure/azure.go
+++ b/pkg/storage/azure/azure.go
@@ -675,8 +675,13 @@ func (d *driver) assurePrivateAccount(cfg *Azure, infra *configv1.Infrastructure
 	}
 
 	if internalConfig.VNetName == "" {
+		upstreamTagKey := fmt.Sprintf("sigs.k8s.io_cluster-api-provider-azure_cluster_%s", infra.Status.InfrastructureName)
 		tagKey := fmt.Sprintf("kubernetes.io_cluster.%s", infra.Status.InfrastructureName)
-		vnet, err := azclient.GetVNetByTag(d.Context, networkResourceGroup, tagKey, "owned", "shared")
+		tagFilter := map[string][]string{
+			upstreamTagKey: {"owned", "shared"},
+			tagKey:         {"owned", "shared"},
+		}
+		vnet, err := azclient.GetVNetByTags(d.Context, networkResourceGroup, tagFilter)
 		if err != nil {
 			return "", fmt.Errorf("failed to discover vnet name, please provide network details manually: %q", err)
 		}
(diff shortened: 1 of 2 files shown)


pkg/storage/azure/azure.go
```diff
@@ -675,7 +678,7 @@ func (d *driver) assurePrivateAccount(cfg *Azure, infra *configv1.Infrastructure
 	}
 
 	if internalConfig.VNetName == "" {
-		tagKey := fmt.Sprintf("kubernetes.io_cluster.%s", infra.Status.InfrastructureName)
+		tagKey := fmt.Sprintf(vnetTagKeyTemplate, infra.Status.InfrastructureName)
```
flavianmissi, line 681: The advice from installer folks is to look for both tags. I'll update this code tomorrow.

flavianmissi:
/retitle OCPBUGS-42196: pkg/storage/azure: use cluster-api tag key to discover vnet

pkg/storage/azure/azureclient/azureclient.go
```diff
@@ -163,20 +163,27 @@ func (c *Client) GetVNetByTag(ctx context.Context, resourceGroupName, tagKey str
...
+				tag, ok := vnet.Tags[tagKey]
+				if !ok {
+					continue
+				}
+				for _, tagValue := range tagValues {
+					if *tag == tagValue {
+						return *vnet, nil
+					}
 				}
 			}
```
r4f4, line 182: Not sure if it's worth it, but this helps a bit with the reading:

```suggestion
             if vnetHasAnyTag(vnet, tagFilter) {
                    return *vnet, nil
             }
```
```
func vnetHasAnyTag(vnet *armnetwork.VirtualNetwork, tagFilter map[string][]string) bool {
		for tagKey, tagValues := range tagFilter {
				tag, ok := vnet.Tags[tagKey]
				if !ok {
					continue
				}
				for _, tagValue := range tagValues {
					if *tag == tagValue {
						return true
					}
				}
		}
        return false
}
```

r4f4 (commented):
Small nit comment but

/lgtm

pkg/storage/azure/azureclient/azureclient.go
flavianmissi: Good point, thanks! I'm on it.

flavianmissi: done

r4f4 (commented):
/lgtm

flavianmissi:
Putting on hold so QE has a chance to pre-merge test.
/hold

flavianmissi:
failures look infra related
/retest

flavianmissi:
PR is approved by QE - removing hold
/hold cancel

PR #1121, opened 2024-09-20
CFE-1129: Added AWS TAGS reconciliation 

anirudhAgniRedhat:
cluster-image-registry-operator provisions S3 bucket to use as a registry backend for the cluster registry. During provisioning the operator reads the ResourceTags from the Infrastructure platform status and passes them as arguments during the creation call.
PR is for updating the tags of S3 bucket created by operator whenever ResourceTags in the Infrastructure gets modified.

### DIFF
--- a/pkg/operator/awsTagController.go
+++ b/pkg/operator/awsTagController.go
@@ -30,9 +30,9 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/storage/s3"
 )
 
-// AWSController is for storing internal data required for
+// AWSTagController is for storing internal data required for
 // performing AWS controller operations
-type AWSController struct {
+type AWSTagController struct {
 	infraConfigClient         configv1client.InfrastructureInterface
 	imageRegistryConfigClient imageregistryv1client.ConfigInterface
 	listers                   *regopclient.Listers
@@ -55,9 +55,9 @@ var kubernetesNamespaceRegex = regexp.MustCompile(`^([^/]*\.)?kubernetes.io/`)
 // openshiftNamespaceRegex is used to check that a tag key is not in the openshift.io namespace.
 var openshiftNamespaceRegex = regexp.MustCompile(`^([^/]*\.)?openshift.io/`)
 
-// NewAWSController is for obtaining AWSController object
-// required for invoking AWS controller methods.
-func NewAWSController(
+// NewAWSTagController is for obtaining AWSTagController object
+// required for invoking AWS Tag controller methods.
+func NewAWSTagController(
 	infraConfigClient configv1client.InfrastructureInterface,
 	imageRegistryConfigClient imageregistryv1client.ConfigInterface,
 	kubeInformerFactory kubeinformers.SharedInformerFactory,
@@ -68,15 +68,15 @@ func NewAWSController(
 	openshiftConfigManagedKubeInformerFactory kubeinformers.SharedInformerFactory,
 	eventRecorder events.Recorder,
 	featureGateAccessor featuregates.FeatureGateAccess,
-) (*AWSController, error) {
-	c := &AWSController{
+) (*AWSTagController, error) {
+	c := &AWSTagController{
 		infraConfigClient:         infraConfigClient,
 		imageRegistryConfigClient: imageRegistryConfigClient,
 		featureGateAccessor:       featureGateAccessor,
 		event:                     eventRecorder,
 		queue: workqueue.NewNamedRateLimitingQueue(
 			workqueue.DefaultControllerRateLimiter(),
-			"AWSController"),
+			"AWSTagController"),
 	}
 
 	infraConfig := configInformerFactory.Config().V1().Infrastructures()
@@ -118,7 +118,7 @@ func NewAWSController(
 }
 
 // eventHandler is the callback method for handling events from informer
-func (c *AWSController) eventHandler() cache.ResourceEventHandler {
+func (c *AWSTagController) eventHandler() cache.ResourceEventHandler {
 	const workQueueKey = "aws"
 	return cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
@@ -126,7 +126,8 @@ func (c *AWSController) eventHandler() cache.ResourceEventHandler {
 			if !ok || infra == nil {
 				return
 			}
-			if infra.Status.PlatformStatus.AWS != nil && len(infra.Status.PlatformStatus.AWS.ResourceTags) != 0 {
+			if infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil &&
+				len(infra.Status.PlatformStatus.AWS.ResourceTags) != 0 {
 				c.queue.Add(workQueueKey)
 				return
 			}
@@ -140,7 +141,8 @@ func (c *AWSController) eventHandler() cache.ResourceEventHandler {
 			if !ok || newInfra == nil {
 				return
 			}
-			if oldInfra.Status.PlatformStatus.AWS != nil && newInfra.Status.PlatformStatus.AWS != nil {
+			if oldInfra.Status.PlatformStatus != nil && oldInfra.Status.PlatformStatus.AWS != nil &&
+				newInfra.Status.PlatformStatus != nil && newInfra.Status.PlatformStatus.AWS != nil {
 				if !reflect.DeepEqual(oldInfra.Status.PlatformStatus.AWS.ResourceTags, newInfra.Status.PlatformStatus.AWS.ResourceTags) {
 					c.queue.Add(workQueueKey)
 					return
@@ -151,7 +153,7 @@ func (c *AWSController) eventHandler() cache.ResourceEventHandler {
 }
 
 // Run is the main method for starting the AWS controller
-func (c *AWSController) Run(ctx context.Context) {
+func (c *AWSTagController) Run(ctx context.Context) {
 	defer k8sruntime.HandleCrash()
 	defer c.queue.ShutDown()
 
@@ -167,14 +169,14 @@ func (c *AWSController) Run(ctx context.Context) {
 	klog.Infof("Shutting down AWS Controller")
 }
 
-func (c *AWSController) runWorker() {
+func (c *AWSTagController) runWorker() {
 	for c.processNextWorkItem() {
 	}
 }
 
 // processNextWorkItem is for prcessing the event received
 // which blocks until a new item is received
-func (c *AWSController) processNextWorkItem() bool {
+func (c *AWSTagController) processNextWorkItem() bool {
 	obj, shutdown := c.queue.Get()
 	if shutdown {
 		return false
@@ -196,14 +198,14 @@ func (c *AWSController) processNextWorkItem() bool {
 // on receiving a informer event.
 // Fetches image registry config data, required for obtaining
 // the S3 bucket configuration and creating a driver out of it
-func (c *AWSController) sync() error {
+func (c *AWSTagController) sync() error {
 	return c.syncTags()
 }
 
 // syncTags fetches user tags from Infrastructure resource, which
 // is then compared with the tags configured for the created S3 bucket
 // fetched using the driver object passed and updates if any new tags.
-func (c *AWSController) syncTags() error {
+func (c *AWSTagController) syncTags() error {
 	cr, err := c.imageRegistryConfigClient.Get(
 		context.Background(),
 		defaults.ImageRegistryResourceName,
--- a/pkg/operator/awstagcontroller.go
+++ b/pkg/operator/awstagcontroller.go
@@ -299,7 +299,7 @@ func validateUserTag(key, value string) error {
 		return fmt.Errorf("key has invalid characters or length")
 	}
 	if strings.EqualFold(key, "Name") {
-		return fmt.Errorf("key cannot be customized by user")
+		return fmt.Errorf("name key is not allowed for user defined tags")
 	}
 	if !tagValRegex.MatchString(value) {
 		return fmt.Errorf("value has invalid characters or length")
--- a/pkg/operator/awstagcontroller.go
+++ b/pkg/operator/awstagcontroller.go
@@ -157,16 +157,16 @@ func (c *AWSTagController) Run(ctx context.Context) {
 	defer k8sruntime.HandleCrash()
 	defer c.queue.ShutDown()
 
-	klog.Infof("Starting AWS Controller")
+	klog.Infof("Starting AWS Tag Controller")
 	if !cache.WaitForCacheSync(ctx.Done(), c.cachesToSync...) {
 		return
 	}
 
 	go wait.Until(c.runWorker, time.Second, ctx.Done())
 
-	klog.Infof("Started AWS Controller")
+	klog.Infof("Started AWS Tag Controller")
 	<-ctx.Done()
-	klog.Infof("Shutting down AWS Controller")
+	klog.Infof("Shutting down AWS Tag Controller")
 }
 
 func (c *AWSTagController) runWorker() {
@@ -183,13 +183,13 @@ func (c *AWSTagController) processNextWorkItem() bool {
 	}
 	defer c.queue.Done(obj)
 
-	klog.V(5).Infof("AWSController: got event from workqueue")
+	klog.V(5).Infof("AWSTagController: got event from workqueue")
 	if err := c.sync(); err != nil {
 		c.queue.AddRateLimited(workqueueKey)
-		klog.Errorf("AWSController: failed to process event: %s, requeuing", err)
+		klog.Errorf("AWSTagController: failed to process event: %s, requeuing", err)
 	} else {
 		c.queue.Forget(obj)
-		klog.V(5).Infof("AWSController: event from workqueue successfully processed")
+		klog.V(5).Infof("AWSTagController: event from workqueue successfully processed")
 	}
 	return true
 }
@@ -234,10 +234,8 @@ func (c *AWSTagController) syncTags() error {
 		return err
 	}
 
-	// tags deletion is not supported. Should the user remove it from
-	// PlatformStatus will be looked up for retaining the tag
-	infraTagSet := make(map[string]string)
-	mergePlatformStatusTags(infra, infraTagSet)
+	// Filtering tags based on validation
+	infraTagSet := filterPlatformStatusTags(infra)
 	klog.V(5).Infof("tags read from Infrastructure resource: %v", infraTagSet)
 
 	// Create a driver with the current configuration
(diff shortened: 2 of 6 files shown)


anirudhAgniRedhat:
Hey @flavianmissi Can you give initial review for this Thanks!!

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+		defaults.InfrastructureResourceName,
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		klog.Errorf("failed to fetch Infrastructure resource: %v", err)
+		return err
+	}
+
+	// tags deletion is not supported. Should the user remove it from
+	// PlatformStatus will be looked up for retaining the tag
```
flavianmissi, line 238: I don't really understand the second sentence in this comment:
> Should the user remove it from PlatformStatus will be looked up for retaining the tag

Could you review and perhaps rephrase it?

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+// processNextWorkItem is for prcessing the event received
+// which blocks until a new item is received
+func (c *AWSTagController) processNextWorkItem() bool {
+	obj, shutdown := c.queue.Get()
+	if shutdown {
+		return false
+	}
+	defer c.queue.Done(obj)
+
+	klog.V(5).Infof("AWSController: got event from workqueue")
```
flavianmissi, line 186: Here and everywhere in this file:

```suggestion
	klog.V(5).Infof("AWSTagController: got event from workqueue")
```

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+	klog.Infof("Starting AWS Controller")
+	if !cache.WaitForCacheSync(ctx.Done(), c.cachesToSync...) {
+		return
+	}
+
+	go wait.Until(c.runWorker, time.Second, ctx.Done())
+
+	klog.Infof("Started AWS Controller")
+	<-ctx.Done()
+	klog.Infof("Shutting down AWS Controller")
```
flavianmissi, line 169:
```suggestion
	klog.Infof("Started AWS Tag Controller")
	<-ctx.Done()
	klog.Infof("Shutting down AWS Tag Controller")
```

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+
+	return nil
+}
+
+// mergePlatformStatusTags is for reading and merging user tags present in both
+// Platform Status of Infrastructure config.
+// There could be scenarios(upgrade, user deletes) where user tags could be missing
+// from the Platform Status, hence using Status too to avoid said scenarios.
+// If a tag exists in Status.
+func mergePlatformStatusTags(infra *configv1.Infrastructure, infraTagSet map[string]string) {
```
flavianmissi, line 274: Is there a specific reason why this function receives `infraTagSet` as a parameter, instead of simply returning it? 
The caller (on line 239) declares `infraTagSet` before calling `mergePlatformStatusTags`, i.e:
```
infraTagSet := make(map[string]string)
mergePlatformStatusTags(infra, infraTagSet)
```
So I'm inclined to think there is no reason to receive this as a parameter, but instead declare `infraTagSet` inside `mergePlatformStatusTags` and simply return it, i.e:
```suggestion
func mergePlatformStatusTags(infra *configv1.Infrastructure) map[string]string {
	infraTagSet := make(map[string]string)
```
Then in the end of the function just `return infraTagSet`.

There's also room for improvement in the function name. This function seem to be more about filtering out invalid tags from the platform status... How about naming it something like `filterPlatformStatusTags` instead?

WDYT?

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+		}
+		klog.Infof("successfully updated/deleted %d tags, tagset: %+v", tagUpdatedCount, s3TagSet)
+		c.event.Eventf("UpdateAWSTags",
+			"Successfully updated/deleted tagset of %s s3 bucket", driver.ID())
+	}
+
+	return nil
+}
+
+// mergePlatformStatusTags is for reading and merging user tags present in both
```
flavianmissi, line 269:
```suggestion
// mergePlatformStatusTags is for reading and merging user tags present in
```

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+// Platform Status of Infrastructure config.
+// There could be scenarios(upgrade, user deletes) where user tags could be missing
+// from the Platform Status, hence using Status too to avoid said scenarios.
```
flavianmissi, line 272: I don't understand this sentence:

> There could be scenarios(upgrade, user deletes) where user tags could be missing from the Platform Status, hence using Status too to avoid said scenarios. 

How do you avoid a scenario where user tags could be missing from the platform status by _using platform status_? What I am missing?

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+// If a tag exists in Status.
```
flavianmissi, line 273: This sentence seems incomplete. Could you please review and perhaps rephrase it?

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+		}
+		val, ok := s3TagSet[key]
+		if !ok || val != value {
+			klog.V(5).Infof("%s tag will be added/updated with value %s", key, value)
+			s3TagSet[key] = value
+			tagUpdatedCount++
+		}
+	}
+	return
+}
```
flavianmissi, line 304: I'm a bit surprised to see that this function actually updates `s3TagSet`. The name does not indicate any potential mutation to its parameters. It would be more clear if this function would 1) not mutate the parameters; and 2) be more explicitly named.

For the parameter mutation, we could simply use a new map instead of mutating `s3TagSet`. 
Here's what it could look like:
```suggestion
// syncInfraTags synchronizes the tags obtained from S3 bucket and Infrastructure CR.
// Returns a new map containing the new tag set, omitting tags that were deleted
// from the infrastructure CR, and adding or updating tags that are new or have been modified.
func syncInfraTags(s3TagSet map[string]string, infraTagSet map[string]string) (newTagSet map[string]string, tagUpdatedCount int) {
	for key, value := range infraTagSet {
		// If a tag is value is empty, it's marked for deletion,
		// so we omit it from the new tag set.
		if value == "" {
			klog.V(5).Infof("%s tag will be deleted", key)
			tagUpdatedCount++
			continue
		}
		val, ok := s3TagSet[key]
		if !ok || val != value {
			klog.V(5).Infof("%s tag will be added/updated with value %s", key, value)
			newTagSet[key] = value
			tagUpdatedCount++
		}
	}
	return
}
```

The caller of this function would also need to change if you adopt this suggestion so that uses the returned map to update the S3 tags.

WDYT?

pkg/operator/awsTagController.go
```diff
@@ -0,0 +1,324 @@
...
+
+// validateUserTag is for validating the user defined tags in Infrastructure CR
+func validateUserTag(key, value string) error {
+	if !tagKeyRegex.MatchString(key) {
+		return fmt.Errorf("key has invalid characters or length")
+	}
+	if strings.EqualFold(key, "Name") {
+		return fmt.Errorf("key cannot be customized by user")
```
flavianmissi, line 312: So `key` cannot be `"Name"`? Why can't the user customize it? I thought that was the point of this feature 🤔

flavianmissi (commented):
Didn't have time to go through the whole thing, but left some comments to get the review started.


---


Files under `pkg/operator` are all lowercase and without a word separator. 
Example: `pkg/operator/imageregistrycertificates.go` , `pkg/operator/controllerimagepruner.go`.
Could you rename `pkg/operator/awsTagController.go` so that it follows this naming convention? It should be `pkg/operator/awstagcontroller.go`.

pkg/operator/awsTagController.go
anirudhAgniRedhat: I think this is required I referred this from the installer code https://github.com/openshift/installer/blob/master/pkg/types/aws/validation/platform.go#L64-L91 

CC: @TrilokGeer

pkg/operator/awsTagController.go
anirudhAgniRedhat: I updated the function name and comments. 
So basically I am following the way upstream CAPI has the way to Merge the [tags](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/c2c96dd3d11d57cd2dfa4faaf28f7476c4a2a005/api/v1beta2/tags.go#L65)

pkg/operator/awsTagController.go
anirudhAgniRedhat: Agreed!!

pkg/operator/awsTagController.go
anirudhAgniRedhat: Okay So the case  wanted to explain was as the resources may have some tags which are not there in the infra status so what we need to do here is to append the new tags and their value as it is to the s3TagSet which will eventually have our's tags as well as tags which are not in the infraStatusTagSet.

anirudhAgniRedhat: Anyways rephrased the comment.

pkg/operator/awstagcontroller.go
```diff
@@ -0,0 +1,314 @@
...
+	return tagUpdatedCount
+}
+
+// validateUserTag is for validating the user defined tags in Infrastructure CR
+func validateUserTag(key, value string) error {
+	if !tagKeyRegex.MatchString(key) {
+		return fmt.Errorf("key has invalid characters or length")
+	}
+	if strings.EqualFold(key, "Name") {
+		return fmt.Errorf("key cannot be customized by user")
```
flavianmissi, line 302: Let's make sure this error includes the key, just like the installer one does: https://github.com/openshift/installer/blob/master/pkg/types/aws/validation/platform.go#L78 (thanks for the link btw). We could even use the exact error string.
i.e 
```suggestion
		return fmt.Errorf("Name key is not allowed for user defined tags")
```

pkg/storage/s3/s3.go
```diff
@@ -773,11 +773,12 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 
 		// at this stage we are not keeping user tags in sync. as per enhancement proposal
 		// we only set user provided tags when we created the bucket.
-		hasAWSStatus := infra.Status.PlatformStatus != nil && infra.Status.PlatformStatus.AWS != nil
-		if hasAWSStatus {
-			klog.Infof("user provided %d tags", len(infra.Status.PlatformStatus.AWS.ResourceTags))
+		tagsInSpec := make(map[string]*struct{})
+		if infra.Status.PlatformStatus.AWS != nil && len(infra.Status.PlatformStatus.AWS.ResourceTags) != 0 {
+			klog.V(5).Infof("infra.Spec has %d user provided tags", len(infra.Status.PlatformStatus.AWS.ResourceTags))
```
flavianmissi, line 778: The code is looking at the `infra.Status` but in this log message (and on line 780) it says `infra.Spec` - should we update the log messages to say `infra.Status` instead?

pkg/storage/s3/s3.go
```diff
@@ -773,11 +773,12 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
...
 			for _, tag := range infra.Status.PlatformStatus.AWS.ResourceTags {
-				klog.Infof("user provided bucket tag: %s: %s", tag.Key, tag.Value)
+				klog.Infof("user provided bucket tag in infra.Spec: %s: %s", tag.Key, tag.Value)
+				tagsInSpec[tag.Key] = nil
```
flavianmissi, line 781: Why is `tagsInSpec` needed? It doesn't seem to be used 🤔

pkg/storage/s3/s3.go
```diff
@@ -998,3 +999,76 @@ func sharedCredentialsDataFromStaticCreds(accessKey, accessSecret string) []byte
 
 	return buf.Bytes()
 }
+
+// PutStorageTags is for adding/overwriting tags of the S3 bucket
+// which name is obtained using the ID() method.
```
flavianmissi, line 1004: nit:
```suggestion
// which name is obtained using this driver's ID() method.
```

pkg/storage/s3/s3.go
```diff
@@ -998,3 +999,76 @@ func sharedCredentialsDataFromStaticCreds(accessKey, accessSecret string) []byte
...
+func (d *driver) PutStorageTags(tagList map[string]string) error {
```
flavianmissi, line 1005: nit: It's misleading to call a map variable `tagList` (because it's not a list/slice of tags). We also don't want to repeat the type name in the variable name, so `tagMap` isn't a good choice either.

How about renaming `tagList` to `userTags` ? `desiredTags` is simply `tags` are also good choices.

pkg/storage/s3/s3.go
```diff
@@ -998,3 +999,76 @@ func sharedCredentialsDataFromStaticCreds(accessKey, accessSecret string) []byte
...
+		}
+		return nil, fmt.Errorf("failed to fetch s3 bucket tags: %s", errMsg)
+	}
+
+	tagList := make(map[string]string)
+	for _, tags := range output.TagSet {
+		tagList[aws.StringValue(tags.Key)] = aws.StringValue(tags.Value)
+	}
+
+	return tagList, nil
```
flavianmissi, line 1073: nit: keep naming consistent throughout this package
```suggestion
	tags := make(map[string]string)
	for _, tag := range output.TagSet {
		tag[aws.StringValue(tag.Key)] = aws.StringValue(tag.Value)
	}

	return tags, nil
```

flavianmissi (commented):
The code looks pretty good now, thank you for taking in my suggestions!

I've seen discussion of tag sync only happening in Hypershift, but `AWSTagController` will run on every AWS cluster where `len(infra.Status.PlatformStatus.AWS.ResourceTags) != 0`, so this seem to apply to the whole of OCP. Have we discussed this with OCP's staff engineers to make sure we're on the same page?

pkg/operator/awstagcontroller.go
```diff
@@ -0,0 +1,314 @@
...
+		infraTagSet[statusTags.Key] = statusTags.Value
+	}
+	return infraTagSet
+}
+
+// syncInfraTags synchronizes the tags obtained from S3 bucket and Infrastructure CR.
+// this modifies the s3TagSet based on new tags which are added and update the value to a key if it has changed.
+func syncInfraTags(s3TagSet map[string]string, infraTagSet map[string]string) int {
+	tagUpdatedCount := 0
+	for key, value := range infraTagSet {
```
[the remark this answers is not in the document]
anirudhAgniRedhat: Hey @flavianmissi I had a discussion with @TrilokGeer on merge logic So he just want to sync the tags which are either in infra.Status or in resources which user already have. So we would not delete tags.

anirudhAgniRedhat:
> The code looks pretty good now, thank you for taking in my suggestions!
> 
> I've seen discussion of tag sync only happening in Hypershift, but `AWSTagController` will run on every AWS cluster where `len(infra.Status.PlatformStatus.AWS.ResourceTags) != 0`, so this seem to apply to the whole of OCP. Have we discussed this with OCP's staff engineers to make sure we're on the same page?

Tagging @sdodson since through this changes user can also update tags on all AWS Cluster 
CC: @TrilokGeer

pkg/operator/awstagcontroller.go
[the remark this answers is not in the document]
flavianmissi: Understood! Marking this comment as resolved in that case.

pkg/storage/s3/s3.go
anirudhAgniRedhat: My Bad 
Fixed it!!

flavianmissi:
`e2e-aws-operator` tests are [flaky](https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-cluster-image-registry-operator-master-e2e-aws-operator), but so far the pruner tests in that suite have not been an issue... I do have another PR where I'm also seeing the failure, so it could be something in the platform. I'll make a note of this and try it from the cmd line to see if I can reproduce and better understand the issue.

flavianmissi:
For the record, I'm still waiting on an answer to this question before approval: https://github.com/openshift/cluster-image-registry-operator/pull/1121#pullrequestreview-2330762444

anirudhAgniRedhat:
> For the record, I'm still waiting on an answer to this question before approval: [#1121 (review)](https://github.com/openshift/cluster-image-registry-operator/pull/1121#pullrequestreview-2330762444)

Hey @flavianmissi apologies for delay. I had a discussion with @TrilokGeer on this, as the status field in not usually editable until unless you use the the inadvisable way to do that, the ROSA/HCP will take care to update this status. But I would still wait for @TrilokGeer to confirm this once!

TrilokGeer:
@anirudhAgniRedhat @flavianmissi , in the standalone ocp case, the infrastructure status is supposed to be immutable. With the present feature, user will be able to overcome a situation of updating tags via cluster re-deployment and might prefer to use un-supported tools to edit status and have the tags updated.

anirudhAgniRedhat:
Hey @flavianmissi Is there anything else you think needs to be done before approving this PR?

flavianmissi:
Hey @anirudhAgniRedhat - we need px, docs and qe approval before merging. QE will be assigned automatically once I approve. Can your docs and px (product experience) give the two other labels? Do we need to involve image registry docs here?
/lgtm

anirudhAgniRedhat:
For px Tagging @Senthamilarasu-STA 
For Docs Tagging @skopacz1 or @subhtk If anyone could help!!

anirudhAgniRedhat:
/assign @xiuwang 
For QE approval!! Thanks!!

xiuwang:
@wewang58 will take this test, thanks

wewang58:
Need time to research and think about how to test it.

anirudhAgniRedhat:
> Need time to research and think about how to test it.

Hey @wewang58 Though I have tested it!! you can spin an AWS production Cluster and can update the infra Status.
But here is a catch you cannot update the status by using `oc patch` commands you might need to use either ways below
```
Use  Command kubectl edit-status infrastructure cluster
Might need to install https://github.com/ulucinar/kubectl-edit-status 
```
or  `kubectl edit infrastructure cluster --subresource='status'`
On editing the resource you can add/edit the infra.Status.PlatformStatus.AWS.ResourceTags as
```
resourceTags:
        - key: "Environment"
          value: "Production"
        - key: "Owner"
          value: "DevOpsTeam"
        - key: "Project"
          value: "OpenShift-Cluster"
```

Let me know if you need any further help!!

wewang58:
@anirudhAgniRedhat Thanks  your advice, do I should launch a hcp cluster to test it,right?

wewang58:
@anirudhAgniRedhat Tested AWS Standalone Production Clusters, it looks works, but failed install a hcp cluster, will check with hcp qe to confirm something.
image

wewang58:
If we need to add unit test for some functions?

flavianmissi:
> If we need to add unit test for some functions?

I agree that would be nice :)

anirudhAgniRedhat:
> @anirudhAgniRedhat Tested AWS Standalone Production Clusters, it looks works, but failed install a hcp cluster, will check with hcp qe to confirm something.

You meant HCP installation failed?? Also For HCP installation you would need to update the image-registry-operator image 

The HCP installation though doc which they have will create the managed cluster on 4.17

wewang58:
Tested in the hcp cluster, it works too
image

wewang58:
@anirudhAgniRedhat  Can you assign to me, seems I cannot add qe-approved label.

flavianmissi:
label assignment worked @wewang58! I assign you anyway :) 

/assign @wewang58

PR #1127, opened 2024-09-27
OCPBUGS-42514: azureclient: stop validating credentials when creating the client

flavianmissi:
different operations in Azure require different sets of credentials. some operations for example only need a storage account key to authenticate with Azure API.

since we cannot predict which operation the caller will need, we cannot validate auth related options on client instantiation time.

this didn't use to be an issue because the client was only used for private endpoint configuration, which did not use account key for auth, so validation was more predictable.

### DIFF
--- a/pkg/storage/azure/azureclient/azureclient.go
+++ b/pkg/storage/azure/azureclient/azureclient.go
@@ -2,6 +2,7 @@ package azureclient
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -30,10 +31,9 @@ const (
 )
 
 type Client struct {
-	creds          azcore.TokenCredential
-	clientOpts     *policy.ClientOptions
-	tagset         map[string]*string
-	subscriptionID string
+	creds      azcore.TokenCredential
+	clientOpts *policy.ClientOptions
+	opts       *Options
 }
 
 type Options struct {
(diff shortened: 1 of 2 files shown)


pkg/storage/azure/azureclient/azureclient.go
```diff
@@ -81,63 +81,78 @@ func New(opts *Options) (*Client, error) {
...
-		clientOpts:     &coreOpts,
-		tagset:         opts.TagSet,
-		subscriptionID: opts.SubscriptionID,
+		creds:      creds,
+		clientOpts: &coreOpts,
+		opts:       opts,
 	}, nil
 }
 
+func (c *Client) getCreds() (azcore.TokenCredential, error) {
```
rajdeepc2792, line 95: `getCreds()` will also be needed when `cfg.key` is not present in `NewBlobClient`

flavianmissi: great catch, fixed!

flavianmissi:
test failures look like flakes to me.
/retest

flavianmissi:
Looks like merge conflicts are happening because this change has not yet merged on 4.17, but 4.18 has it: https://github.com/openshift/cluster-image-registry-operator/pull/1125
It's probably easier and safer to wait until that PR merges, then try the automated cherry-pick again, so I'll do that.

PR #1129, opened 2024-10-03
OCPBUGS-42732: pkg/storage/azure: also check for auth failure error code on deletion

flavianmissi:
I cannot explain the behavior, but after `networkAccess` to `Internal` in the operator config, if the operator tries to delete the container (which happens setting `managementState` to `Removed` in the config), the call to `DeleteStorageContainer` fails with a
`bloberror.AuthorizationFailure`, despite the operator having the `containers/delete` permission in its `CredentialsRequest`.

Note that setting `networkAccess` to `Internal` triggers the operator to create a private endpoint in Azure, then disable public access to the storage account in question. This could be related to the issue.


flavianmissi:
can't see how the test failures related to the code changes here.
/retest

PR #1131, opened 2024-10-07
HOSTEDCP-2019: Use Client Cert Auth for ARO HCP deployments

bryan-cox:
Use Client Certificate Authentication for ARO HCP deployments. HyperShift will pass the needed environment variables for this authentication method: ARO_HCP_MI_CLIENT_ID, ARO_HCP_TENANT_ID, and ARO_HCP_CLIENT_CERTIFICATE_PATH.


pkg/storage/util/util.go
```diff
@@ -135,3 +150,98 @@ func GenerateStorageName(listers *regopclient.StorageListers, additionalInfo ...
...
+				if !ok {
+					klog.Errorf("failed to watch config file %s", p)
+					return
+				}
+				klog.Errorf("fsnotify error: %v", err)
+			}
+		}
+	}()
+	return nil
+}
```
flavianmissi, line 247: Things specific to any storage driver do not belong in this package.
I suggest you place them in `pkg/storage/azure/azureclient` instead.

pkg/storage/azure/azureclient/azureclient.go
```diff
@@ -102,47 +102,51 @@ func (c *Client) getCreds() (azcore.TokenCredential, error) {
...
+		ClientOptions: azcore.ClientOptions{
+			Cloud: c.clientOpts.Cloud,
+		},
+	}
+
+	creds, err = util.AzureCreds(options)
+	if err != nil {
+		return nil, err
 	}
+
```
flavianmissi, line 149: I'm hesitant to merge this - it's changes every flavor of auth in the azure driver, which is very invasive and bound to break _something_.

Relatedly, it doesn't look like `NewDefaultAzureCredential` supports `ClientSecretCredential` ? Source: https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/default_azure_credential.go#L39-L57

flavianmissi (commented):
I would be more comfortable merging these changes if we would add a new authentication path on top of the existing ones. This would make it way less likely that we accidentally break something (which has happened a few times in the last few months). We do have test coverage but I doubt our tests cover every edge case.

Have you explored something like that during your implementation process? Are there any cons I'm overlooking?

pkg/storage/azure/azureclient/azureclient.go
bryan-cox: It does. It's just buried down in the environment credentials:
- https://github.com/Azure/azure-sdk-for-go/blob/8759ee622df98a0f205a05de9fc558d3ad0c7d65/sdk/azidentity/default_azure_credential.go#L77
- https://github.com/Azure/azure-sdk-for-go/blob/8759ee622df98a0f205a05de9fc558d3ad0c7d65/sdk/azidentity/environment_credential.go#L102

pkg/storage/util/util.go
bryan-cox: Ok, I actually had it someplace else but was getting a circular dependency.

flavianmissi: Yeah I suppose `pkg/storage/azure` is a logical place to have them, but you'd certainly get a circular import there. `pkg/storage/azure/azureclient` should be safe though.

flavianmissi:
Hi @bryan-cox, are you planning on addressing this [comment](https://github.com/openshift/cluster-image-registry-operator/pull/1131#pullrequestreview-2359596681)?
Relatedly, if your preference would be to change the entire auth path for Azure, will you be available in the future to aid investigation of potential bugs related to this change?

Additionally, what is the automated test plan for this change, if any?

flavianmissi:
worth mentioning that the `e2e-*-operator` tests can be quite flaky and are also failing in a different way each run (_sigh_).
the `e2e-hypershift` failures look unrelated to me.

/test e2e-aws-operator
/test e2e-azure-operator

bryan-cox:
I tested this PR with the following PRs and was able to deploy an Azure Hosted Cluster from an AKS management cluster:

https://github.com/openshift/hypershift/pull/4811
https://github.com/openshift/hypershift/pull/4877
https://github.com/openshift/hypershift/pull/4888

PR #1130, opened 2024-10-07
OCPBUGS-42812: azureclient: stop validating credentials when creating the client

flavianmissi:
This is a manual cherry pick of https://github.com/openshift/cluster-image-registry-operator/pull/1127 (automated cherry pick fail due to git conflict).

-----

different operations in Azure require different sets of credentials. some operations for example only need a storage account key to authenticate with Azure API.

since we cannot predict which operation the caller will need, we cannot validate auth related options on client instantiation time.

this didn't use to be an issue because the client was only used for private endpoint configuration, which did not use account key for auth, so validation was more predictable.

PR #1134, opened 2024-10-09
[release-4.16] OCPBUGS-42933: azureclient: stop validating credentials when creating the client

flavianmissi:
Manually cherry-picking https://github.com/openshift/cluster-image-registry-operator/pull/1130 due to merge conflicts.

---

different operations in Azure require different sets of credentials. some operations for example only need a storage account key to authenticate with Azure API.

since we cannot predict which operation the caller will need, we cannot validate auth related options on client instantiation time.

this didn't use to be an issue because the client was only used for private endpoint configuration, which did not use account key for auth, so validation was more predictable.

### DIFF
--- a/pkg/storage/azure/azureclient/azureclient.go
+++ b/pkg/storage/azure/azureclient/azureclient.go
@@ -2,6 +2,7 @@ package azureclient
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/storage/azure/azureclient/azureclient.go
```diff
@@ -80,50 +80,65 @@ func New(opts *Options) (*Client, error) {
...
+	if c.creds != nil {
+		return c.creds, nil
+	}
+
+	var (
+		err   error
+		creds azcore.TokenCredential
+	)
+
+	if strings.TrimSpace(c.opts.ClientSecret) == "" {
```
flavianmissi, line 104: Note that this condition differs from the one on 4.17+ because the MSI condition had not been backported this far. This is correct and expected.

flavianmissi:
I cherry picked the commit in this PR against 4.15 without conflicts, let's see if we can use the bot from here on 🤞🏼 
/cherry-pick release-4.15 release-4.14

flavianmissi:
test failure looks like a flake
/test e2e-azure-operator

PR #1142, opened 2024-10-17
OCPBUGS-41903: operator/status: clear azure path fix job conditions on operator removal

flavianmissi:
there is sometimes a race condition when setting .spec.managementState back and forth between Removed and Managed where the azure path fix controller will kick off the job, but the resources needed for the job to run will get removed (as expected) before the job can finish and its controller update the operator's progressing condition to reflect that.

this has been happening often in the TestLeaderElection e2e test. this test does not wait for the image registry operator to become available before removing it, and this is what triggers the race. a customer has reported a similar issue, although their error was slightly different and proved harder to reproduce. this commit should fix both problems.

clearing the status is most important when users upgrade from a version of the azure path fix controller that deploys the job. OCP versions that do not deploy the job should not have the problem, and including this code in them should be harmless.
in OCP >= 4.17 this job is no longer deployed, though the controller is kept it can probably be safely removed in OCP >= 4.18.

---

## Testing

We need upgrade tests as well as regression tests for this one. Testing operator removal (`.spec.managementState: Removed`) would also be great.

### DIFF
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -393,6 +393,25 @@ func (c *Controller) syncStatus(
 
 	updateCondition(cr, defaults.OperatorStatusTypeRemoved, operatorRemoved)
 
+	// ensure the operator conditions will not remain progressing or degraded
+	// because the azure path fix job hasn't completed or has failed.
+	// when the operator is set to Removed, it will delete the storage
+	// account that it provisioned (unless .spec.storage.managementState is
+	// Unmanaged). since the storage account is (or will be) removed, there
+	// is no point in reporting the state of the azure path fix job.
+	if cr.Spec.ManagementState == operatorapiv1.Removed && cr.Spec.Storage.ManagementState != imageregistryv1.StorageManagementStateUnmanaged {
+		progressing := "AzurePathFixProgressing"
+		degraded := "AzurePathFixControllerDegraded"
+		progressingConditionFound := v1helpers.FindOperatorCondition(cr.Status.Conditions, progressing) != nil
+		if progressingConditionFound {
+			v1helpers.RemoveOperatorCondition(&cr.Status.Conditions, progressing)
+		}
+		degradedConditionFound := v1helpers.FindOperatorCondition(cr.Status.Conditions, degraded) != nil
+		if degradedConditionFound {
+			v1helpers.RemoveOperatorCondition(&cr.Status.Conditions, degraded)
+		}
+	}
+
 	if deploy == nil {
 		cr.Status.ReadyReplicas = 0
 	} else {
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -174,6 +174,17 @@ func NewController(
 			c.listers.Infrastructures = informer.Lister()
 			return informer.Informer()
 		},
+		func() cache.SharedIndexInformer {
+			// here we don't need the lister but we do want to be informed
+			// when jobs change because of the azure-path-fix job.
+			// this isn't so important in OCP >= 4.17 (because the job
+			// is not deployed starting on 4.17), but in OCP 4.14 through
+			// OCP 4.16 it is vital that this controller gets triggered
+			// when the job status changes, speacially in situations where
+			// the operator is set to Removed before the job is finished.
+			informer := kubeInformerFactory.Batch().V1().Jobs()
+			return informer.Informer()
+		},
 	} {
 		informer := ctor()
 		if _, err := informer.AddEventHandler(c.handler()); err != nil {


flavianmissi:
`e2e-aws-operator` failed on a known flake in `TestImageRegistryRemovedWithImages`.
/test e2e-aws-operator

will re-run `e2e-azure-operator` for good measure.
/test e2e-azure-operator 

`e2e-hypershift` failure seem unrelated
/test e2e-hypershift

flavianmissi:
This seem to behave well overall but I'm still slightly skeptical :D 
/payload-aggregate periodic-ci-openshift-release-master-ci-4.18-upgrade-from-stable-4.17-e2e-azure-ovn-upgrade 5

flavianmissi:
Test failures are infra related - I'll allow CI some time to recover before retesting. 
Failure:
```
error: error creating buildah builder: copying system image from manifest list: determining manifest MIME type for docker://registry.build10.ci.openshift.org/ci/managed-clonerefs@sha256:84b65cb47df5321b0f344d66290144e853eca71055f9553a6056193f6f9ebb6c: reading manifest sha256:9514ef6fdb801fca35c35c6749e6382ecdc0498f09d7b023cb01d7b728454272 in registry.build10.ci.openshift.org/ci/managed-clonerefs: manifest unknown
```

flavianmissi:
/retest
/payload-aggregate periodic-ci-openshift-release-master-ci-4.18-upgrade-from-stable-4.17-e2e-azure-ovn-upgrade 5

flavianmissi:
`e2e-azure-operator` failed because `TestS3Minio` failed. It seems unrelated to the previous flakes. I've seen quite a few errors in the operator logs when it tries to communicate with minio, I suspect the minio deployment might just be unreliable. I've seen `TestS3Minio` fail previously but never spent time investigating it.
This was seen in the logs for this test, which I think corroborates my understanding of the failure:
```
imageregistry.go:399: the imageregistry resource is processed, but the the image registry is not available
```

---

`e2e-aws-operator` failure was infra releated:
```
* could not run steps: step [input:origin-centos-8] failed: failed to wait for importing imagestreamtags on ci-op-nz0w8tln/pipeline:origin-centos-8: failed to import tag(s) [origin-centos-8] on image stream ci-op-nz0w8tln/pipeline because of missing definition in the spec
```

---

/retest

pkg/operator/azurepathfixcontroller.go
```diff
@@ -179,11 +179,11 @@ func (c *AzurePathFixController) sync() error {
 	}
 
 	azureStorage := imageRegistryConfig.Status.Storage.Azure
-	if azureStorage == nil || len(azureStorage.AccountName) == 0 {
-		return fmt.Errorf("storage account not yet provisioned")
-	}
-	if azureStorage == nil || len(azureStorage.Container) == 0 {
-		return fmt.Errorf("storage container not yet provisioned")
+	if azureStorage == nil || len(azureStorage.AccountName) == 0 || len(azureStorage.Container) == 0 {
+		// no need to reque when azure storage isn't configured.
```
atiratree, line 183:
```suggestion
		// no need to re-queue when azure storage isn't configured.
```

pkg/operator/azurepathfixcontroller.go
```diff
@@ -179,11 +179,11 @@ func (c *AzurePathFixController) sync() error {
 	}
 
 	azureStorage := imageRegistryConfig.Status.Storage.Azure
-	if azureStorage == nil || len(azureStorage.AccountName) == 0 {
-		return fmt.Errorf("storage account not yet provisioned")
-	}
-	if azureStorage == nil || len(azureStorage.Container) == 0 {
-		return fmt.Errorf("storage container not yet provisioned")
+	if azureStorage == nil || len(azureStorage.AccountName) == 0 || len(azureStorage.Container) == 0 {
```
atiratree, line 182: Is this going to get backported? Don't we want to ensure that the job is removed in the older versions when someone upgrades and changes the management state to removed?

pkg/operator/status.go
```diff
@@ -393,6 +393,25 @@ func (c *Controller) syncStatus(
 
 	updateCondition(cr, defaults.OperatorStatusTypeRemoved, operatorRemoved)
 
+	// ensure the operator conditions will not remain progressing or degraded
+	// because the azure path fix job hasn't completed or has failed.
+	// when the operator is set to Removed, it will delete the storage
+	// account that it provisioned (unless .spec.storage.managementState is
+	// Unmanaged). since the storage account is (or will be) removed, there
+	// is no point in reporting the state of the azure path fix job.
+	if cr.Spec.ManagementState == operatorapiv1.Removed && cr.Spec.Storage.ManagementState != imageregistryv1.StorageManagementStateUnmanaged {
```
atiratree, line 402: If we expect an update from the job informer to cleanup the conditions, shouldn't we also cleanup the job itself?

pkg/operator/azurepathfixcontroller.go
flavianmissi: Good point... Yes, the idea is to backport this (though we might need to do it manually since the code is different on 4.14 and 4.15).
I initially put this check here as a way to avoid running the entire loop unnecessarily, but this brought more issues than it solved.
What's a good way forward? Should we simply remove the check entirely?

pkg/operator/status.go
flavianmissi: I figured it would probably get GC'd, but that's sloppy of me.
If I remove the early return on https://github.com/openshift/cluster-image-registry-operator/pull/1142/files/163815b88a48d7396a5b92ccc7da0c0e1153aa8d#diff-6b69f0f2a6978665596dfc3f74d52fd4a1aa430198d1f7fb1b84e6b773c9fbc4R182 then I think that should be taken care of. 
WDYT?

pkg/operator/azurepathfixcontroller.go
atiratree: Yeah, I suppose that would make sense. I don't know how big code changes you want to do (considering backporting), but we could even remove the whole `NewGeneratorAzurePathFixJob` part and most of the informers (to prevent superfluous resyncs). Our only inputs are the job name and the config.

pkg/operator/status.go
atiratree: It would be if it was created with that in mind: https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/

pkg/operator/azurepathfixcontroller.go
flavianmissi: I like your suggestions, and I'm open to bigger improvements as long as I can understand and maintain them 😅 

When it comes to removing `NewGeneratorAzurePathFixJob`, in 4.14-4.16 we're still creating the job when it's not there, so it needs to stay, but you're right that from 4.17 onward we do not need it at all. In fact, we can even remove the generator code itself, though I'll handle that on a separate PR to simplify backporting of this one.

I'll let you know when it's ready for another pass.

pkg/operator/status.go
flavianmissi: Ah then we must do this manually indeed. I'm on it.

flavianmissi: I think this might no longer be required after removing the early return on https://github.com/openshift/cluster-image-registry-operator/pull/1142/files#diff-6b69f0f2a6978665596dfc3f74d52fd4a1aa430198d1f7fb1b84e6b773c9fbc4R182 🤔 
When management state is set to Removed, the AzurePathFix controller will run (because it watches the image registry config). Setting management state to Removed clears up `.status.storage`, and since the controller is now allowed to run even though that is no longer set, the job will get removed and the conditions cleared.

I will comment this code for now and run some tests. Things will need to look a bit different in the AzurePathFix controller on 4.14-4.16 (I think we might need to explicitly check for `.spec.managementState: Removed`), but I think that even then we won't need this code block here.

I'll report back with my findings.

flavianmissi: I adapted the code here and applied it onto 4.16: https://github.com/openshift/cluster-image-registry-operator/pull/1158
My tests showed that the fix works without changes to `status.go`. The only changes required are the removal of the early return in cases when storage status is unset, and specific handling of `managementState: Removed`.
I have cleaned up the commits in this PR to reflect these findings.
PTAL @atiratree

flavianmissi:
@xiuwang this PR has changed since your last tagged it, could you please have another look as well? The same code no longer cleanly applies to 4.14-4.16, but I have adapted the changes on https://github.com/openshift/cluster-image-registry-operator/pull/1158 which we should be able to backport between 4.14-4.16.

pkg/operator/azurepathfixcontroller.go
```diff
@@ -90,31 +66,8 @@ func NewAzurePathFixController(
 	}
 	c.cachesToSync = append(c.cachesToSync, imageRegistryConfigInformer.Informer().HasSynced)
 
-	if _, err := infrastructureInformer.Informer().AddEventHandler(c.eventHandler()); err != nil {
```
atiratree, line 93: Just to confirm, we have removed this because we check only for the infra type now, which cannot change, right?

atiratree (commented):
The changes lgtm. Thanks for the update & testing.

pkg/operator/azurepathfixcontroller.go
flavianmissi: Yes, we only need the infrastructure object to check for the platform type which I don't expect to change: https://github.com/openshift/cluster-image-registry-operator/blob/92d64a771380eab0069fca9e4db82c784874bb4c/pkg/operator/azurepathfixcontroller.go?plain=1#L168-L174

Looking at the api implementation for PlatformStatus.Type I can't see any validation that the field cannot be updated, but I cannot think how a cluster's platform can change at runtime in a supported way. Source: https://github.com/openshift/api/blob/caf97963ed302881f01e67b791ae4efc6bcf723c/config/v1/types_infrastructure.go#L391-L402

Let me know if I'm missing something.

flavianmissi: I've updated the code, adding a comment explaining why we don't need an event handler for infrastructure object changes.

flavianmissi:
I'm putting this on hold to give QE some room to test the latest changes.
/hold

@atiratree if you feel the changes are good to go, could you give the PR the lgtm stamp? Tack!

pkg/operator/azurepathfixcontroller.go
atiratree: Ok, cool.

flavianmissi:
`e2e-aws-ovn-image-registry` failed due to infra issue
/retest

flavianmissi:
same issue with `e2e-aws-ovn-image-registry`:
```{  error occurred handling build cluster-image-registry-operator-amd64: build didn't start running within 1h0m0s (phase: Pending):
Found 3 events for Pod cluster-image-registry-operator-amd64-build:
* 0001-01-01T00:00:00Z 0x : 0/87 nodes are available: 1 Insufficient cpu, 1 node(s) had untolerated taint {node-role.kubernetes.io/ci-longtests-worker: ci-longtests-worker}, 1 node(s) had untolerated taint {node-role.kubernetes.io/ci-prowjobs-worker: ci-prowjobs-worker}, 1 node(s) had untolerated taint {node.cloudprovider.kubernetes.io/uninitialized: true}, 12 node(s) had untolerated taint {node-role.kubernetes.io/ci-tests-worker: ci-tests-worker}, 3 node(s) had untolerated taint {node-role.kubernetes.io/infra: reserved}, 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 4 node(s) didn't match Pod's node affinity/selector, 61 node(s) had untolerated taint {node.kubernetes.io/not-ready: }. preemption: 0/87 nodes are available: 1 No preemption victims found for incoming pod, 86 Preemption is not helpful for scheduling.
* 0001-01-01T00:00:00Z 0x : 0/87 nodes are available: 1 Insufficient cpu, 1 node(s) had untolerated taint {node-role.kubernetes.io/ci-longtests-worker: ci-longtests-worker}, 1 node(s) had untolerated taint {node-role.kubernetes.io/ci-prowjobs-worker: ci-prowjobs-worker}, 1 node(s) had untolerated taint {node.cloudprovider.kubernetes.io/uninitialized: true}, 1 node(s) were unschedulable, 11 node(s) had untolerated taint {node-role.kubernetes.io/ci-tests-worker: ci-tests-worker}, 3 node(s) had untolerated taint {node-role.kubernetes.io/infra: reserved}, 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 4 node(s) didn't match Pod's node affinity/selector, 61 node(s) had untolerated taint {node.kubernetes.io/not-ready: }. preemption: 0/87 nodes are available: 1 No preemption victims found for incoming pod, 86 Preemption is not helpful for scheduling.
* 0001-01-01T00:00:00Z 0x : 0/86 nodes are available: 1 Insufficient cpu, 1 node(s) had untolerated taint {node-role.kubernetes.io/ci-longtests-worker: ci-longtests-worker}, 1 node(s) had untolerated taint {node-role.kubernetes.io/ci-prowjobs-worker: ci-prowjobs-worker}, 1 node(s) had untolerated taint {node.cloudprovider.kubernetes.io/uninitialized: true}, 11 node(s) had untolerated taint {node-role.kubernetes.io/ci-tests-worker: ci-tests-worker}, 3 node(s) had untolerated taint {node-role.kubernetes.io/infra: reserved}, 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 4 node(s) didn't match Pod's node affinity/selector, 61 node(s) had untolerated taint {node.kubernetes.io/not-ready: }. preemption: 0/86 nodes are available: 1 No preemption victims found for incoming pod, 85 Preemption is not helpful for scheduling.}```

I'll give CI some time to recover before retesting.

flavianmissi:
This PR can be backported as is to 4.18 and 4.17. I also plan to backport this fix to 4.16, 4.15 and 4.14, but the code will need to change - see https://github.com/openshift/cluster-image-registry-operator/pull/1158.

I'm requesting the bot to backport down to 4.16 here so we have the bugs created by automation, but we'll use https://github.com/openshift/cluster-image-registry-operator/pull/1158 for 4.16 instead (the bot will probably not be able to create a PR, and even if it does we'll have to discard it in favor of 1158). Once the 4.16 backport PR merges the backport can be automated again.

/jira backport release-4.18,release-4.17,release-4.16

PR #1147, opened 2024-10-24
IR-490: manifests: add pvc related alerts to prometheus rules

flavianmissi:
alerts when detecting metrics for read-only filesystem and device out of space errors.

the alerts in this commit follow the conventions specified by https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md.

---

Please see the description of https://github.com/openshift/image-registry/pull/412 for steps on configuring the image registry with a read-only pvc.

### DIFF
--- a/manifests/09-prometheus-rules.yaml
+++ b/manifests/09-prometheus-rules.yaml
@@ -5,8 +5,31 @@ metadata:
   namespace: openshift-image-registry
   annotations:
     capability.openshift.io/name: ImageRegistry
-    release.openshift.io/delete: "true"
     include.release.openshift.io/hypershift: "true"
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+spec:
+  groups:
+    - name: pvc-problem-detector.rules
+      rules:
+      - alert: ImageRegistryStorageReadOnly
+        for: 10m
+        expr: sum without(instance, pod, operation) (rate(imageregistry_storage_errors_total{code="READ_ONLY_FILESYSTEM"}[5m])) > 0
+        labels:
+           kubernetes_operator_part_of: image-registry
+           severity: warning
+        annotations:
+           summary: The image registry storage is read-only and no images will be committed to storage.
+           description: The image registry storage is read-only. Read-only storage affects direct pushes to the image registry, and pull-through proxy caching. In the case of pull-through proxy caching, read-only storage is particularly important because without it the image registry won't be actually caching anything. Please verify your backing storage solution and make sure the volume mounted on the image-registry pods is writable to avoid potential outages.
+           message: The image registry storage is read-only and no images will be committed to storage.
+      - alert: ImageRegistryStorageFull
+        for: 10m
+        expr: sum without(instance, pod, operation) (rate(imageregistry_storage_errors_total{code="DEVICE_OUT_OF_SPACE"}[5m])) > 0
+        labels:
+           kubernetes_operator_part_of: image-registry
+           severity: warning
+        annotations:
+           summary: The image registry storage disk is full and no images will be committed to storage.
+           description: The image registry storage disk is full. A full disk affects direct pushes to the image registry, and pull-through proxy caching. In the case of pull-through proxy caching, disk space is particularly important because without it the image registry won't be actually caching anything. Please verify your backing storage solution and make sure the volume mounted on the image-registry pods have enough free disk space to avoid potential outages.
+           message: The image registry storage disk is full and no images will be committed to storage.


manifests/09-prometheus-rules.yaml
```diff
@@ -10,3 +10,27 @@ metadata:
...
+        labels:
+           namespace:
+           severity: warning
+        annotations:
+           summary: The image registry storage is read-only and no images will be committed to storage.
+           description: The image registry storage is read-only. Read-only storage affects direct pushes to the image registry, and pull-through proxy caching. In the case of pull-through proxy caching, read-only storage is particularly important because without it the image registry won't be actually caching anything. Please verify your backing storage solution and make sure the volume mounted on the image-registry pods is writable to avoid potential outages.
+           message: The image registry storage is read-only and no images will be committed to storage.
+      - alert: ImageRegistryStorageFull
+        for: 10m
+        expr: imageregistry_storage_errors_total imageregistry_storage_errors_total {code="DEVICE_OUT_OF_SPACE"}
```
wewang58, line 29: Seems duplicated expr:imageregistry_storage_errors_total

flavianmissi: good catch! copy/pasta fail 😂

flavianmissi: fixed

flavianmissi:
`e2e-aws-operator` failure seems like a flake 🤔 
`TestImageRegistryGracefulShutdown` failed with:
```
imageregistry.go:552: registry config resource version was updated when it should have been stable, went from 81238 to 81503
```

/retest

simonpasquier (commented):
As a follow-up, you might want to consider writing a runbook to help users identifying and fixing the root cause (https://github.com/openshift/runbooks/tree/master/alerts).

manifests/09-prometheus-rules.yaml
```diff
@@ -5,8 +5,31 @@ metadata:
...
     include.release.openshift.io/ibm-cloud-managed: "true"
     include.release.openshift.io/self-managed-high-availability: "true"
     include.release.openshift.io/single-node-developer: "true"
+spec:
+  groups:
+    - name: pvc-problem-detector.rules
+      rules:
+      - alert: ImageRegistryStorageReadOnly
+        for: 10m
+        expr: imageregistry_storage_errors_total {code="READ_ONLY_FILESYSTEM"}
```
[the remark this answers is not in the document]
flavianmissi: The metric is indeed a counter ([source](https://github.com/openshift/image-registry/blob/master/pkg/dockerregistry/server/metrics/prometheus.go?plain=1#L126-L134)).

> If the metric is a counter, you need a rate() (or increase()) otherwise the alert will never go back inactive when the problem is fixed.

This makes sense. With your suggestion will the alert reset after 5m of the metric not being present? I'm still a bit confused by these time filters 😅 

> Otherwise the alert would reset everytime a new pod is rolled out.

Ah that explains why I saw the alert "fix itself" after I fixed the storage (which forced a new pod to be rolled out). This was a total hack - I'll incorporate your suggestion.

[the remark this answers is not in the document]
flavianmissi: Additionally, I've set `for` to `10m`, but the `expr` you suggest has `5m`. Should these times match? 🤔

manifests/09-prometheus-rules.yaml
```diff
@@ -5,8 +5,31 @@ metadata:
...
+        labels:
+           kubernetes_operator_part_of: image-registry
+           severity: warning
+        annotations:
+           summary: The image registry storage is read-only and no images will be committed to storage.
+           description: The image registry storage is read-only. Read-only storage affects direct pushes to the image registry, and pull-through proxy caching. In the case of pull-through proxy caching, read-only storage is particularly important because without it the image registry won't be actually caching anything. Please verify your backing storage solution and make sure the volume mounted on the image-registry pods is writable to avoid potential outages.
+           message: The image registry storage is read-only and no images will be committed to storage.
+      - alert: ImageRegistryStorageFull
+        for: 10m
+        expr: imageregistry_storage_errors_total {code="DEVICE_OUT_OF_SPACE"}
```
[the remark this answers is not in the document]
flavianmissi: done!

flavianmissi:
Saw this firing after updated `expr` with Simon's suggestion (see screenshot). I lost my cluster before I could observe it clear up though. @wewang58 since the metric changed, would you mind doing your tests again? thank you!

Screenshot 2024-11-04 at 15 21 31

flavianmissi:
Failure was unrelated to the registry.
/test e2e-aws-ovn-image-registry

manifests/09-prometheus-rules.yaml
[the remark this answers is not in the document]
simonpasquier: `for: 10m` means that the alert will go firing (meaning delivered to Alertmanager) only after being active for 10 minutes. The `[5m]` range is the interval on which the rate is applied.

simonpasquier (commented):
/lgtm

flavianmissi:
FTR I created a spike to track the research work for writing a runbook for the new alerts, as suggested: https://issues.redhat.com/browse/IR-508

wewang58:
> Saw this firing after updated `expr` with Simon's suggestion (see screenshot). I lost my cluster before I could observe it clear up though. @wewang58 since the metric changed, would you mind doing your tests again? thank you!
> 
> Screenshot 2024-11-04 at 15 21 31

@flavianmissi Ok, I will test it today.

wewang58:
@flavianmissi I tested ImageRegistryStorageReadOnly alert, it got expected result

flavianmissi:
Thanks for testing this again @wewang58! The alert becomes "Pending" when the metric in `expr` is being observed, but it will only become "Firing" once 10 minutes is passed as specified by the `for: 10m` in the alert yaml. You might need to adjust your tests so that you trigger the metric for more than 10 minutes, then the alert will fire. 
Does that make sense?

wewang58:
@flavianmissi Actually, the metrics and alert just exists about 5 minutes, and cannot get them anymore, at the same time, registry storage is still Readonly, I tried to call api:

flavianmissi:
What happens if you try another write operation @wewang58? Do you mean the alert itself disappears? this would be very strange!

wewang58:
I can get the firing alert, I think the pr can be merged.
image

flavianmissi:
Thanks @wewang58! I've asked Eric and Steven for help with px and docs approval. The PR will merge once they have tagged it.

manifests/09-prometheus-rules.yaml
```diff
@@ -5,8 +5,31 @@ metadata:
...
+        labels:
+           kubernetes_operator_part_of: image-registry
+           severity: warning
+        annotations:
+           summary: The image registry storage is read-only and no images will be committed to storage.
+           description: The image registry storage is read-only. Read-only storage affects direct pushes to the image registry, and pull-through proxy caching. In the case of pull-through proxy caching, read-only storage is particularly important because without it the image registry won't be actually caching anything. Please verify your backing storage solution and make sure the volume mounted on the image-registry pods is writable to avoid potential outages.
+           message: The image registry storage is read-only and no images will be committed to storage.
+      - alert: ImageRegistryStorageFull
+        for: 10m
+        expr: sum without(instance, pod, operation) (rate(imageregistry_storage_errors_total{code="DEVICE_OUT_OF_SPACE"}[5m])) > 0
```
sferich888, line 28: Do we need a warning for a storage threshold? IE: your at 80% of your storage (or some percentage; configurable in a CR somewhere); that triggers warning before there are issues?

flavianmissi: cross-commenting from slack:
You are right, that would be a better alert. We don't really have any metrics to alert on that though.
The original ask was to alert on read-only storage, the storage full alert is there by request of the PM to alert on any catastrophic error that we could easily detect with current metrics.
I'd prefer to go with what we have in the PR, or to remove the storage full alert entirely. We'd still be covering the customer request by delivering only the read-only alert.
What do you think?

sferich888:
I filed https://issues.redhat.com/browse/IR-511 to track my request for additional metrics. 

/label px-approved

manifests/09-prometheus-rules.yaml
simonpasquier: PVs getting full should be covered by the `KubePersistentVolumeFillingUp` alerting rules.
 https://github.com/openshift/cluster-storage-operator/blob/ad3c5b5427d86ce2dcfdf83498290bec9482b5e5/manifests/12_prometheusrules.yaml#L74-L118

PR #1155, opened 2024-11-06
HOSTEDCP-1994: Add filewatcher for Azure client certificate authentication

bryan-cox:
Updates and moves the function watching the Azure client certificate file for changes to its own file.


go.mod
```diff
@@ -50,7 +50,6 @@ require (
 	k8s.io/apimachinery v0.30.1
 	k8s.io/client-go v0.30.1
 	k8s.io/klog/v2 v2.120.1
-	k8s.io/kubernetes v1.30.2
```
flavianmissi, line 53: Always nice to get rid of some deps, thanks!

flavianmissi (commented):
Hi @bryan-cox, 

The code looks cleaner and it's easier to understand after your changes, thanks for spending the time improving it.

I left you this question on slack, but will write it here too for posterity: Could you expand on what exactly didn't work before? and of the changes you made, which one(s) fixed the bits that weren't working?

Additionally, as a follow-up to this PR, would you be willing to give a shot at unit testing `WatchFileForChanges`? It might not be straightforward to test, and you may have to make some code changes to make it feasible, but it would make potential future refactoring easier. 

WDYT?

---

For next time: it's difficult to review specific changes to `WatchFileForChanges`  because the entire function also moved to another file in the same commit. When that happens what I like to do is first move the code (untouched) to the new location, make a commit, then do the refactor in a different commit. This will make actual changes to code easier to see instead of just showing up with no useful diff. I had to use `vimdiff` here 😄

bryan-cox:
Basically the old file watcher code just wasn't working. I was testing out the same code with our control plane operator (CPO). The previous file watcher code was not catching the changes when I was testing this function against the CPO. I could exec into the pod and see the file changed but the pod was not restarted nor did I see any messages in the logs about the file changing.

re: moving the function to another file, I will make a note of this for next time. Thanks for the feedback!

xiuwang:
Do regression test on azure workload identity and normal azure cluster, no issue found.
Will check the MI feature when got a hypershift cluster enabled workload identity.

PR #1161, opened 2024-11-25
[release-4.17] OCPBUGS-44861: Use Client Cert Auth for ARO HCP deployments  

bryan-cox:
This PR is a manual cherrypick of the following PRs
- https://github.com/openshift/cluster-image-registry-operator/pull/1131
- https://github.com/openshift/cluster-image-registry-operator/pull/1155


bryan-cox:
/close

Closing for now since it looks like ARO HCP will private preview with 4.18.0 rather than 4.17.z.

PR #1168, opened 2024-12-20
OCPBUGS-47503: Power VS: Revert #1076 to match new API validation

mjturek:
After https://github.com/openshift/api/pull/2076, the validation in the image registry operator for the Power VS platform does not match the API's expectations.  This PR reverts #1076 which intended to work around the API's initial validation issues.


mjturek:
whoops sorry @r4f4 wrong PR!

PR #1177, opened 2025-02-17
OCPBUGS-50950: ensure that storage names don't end in dashes

BraeTroutman:
to ensure compatibility with all cloud region names, specifically the longer region names used in AWS GovCloud.

### DIFF
--- a/pkg/storage/util/util.go
+++ b/pkg/storage/util/util.go
@@ -119,6 +119,9 @@ func GenerateStorageName(listers *regopclient.StorageListers, additionalInfo ...
 
 	// Check the length and pad or truncate as needed
 	switch {
+	case len(name) == 61:
+		// if the name is 61 chars, we don't have room for a dash before the padding
+		name = name + string(byte(97+rand.Intn(25)))
 	case len(name) < 62:
 		padding := 62 - len(name) - 1
 		bytes := make([]byte, padding)
@@ -126,10 +129,10 @@ func GenerateStorageName(listers *regopclient.StorageListers, additionalInfo ...
 			bytes[i] = byte(97 + rand.Intn(25)) // a=97 and z=97+25
 		}
 		name = fmt.Sprintf("%s-%s", name, string(bytes))
-	case len(name) > 62:
+	case len(name) >= 62:
 		name = name[0:62]
-		if strings.HasSuffix(name, "-") {
-			name = name[0:61] + string(byte(97+rand.Intn(25)))
+		if before, hasDash := strings.CutSuffix(name, "-"); hasDash {
+			name = before + string(byte(97+rand.Intn(25)))
 		}
 	}
 
--- a/pkg/storage/util/util_test.go
+++ b/pkg/storage/util/util_test.go
@@ -88,6 +88,11 @@ func TestGenerateStorageName(t *testing.T) {
 			infraName:      "MY-INFRA",
 			additionalInfo: []string{"test1", "test2"},
 		},
+		{
+			name:           "govcloud region name",
+			infraName:      "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
+			additionalInfo: []string{"us-gov-west-1"},
+		},
 	} {
 		t.Run(tt.name, func(t *testing.T) {
 			l := regopclient.StorageListers{Infrastructures: MockInfrastructureLister{
@@ -122,6 +127,10 @@ func TestGenerateStorageName(t *testing.T) {
 
 			if n != strings.ToLower(n) {
 				t.Errorf("name should not contain upper case: %s", n)
+			}
+
+			if strings.HasSuffix(n, "-") {
+				t.Errorf("name should not end in a dash: %s", n)
 			}
 		})
 	}


ricardomaraschini (approved):

flavianmissi:
`hypershift-e2e-ask` is failing in other places too: https://search.dptools.openshift.org/?search=failed+to+assign+Azure+Red+Hat+OpenShift+Azure+Files+Storage+Operator+Role+role+to+service+principal&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

I pinged Hypershift folks about it: https://redhat-external.slack.com/archives/C01C8502FMM/p1739868778265999

pkg/storage/util/util.go
```diff
@@ -133,5 +133,9 @@ func GenerateStorageName(listers *regopclient.StorageListers, additionalInfo ...
 		}
 	}
 
+	// ensure that the name does not end in a trailing dash
+	if truncatedName, endsWithDash := strings.CutSuffix(name, "-"); endsWithDash {
```
flavianmissi, line 137: Isn't this already done in line 131?

flavianmissi: After reading the bug again I notice that the problematic name (`2gt95p48ekt031m3qvnpec63eva7hm7f-image-registry-us-gov-west-1-`) has exactly 62 characters. The switch handles `len(name) < 62` and `len(name) > 62`, so it overlooks `len(name) == 62`.
Let's get that fixed instead of adding similar code down below?

BraeTroutman: Totally that makes sense! 
I think didn't add enough information around that example name, my bad-- that bucket name above, `gt95p48ekt031m3qvnpec63eva7hm7f-image-registry-us-gov-west-1-`, is the resulting bucket name _after_ truncating the full name, which is presumably longer than 62 characters. It happens to be that the final (62nd) character after truncating is a dash, which leads to the invalid bucket name

BraeTroutman: so the additional code at the bottom is meant to handle the edge case where a bucket name is longer than 62 characters, and the 62nd character is a dash and therefore truncation would lead to a name that ends in a dash

flavianmissi: > so the additional code at the bottom is meant to handle the edge case where a bucket name is longer than 62 characters, and the 62nd character is a dash and therefore truncation would lead to a name that ends in a dash

Yeah I get that, but you don't really need a new check for that. The `case len(name) > 62` already does what this block is doing.

Here's what I mean, with a slight modification the current code handles this case: https://go.dev/play/p/R-tbSI7gQJy

BraeTroutman: Ah ok yes, I understand now, sorry you had to hold my hand to get me there XD
Let me try updating making the update, and I'll make sure that the unit test still passes

BraeTroutman: Alright, so I've discovered that I was looking for the bug in the wrong branch-- it's not that the full name was too long and was truncated and that truncation ends in a dash, but rather the full name was too short (61 characters) and the dash was added before appending the padding: https://go.dev/play/p/C_t4RdGI5UO

BraeTroutman: I've updated the branching now to handle the special case of a name of 61 characters, and merged the unchecked "exactly 62 characters" branch in with the "greater than 62 characters" branch

flavianmissi: > Alright, so I've discovered that I was looking for the bug in the wrong branch-- it's not that the full name was too long and was truncated and that truncation ends in a dash, but rather the full name was too short (61 characters) and the dash was added before appending the padding: https://go.dev/play/p/C_t4RdGI5UO

Aha, I also missed this - nice catch!
But so to me that means the `case len(name) < 62` is buggy. It's clearly supposed to handle the length being 61. I'm a sucker for simplicity though (and your current solution is nice and simple), so I'm a bit torn if we should do anything about this or leave it as is.

One other thing that might be relevant is that the new version of this code will append a character directly after the last part of the bucket name, which in S3's case is the region, i.e. `us-gov-west-1`. So when the first code branch executes (`case len(name) == 61`), the bucket name will end in something like `us-gov-west-1f`, which isn't an S3 region. I'm fine with this but it makes me wonder why we want a fixed length of 62. 

I'm going too deep on this though 😅 

@ricardomaraschini do you have an opinion?

flavianmissi:
I'll go ahead and approve this. @ricardomaraschini as you were the original reviewer I'll let the `lgtm` for you.

/approve

pkg/storage/util/util.go
BraeTroutman: I mean yeah I agree there, also the letter after a region number has the added semantics of an availability zone-- I guess since it's just one specific character to the end in this specific case rather than a random one, but I can't think of one that's neutral enough not to imply any further meaning. Maybe `x`?

xiuwang:
@BraeTroutman  @ricardomaraschini We need this label to merge code.
/label acknowledge-critical-fixes-only

PR #1187, opened 2025-03-07
CNTRLPLANE-112: Bump msi-dataplane dependency

bryan-cox:
This pull request bumps the msi-dataplane dependency to the latest version, which contains updated error handling around filewatcher.


bryan-cox:
@xiuwang could I get a QE tag for this PR please?

PR #1188, opened 2025-03-08
AUTH-482: Move required-scc annotation from pruner job to pod

jacobsee:
The `required-scc` annotation should be on the template that gets applied to the pod itself, not the template for the job resource.


flavianmissi:
Hi @jacobsee, how does this change affects users, if at all? I need to know so that I can properly evaluate whether we need to ask PX and Docs approval, or if I can tag the PR myself.

PR #1190, opened 2025-03-12
OSASINFRA-3747: Prefer CA cert from credentials secret

stephenfin:
In openshift/cloud-credential-operator/pull/780, we have added the ability for `cloud-credential-operator` to consume a CA cert from the root credentials secret and to include in the credentials secrets it provisions.
In openshift/installer/pull/9194, we have modified the Installer to start setting this field where necessary.

Adapt `cluster-image-registry-operator` to allow it to start consuming the CA cert from this place. We maintain fallbacks for the previous locations of the cert for now, but these can be removed in the next release.

Dependencies:

- [x] https://github.com/openshift/cloud-credential-operator/pull/780


stephenfin:
/remove-label qe-approved

Let's leave this to a member of ShiftStack QE

PR #1223, opened 2025-06-10
OCPBUGS-51594: bump golang.org/x/oauth2

sbiradar10:
Bump to prevent [CVE-2025-22868 ](https://github.com/advisories/GHSA-3wqc-mwfx-672p). As we can't simply use the upstream package we are using the openshift's fork. We can't use the upstream because we don't have a go v1.23 builder for older openshift versions.


sbiradar10:
@adambkaplan can you please review this ? 
Need one more lgtm for this

PR #1225, opened 2025-06-17
NO-JIRA: Set ownership annotation for image-registry-ca CA bundle

vrutkovs:
Add missing ownership annotations for CA bundles


flavianmissi:
/lgtm

aren't we missing this annotation in the other resources as well though?

/retest

PR #1229, opened 2025-07-15
OCPBUGS-53867: Bump github.com/golang-jwt/jwt

sbiradar10:
Prevent https://github.com/advisories/GHSA-mh63-6h87-95cp. We are relying on
https://github.com/Azure/go-autorest that has been archived and isn't receiving more updates. This will force us to move to a different package but we can't wait for that to happen to address the CVE therefore this PR adds a replacement on github.com/golang-jwt/jwt/v4 pointing to v4.5.2.


shannon:
@sbiradar10 Yea I believe it needs to be run inside the `cmd/move-blobs` directory as well as the root, we missed that piece.

sbiradar10:
@shannon just made changes now

shannon:
/assign flavianmissi

Can you please review/approve this? This was a follow up to #1224 because we missed the second go.mod file.

ricardomaraschini (changes requested):
Can you squash your commits ?

ricardomaraschini (approved):

shannon:
/hold

Hi @sbiradar10, I noticed that v5 version is not quite right. It should be `v5.2.2` not `v5.2.1`.

So I will put a hold on this until it's updated. `go get github.com/golang-jwt/jwt/v5@v5.2.2` followed by vendor and tidy should be sufficient here. Then squash it again.

PR #1253, opened 2025-09-04
CORS-4204: Support Storage Endpoint override

barbacbd:
** Go Get update on openshift/api. github.com/openshift/api v0.0.0-20250806102053-6a7223edb2fc ** Go mod vendor (requires updates to numerous packages).


barbacbd:
Closing in favor of a solution that does not require explicit api endpoint overrides for each new service.

PR #1277, opened 2026-01-13
OCPBUGS-63691: nodecadaemon.yaml: atomically update certificates

flavianmissi:


bindata/nodecadaemon.yaml
```diff
@@ -51,10 +51,18 @@ spec:
                 f=$(echo $f | sed  -r 's/(.*)\.\./\1:/')
                 reg_dir_path="/etc/docker/certs.d/${f}"
                 if [ -e "${reg_dir_path}" ]; then
-                    cp -u $ca_file_path $reg_dir_path/ca.crt
+                    # Use atomic file replacement to prevent readers from seeing
+                    # empty or partial certificates during update. The mv command
+                    # uses rename(2) syscall which atomically replaces the file.
+                    # See: https://man7.org/linux/man-pages/man2/rename.2.html
+                    cp -u $ca_file_path $reg_dir_path/ca.crt.tmp && mv $reg_dir_path/ca.crt.tmp $reg_dir_path/ca.crt
```
ricardomaraschini, line 58: The `-u` flag can be removed here, there won't be anything to compare to before copying.

bindata/nodecadaemon.yaml
```diff
@@ -51,10 +51,18 @@ spec:
...
                 else
                     mkdir $reg_dir_path
-                    cp $ca_file_path $reg_dir_path/ca.crt
+                    # Use atomic file replacement to prevent readers from seeing
+                    # empty or partial certificates during update. The mv command
+                    # uses rename(2) syscall which atomically replaces the file.
+                    # See: https://man7.org/linux/man-pages/man2/rename.2.html
+                    cp $ca_file_path $reg_dir_path/ca.crt.tmp && mv $reg_dir_path/ca.crt.tmp $reg_dir_path/ca.crt
```
ricardomaraschini, line 65: I would argue that we can directly `cp` here, the directory wasn't even there.

ricardomaraschini (changes requested):
Hey @flavianmissi , I am resuming the work on this ticket. Do you think you can address these comments ? If you prefer I can raise a new PR and move from there. Let me know.

bindata/nodecadaemon.yaml
ricardomaraschini: Oh, wait. We should not be copying if the target file is the same date/time. Somehow we need to mimic the `-u` behavior but using `mv`.

ricardomaraschini:
@flavianmissi I decide to port this to Go to have a better control on when/how to copy files. Can you please take a look at https://github.com/openshift/cluster-image-registry-operator/pull/1325 ?

PR #1282, opened 2026-01-28
MON-4483: chore: add permissions on endpointslice to Prometheus Role nd use serviceDiscoveryRole: EndpointSlice in ServiceMonitors

machine424:
This PR migrates Prometheus service discovery from the deprecated Endpoints API to the EndpointSlices API, by:

- Setting `serviceDiscoveryRole: EndpointSlice` on ServiceMonitors.
- Granting Prometheus `endpointslices` permissions.

We're taking a conservative approach by keeping the existing `endpoints` permissions alongside the new `endpointslices` ones. This provides a safety net in case any ServiceMonitors, whether deployed from this repo or from another source, still rely on the same Role and were missed during the migration.

That said, since both resources provide essentially the same data, keeping both isn't meaningfully more permissive from a security standpoint.

**These changes target OpenShift 4.22+ and should not be backported to earlier releases.**

### DIFF
--- a/manifests/0000_90_cluster-image-registry-operator_00_servicemonitor-rbac.yaml
+++ b/manifests/0000_90_cluster-image-registry-operator_00_servicemonitor-rbac.yaml
@@ -59,6 +59,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - discovery.k8s.io
+  resources:
+  - endpointslices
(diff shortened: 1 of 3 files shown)


flavianmissi:
This is not user facing, so I'll apply docs and px labels myself.

/label docs-approved
/label px-approved

PR #1318, opened 2026-04-09
OCPBUGS-83282: Allow longer window before going Degraded

tjungblu:
…istry

Raise the grace period from 1m to 2m before mirroring a false Available condition into Degraded with reason Unavailable. This reduces flakes when the deployment temporarily has no available replicas (e.g. Recreate strategy with slow volume attach on some platforms).

Update syncStatus tests to use LastTransitionTime beyond the new threshold.

Made-with: Cursor


ricardomaraschini (approved):

tjungblu:
/retitle OCPBUGS-83282: Allow longer window before going Degraded

PR #1325, opened 2026-04-22
OCPBUGS-63691: replace node-ca shell script with go implementation

ricardomaraschini:
Replaces the shell script in the `node-ca` DaemonSet with a Go implementation. The old implementation copied files directly, potentially causing partial reads if other processes were reading them at the same time. The new implementation uses "atomic copy" (temp file and then rename) to avoid this race condition.

- Adds `pkg/nodeca` package with `SyncCerts()` function and `node-ca-sync` subcommand.
- Updates `bindata/nodecadaemon.yaml` to use the new go command.
- Makes `pkg/resource/nodecadaemon.go` to use `OPERATOR_IMAGE` instead of `IMAGE` env var.

## Summary by CodeRabbit

* **Refactor**
  * Modernized node-CA certificate synchronization from shell-based polling to integrated operator logic with enhanced error handling and logging.

* **Tests**
  * Added comprehensive test coverage for certificate synchronization and graceful shutdown workflows.

* **Improvements**
  * Enhanced graceful shutdown detection for node-CA operations.

### DIFF
--- a/pkg/resource/nodecadaemon.go
+++ b/pkg/resource/nodecadaemon.go
@@ -60,7 +60,7 @@ func (ds *generatorNodeCADaemonSet) Get() (runtime.Object, error) {
 
 func (ds *generatorNodeCADaemonSet) expected() *appsv1.DaemonSet {
 	daemonSet := resourceread.ReadDaemonSetV1OrDie(assets.MustAsset("nodecadaemon.yaml"))
-	daemonSet.Spec.Template.Spec.Containers[0].Image = os.Getenv("IMAGE")
+	daemonSet.Spec.Template.Spec.Containers[0].Image = os.Getenv("OPERATOR_IMAGE")
 	return daemonSet
 }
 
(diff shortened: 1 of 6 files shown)


pkg/resource/nodecadaemon.go
dusk125, line 1: Is this intentional? It seems otherwise unrelated to the rest of the changes.

ricardomaraschini: Yes, it is intentional. Prior to this we were using the operand image (`IMAGE` env) to run the shell script. Now that the code has been migrated to the operator we need to use the operator image (`OPERATOR_IMAGE` env) instead.  Does that make sense ?

ricardomaraschini:
This PR does not change the product in any way so I will flag this accordingly.

/label px-approved
/label docs-approved

pkg/resource/nodecadaemon.go
dusk125: Yep, thanks! Just wanted to make sure an unintended change wasn't sneaking its way in

ricardomaraschini:
Failure not related. Trying once more.

/restest

ricardomaraschini:
Again, the failure on e2e-hypershift-conformance does not seem to be related to the pr.  Let's retry it once more.

/retest

PR #1329, opened 2026-04-30
OCPBUGS-84725: Fix stale config cache causing incorrect deployment

ropatil010:
Hi Team,

PTAL on this PR. 

Problem:
When patching the image registry config rapidly in succession (e.g., changing storage from S3 to PVC and replicas from 2 to 1), the operator creates deployments using stale cached configuration instead of the latest values from the API server. This results in pods being created with incorrect replica count and storage configuration, causing CrashLoopBackOff.

Symptoms on 4.22:
- Config shows: replicas=1, storage=PVC
- Deployment created with: replicas=2, storage=S3 (stale values)
- Result: 2 pods crash (CrashLoopBackOff) + 1 old pod still running

This works correctly on 4.20 and 4.21, making this a regression in 4.22.

Root Cause:
The bug is triggered by a combination of two caching layers:

1. Informer cache (controller.go:257) - reads config from watch cache
2. Persistent ResourceCache (introduced in 4.22)

In commit 550fc9977 "resource: persist ResourceCache across reconcile loops" (March 27, 2026), a persistent ResourceCache was added to the Generator struct. This change, while beneficial for network policy resources, exposed a race condition when combined with rapid config updates:

- Patch 1: Clear storage, set managementState=Unmanaged
- Patch 2: Set PVC storage, replicas=1, managementState=Managed
- Reconciliation starts before informer cache updates
- List() receives stale config from cache (S3, replicas=2)
- Deployment created with stale values → pods crash

This race condition exists in all versions, but 4.22 is more susceptible due to the persistent cache changing reconciliation timing.

Fix:
Read fresh config directly from the API server at the beginning of List() function (pkg/resource/generator.go) to ensure we always use the latest configuration when creating resources. Falls back to cached config if the API call fails, providing graceful degradation.

This adds one extra API GET call per reconciliation, which is acceptable for correctness.

Testing:
Verified with test case OCP-76854 that reproduces the issue on 4.22. After the fix, deployment is correctly created with replicas=1 and PVC storage configuration.

Related commits:
- 550fc9977 (introduced regression in 4.22)
- OCPBUGS-66225 (fixed stale cache issues in ImageConfigController)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration sync: controller now tries a fresh server fetch with a short timeout and falls back to a cached copy on transient failures.
  * If the remote configuration is missing, the system triggers an immediate bootstrap.
  * Added diagnostic logging to indicate when fresh vs. cached configuration is used.

### DIFF
--- a/pkg/operator/controller.go
+++ b/pkg/operator/controller.go
@@ -254,14 +254,23 @@ func (c *Controller) getRoutes(cr *imageregistryv1.Config) ([]*routev1.Route, er
 }
 
 func (c *Controller) sync() error {
-	cr, err := c.listers.RegistryConfigs.Get(defaults.ImageRegistryResourceName)
+	// OCPBUGS-84725: Read config directly from API server to avoid stale cache issues.
+	// The informer cache can be stale when config is patched in rapid succession,
+	// which combined with multiple deployment updates can lead to ReplicaSets being
+	// created with incorrect replica counts and storage configuration.
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	cr, err := c.clients.RegOp.ImageregistryV1().Configs().Get(
+		ctx, defaults.ImageRegistryResourceName, metaapi.GetOptions{},
+	)
 	if err != nil {
 		if errors.IsNotFound(err) {
 			return c.Bootstrap()
 		}
-		return fmt.Errorf("failed to get %q registry operator resource: %s", defaults.ImageRegistryResourceName, err)
+		return fmt.Errorf("failed to get %q from API server: %w", defaults.ImageRegistryResourceName, err)
 	}
-	cr = cr.DeepCopy() // we don't want to change the cached version
+
 	prevCR := cr.DeepCopy()
 
 	if cr.ObjectMeta.DeletionTimestamp != nil {
--- a/pkg/resource/deployment.go
+++ b/pkg/resource/deployment.go
@@ -213,12 +213,32 @@ func (gd *generatorDeployment) Update(o runtime.Object) (runtime.Object, bool, e
 
 	original, expected := o.(*appsapi.Deployment), exp.(*appsapi.Deployment)
 
-	// XXX if we are removing the affinities from the spec we need to do
-	// that on its own api call so not to mess with scheduling logic.
-	// in other words: already scheduled pods have those anti-affinities
-	// in place and we may hit a scenario where new pods will conflict with
-	// those anti-affinities.
+	// OCPBUGS-66203: If removing affinity while INCREASING replicas, use separate call
+	// to avoid scheduling conflicts where new pods can't schedule due to existing pods'
+	// anti-affinity rules.
+	// OCPBUGS-84725: For all other cases (replica decrease, storage changes, etc.),
+	// use atomic update to prevent intermediate ReplicaSets with stale configuration.
+	needsSeparateAffinityCall := false
 	if original.Spec.Template.Spec.Affinity != nil && expected.Spec.Template.Spec.Affinity == nil {
+		// Affinity is being removed - check if replicas are also increasing
+		origReplicas := int32(1)
+		if original.Spec.Replicas != nil {
+			origReplicas = *original.Spec.Replicas
+		}
+		expReplicas := int32(1)
+		if expected.Spec.Replicas != nil {
+			expReplicas = *expected.Spec.Replicas
+		}
+
+		// Only use separate call if replicas are INCREASING (scheduling conflict risk)
+		// For replica decrease or same count, use atomic update to avoid stale ReplicaSets
+		if expReplicas > origReplicas {
+			needsSeparateAffinityCall = true
+		}
+	}
+
+	if needsSeparateAffinityCall {
+		// OCPBUGS-66203: Remove affinity first to avoid scheduling conflicts with new pods
 		original = original.DeepCopy()
 		original.Spec.Template.Spec.Affinity = nil
 		dep, err := gd.client.Deployments(gd.GetNamespace()).Update(
@@ -230,6 +250,8 @@ func (gd *generatorDeployment) Update(o runtime.Object) (runtime.Object, bool, e
 		return dep, true, nil
 	}
 
+	// OCPBUGS-84725: Apply all changes atomically to prevent stale ReplicaSets
+	// This handles replica decrease, storage changes, and other updates safely
 	dep, updated, err := resourceapply.ApplyDeployment(
 		context.TODO(), gd.client, gd.eventRecorder, exp.(*appsapi.Deployment), gd.LastGeneration(),
 	)


ropatil010:
// Execution results

Cluster built using: build 4.23,openshift/cluster-image-registry-operator#1329
Logs to get the Payload: https://storage.googleapis.com/test-platform-results/logs/release-openshift-origin-installer-launch-aws-modern/2051145156031156224/build-log.txt
Use the payload build the cluster from flexy profile
```

oc get pvc image-registry-test-claim-76854 -n openshift-image-registry -w
NAME                              STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
image-registry-test-claim-76854   Pending                                      gp3-csi        <unset>                 26s
image-registry-test-claim-76854   Pending                                      gp3-csi        <unset>                 6m8s
image-registry-test-claim-76854   Pending                                      gp3-csi        <unset>                 6m8s
image-registry-test-claim-76854   Pending   pvc-a2725569-2fb2-4c8b-bca9-f17549a94462   0                         gp3-csi        <unset>                 6m11s
image-registry-test-claim-76854   Bound     pvc-a2725569-2fb2-4c8b-bca9-f17549a94462   5Gi        RWO            gp3-csi        <unset>                 6m11s

oc get pods -n openshift-image-registry -l docker-registry=default -w
NAME                              READY   STATUS    RESTARTS   AGE
image-registry-7c688755d6-6bjt7   1/1     Running   0          146m
image-registry-7c688755d6-cjz6n   1/1     Running   0          146m
image-registry-7c688755d6-6bjt7   1/1     Terminating   0          151m
image-registry-7c688755d6-6bjt7   1/1     Terminating   0          151m
image-registry-7c688755d6-cjz6n   1/1     Terminating   0          151m
image-registry-7c688755d6-cjz6n   1/1     Terminating   0          151m
image-registry-7c688755d6-6bjt7   0/1     Completed     0          151m
image-registry-7c688755d6-cjz6n   0/1     Completed     0          151m
image-registry-59b8c9cc76-xqz7d   0/1     Pending       0          0s
image-registry-59b8c9cc76-xqz7d   0/1     Pending       0          0s
image-registry-7c688755d6-6bjt7   0/1     Completed     0          151m
image-registry-7c688755d6-6bjt7   0/1     Completed     0          151m
image-registry-7c688755d6-cjz6n   0/1     Completed     0          151m
image-registry-7c688755d6-cjz6n   0/1     Completed     0          151m
image-registry-59b8c9cc76-xqz7d   0/1     Pending       0          3s
image-registry-59b8c9cc76-xqz7d   0/1     Pending       0          4s
image-registry-59b8c9cc76-xqz7d   0/1     ContainerCreating   0          4s
image-registry-59b8c9cc76-xqz7d   0/1     ContainerCreating   0          14s
image-registry-59b8c9cc76-xqz7d   0/1     Running             0          14s

oc get pods -n openshift-image-registry -l docker-registry=default
NAME                              READY   STATUS    RESTARTS   AGE
image-registry-59b8c9cc76-xqz7d   1/1     Running   0          41s

oc logs -n openshift-image-registry image-registry-59b8c9cc76-xqz7d --tail=20 | grep -i "error\|s3\|credential" || echo "No S3 or credential errors found ✅"
No S3 or credential errors found ✅

oc describe deployment -n openshift-image-registry image-registry | grep -A30 "^Events:"
Events:
  Type    Reason             Age    From                   Message
  ----    ------             ----   ----                   -------
  Normal  ScalingReplicaSet  161m   deployment-controller  Scaled up replica set image-registry-7c688755d6 from 0 to 2
  Normal  ScalingReplicaSet  10m    deployment-controller  Scaled down replica set image-registry-7c688755d6 from 2 to 1
  Normal  ScalingReplicaSet  10m    deployment-controller  Scaled down replica set image-registry-7c688755d6 from 1 to 0
  Normal  ScalingReplicaSet  9m44s  deployment-controller  Scaled up replica set image-registry-59b8c9cc76 from 0 to 1

  Comparison: Bug vs Fix                                                                                                                                                                                            
                                                                                                                                                                                                                    
  ┌───────────────────────┬────────────────────────────────┬────────────────────────────┐                                                                                                                           
  │        Metric         │ Buggy Behavior (OCPBUGS-84725) │ Fixed Behavior (This Test) │                                                                                                                           
  ├───────────────────────┼────────────────────────────────┼────────────────────────────┤
  │ Deployment updates    │ 2 separate updates             │ 1 atomic update ✅         │                                                                                                                           
  ├───────────────────────┼────────────────────────────────┼────────────────────────────┤                                                                                                                           
  │ ReplicaSets created   │ 2 (stale + correct)            │ 1 (correct only) ✅        │                                                                                                                           
  ├───────────────────────┼────────────────────────────────┼────────────────────────────┤                                                                                                                           
  │ Generation increments │ 2 → 3 → 4                      │ 1 → 2 ✅                   │                                                                                                                           
  ├───────────────────────┼────────────────────────────────┼────────────────────────────┤
  │ Storage config in pod │ S3 (stale)                     │ filesystem (correct) ✅    │
  ├───────────────────────┼────────────────────────────────┼────────────────────────────┤                                                                                                                           
  │ Pod errors            │ S3 credential errors           │ None ✅                    │                                                                                                                           
  ├───────────────────────┼────────────────────────────────┼────────────────────────────┤                                                                                                                           
  │ Final replicas        │ 2 pods CrashLoopBackOff        │ 1 pod Running ✅           │                                                                                                                           
  └───────────────────────┴────────────────────────────────┴────────────────────────────┘
```
/assign @flavianmissi @ricardomaraschini  PTAL for the fix on this bug.

ropatil010:
/assign  @ricardomaraschini @flavianmissi 
PTAL on the bug fix.

ricardomaraschini:
@andreacv98 I need you to look into this. This cache has been introduced recently and you have more context on it.

andreacv98:
@ropatil010 @ricardomaraschini overall this fix looks good to me and it shouldn't bother the resource cache mechanism introduced with the network policies. I would only take a look at [this comment](https://github.com/openshift/cluster-image-registry-operator/pull/1329/changes#r3167650525), but I'd say it's fine.

/lgtm

pkg/operator/controller.go
```diff
@@ -261,7 +261,34 @@ func (c *Controller) sync() error {
...
+	// OCPBUGS-84725: Fetch fresh config from API server to avoid stale cache issues.
+	// This ensures we use the latest config when rapidly updating replicas/storage.
+	// The informer cache can be stale when config is patched in rapid succession,
+	// leading to deployment creation with incorrect replica counts and storage config.
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	freshCR, err := c.clients.RegOp.ImageregistryV1().Configs().Get(
+		ctx, defaults.ImageRegistryResourceName, metaapi.GetOptions{},
+	)
```
ricardomaraschini, line 266: Why are we reading from the lister on line `257` above just to read from the API here ? Can't we read from the API directly instead ?

pkg/resource/deployment.go
```diff
@@ -213,12 +214,29 @@ func (gd *generatorDeployment) Update(o runtime.Object) (runtime.Object, bool, e
...
-	// in other words: already scheduled pods have those anti-affinities
-	// in place and we may hit a scenario where new pods will conflict with
-	// those anti-affinities.
+	// OCPBUGS-84725: Check if affinity removal is the ONLY spec change.
+	// If other fields are changing simultaneously (replicas, storage, strategy, etc.),
+	// we must batch all changes into a single update to prevent race conditions where:
+	//   1. First update removes affinity -> K8s creates ReplicaSet with stale config
+	//   2. Second update changes storage/replicas -> ReplicaSet already exists with wrong spec
+	// This caused pods to crash with S3 credential errors when switching to PVC storage.
+	affinityOnlyChange := false
```
ricardomaraschini: Would you mind explaining how this affinity is impacting the issue on "stalled cache" ?

ricardomaraschini (changes requested):

pkg/resource/deployment.go
ropatil010: The Problem:                                                                                                                                                                               
  The original code always removed affinity in a separate API call (to avoid pod scheduling conflicts). However, when affinity removal coincides with other changes (storage, replicas), this
   creates a two-step update:                                                                                                                                                                
  1. First update removes affinity → Kubernetes creates ReplicaSet                                                                                                                         
  2. Second update applies storage/replicas changes                                                                                                                                          
                                                                                                                                                                                           
  If the controller has stale cached config (the bug we're fixing), step 1 causes Kubernetes to create the ReplicaSet using the OLD cached values (e.g., S3 storage). Then step 2 updates the
   deployment spec, but the ReplicaSet already exists with incorrect configuration, causing pods to crash.                                                                                   
   
  The Fix:                                                                                                                                                                                   
  The new logic detects whether affinity is the only field changing:                                                                                                                       
  - If yes: Keep the original behavior (separate affinity update to avoid scheduling conflicts)                                                                                              
  - If no: Batch all changes (affinity + storage + replicas) into a single atomic update (lines 239-252)
                                                                                                                                                                                             
  This prevents Kubernetes from creating intermediate ReplicaSets with partial/stale configuration.                                                                                          
   
  The affinity fix is a defensive measure that works together with the fresh API read - even if we get fresh config, we need atomic updates to prevent race conditions during the deployment 
  update itself.

ricardomaraschini: How the intermediate ReplicaSet (the one that got the affinities removed) has stale configuration ?

ropatil010: Hi @ricardomaraschini

here is the more explanation
 The Problematic Flow (Before the Fix)                                                                                                                                                                                                                                                                                                                                                
  1. User patches the config: Changes storage S3→PVC + replicas 2→1                                                                                                                          
  2. Controller's sync() fetches fresh config from API server (lines 264-284 in controller.go)                                                                                             
  3. First deployment update (in deployment.go's old code):                                                                                                                                  
    - Operator detects affinity needs removal (because replicas 2→1)                                                                                                                         
    - Makes a separate API call to remove affinity ONLY                                                                                                                                      
    - Returns immediately after this update                                                                                                                                                  
  4. Kubernetes deployment controller reacts:                                                                                                                                                
    - Sees the deployment spec changed (generation 2→3)                                                                                                                                      
    - Creates ReplicaSet RIGHT NOW using the current deployment spec in etcd                                                                                                                 
    - Problem: The deployment spec at this moment still has S3 storage and replicas=2 because the operator hasn't applied those changes yet!                                               
    - ReplicaSet created: image-registry-548b84b979 with stale S3 config                                                                                                                     
  5. Next reconciliation loop:                                                                                                                                                               
    - Operator tries to apply storage/replica changes (generation 3→4)                                                                                                                       
    - Too late! ReplicaSet already exists with wrong spec                                                                                                                                    
    - Result: Pods crash with S3 credential errors    

Key Point                                                                                                                                                                                  
  The "staleness" isn't about the controller's cache being stale (that's fixed by the fresh API read in controller.go). The staleness is that the deployment object in etcd itself temporarily has a partially updated spec between the two operator updates. When the intermediate ReplicaSet is created after update #1, it captures this incomplete state from the deployment spec.

The Fix                                                                                                                                                                                    
  By batching all changes into a single atomic update when multiple fields change together, Kubernetes only sees one deployment change and creates only one ReplicaSet with the complete, correct configuration.  

Here is the execution results on 4.20/4.21 and buggy 4.22 clusters https://redhat.atlassian.net/browse/OCPBUGS-84725. 

With fix here is the execution result: https://github.com/openshift/cluster-image-registry-operator/pull/1329#issuecomment-4373026720

pkg/operator/controller.go
ropatil010: Fixed per your suggestions.

ropatil010:
Tested with new fix as per your suggestions.
```
oc get clusterversion
NAME      VERSION                                                AVAILABLE   PROGRESSING   SINCE   STATUS
version   4.23.0-0-2026-05-08-044806-test-ci-ln-9kgkj4b-latest   True        False         112s    Cluster version is 4.23.0-0-2026-05-08-044806-test-ci-ln-9kgkj4b-latest

oc get pods -n openshift-image-registry -l docker-registry=default
NAME                             READY   STATUS    RESTARTS   AGE
image-registry-977888695-vbbns   1/1     Running   0          58s

oc get pvc image-registry-test-claim-76854 -n openshift-image-registry -w
NAME                              STATUS    VOLUME   CAPACITY   ACCESS MODES   STORAGECLASS   VOLUMEATTRIBUTESCLASS   AGE
image-registry-test-claim-76854   Pending                                      gp3-csi        <unset>                 76s
image-registry-test-claim-76854   Pending                                      gp3-csi        <unset>                 5m35s
image-registry-test-claim-76854   Pending                                      gp3-csi        <unset>                 5m35s
image-registry-test-claim-76854   Pending   pvc-e6f2fdfb-336a-4219-80a6-69b54f1f6f55   0                         gp3-csi        <unset>                 5m37s
image-registry-test-claim-76854   Bound     pvc-e6f2fdfb-336a-4219-80a6-69b54f1f6f55   5Gi        RWO            gp3-csi        <unset>                 5m37s
```

pkg/resource/deployment.go
ricardomaraschini: Let's assume that `affinityOnlyChange` was set to `true` on  line `232`. What is the difference if it is updated on line `242` or in line `252` ? Isn't the outcome the same ? Isn't this going to bring back the bug for which we had a comment ?

ropatil010: When `affinityOnlyChange = true`, both paths produce the same result - a single deployment update.
My New solution: Remove the `affinityOnlyChange` detection entirely and always use the atomic `ApplyDeployment` path. Will this works?

ricardomaraschini: My point is that it does not matter if it is `true` or `false`, the result is always the same (we always update the whole thing). There was a comment your patch is removing:

```
	// XXX if we are removing the affinities from the spec we need to do
	// that on its own api call so not to mess with scheduling logic.
	// in other words: already scheduled pods have those anti-affinities
	// in place and we may hit a scenario where new pods will conflict with
	// those anti-affinities.
```

We were specifically removing the affinities in a separate call. Giving the comment we had a reason to do so. The commit that introduces this change must point to an old Bugzilla (now Jira). You need to check if your change isn't going to reintroduce it. If it doesn't then I am OK with merging this.

ropatil010: Fixed.
OCPBUGS-66203: If removing affinity while INCREASING replicas, use separate call
to avoid scheduling conflicts where new pods can't schedule due to existing pods' anti-affinity rules.	
OCPBUGS-84725: For all other cases (replica decrease, storage changes, etc.),
use atomic update to prevent intermediate ReplicaSets with stale configuration.

ricardomaraschini: This makes sense to me. Have you tested if we aren't hitting `OCPBUGS-66203` ?

ropatil010: Hi @ricardomaraschini Tested both scenarios with this fix
```
   ========================================                                                                                                                                                                       
     SCENARIO 1: OCPBUGS-66203                                                                                                                                                                                      
     Replica Increase Test (2→3)                                                                                                                                                                                    
     ========================================                                                                                                                                                                     
     Test: Replica increase 2→3 with affinity removal                                                                                                                                                               
     Expected: 2 deployment updates (separate affinity call)                                                                                                                                                        
     Result: 2 deployment update(s)                                                                                                                                                                                 
     Pods: 3/3 running                                                                                                                                                                                              
                        
     [SUCCESS] ✅ SCENARIO 1 PASS: 2 deployment updates detected                                                                                                                                                    
     [SUCCESS]    Separate affinity call was used (OCPBUGS-66203 fix working)                                                                                                                                       
     [INFO] Waiting for deployment to have 3 ready replicas (timeout: 300s)...                                                                                                                                    
     ........[SUCCESS] Deployment ready with 3 replicas                                                                                                                                                             
     [SUCCESS] ✅ All 3 pods running (no scheduling conflict)                                                                                                                                                       
     [SUCCESS] ✅ Anti-affinity successfully removed   
                                                                                                                                                                                                                  
  Great! Scenario 1 passed with 2 deployment updates as expected. Now I need to:                                                                                                                                    
  1. Clean up the cluster back to original state                                                                                                                                                                    
  2. Run scenario 2         

    ========================================                                                                                                                                                                       
     SCENARIO 2: OCPBUGS-84725                                                                                                                                                                                      
     Replica Decrease + Storage Change                                                                                                                                                                              
     ========================================  
     [SUCCESS] ✅ SCENARIO 2 PASS: Single deployment update (1 generation)                                                                                                                                          
     [SUCCESS]    Atomic update was used (OCPBUGS-84725 fix working)                                                                                                                                                
     [INFO] Waiting for deployment to have 1 ready replicas (timeout: 300s)...                                                                                                                                      
     .....[SUCCESS] Deployment ready with 1 replicas                                                                                                                                                                
     [SUCCESS] ✅ Pod image-registry-7c775c5bb-sm8tg: Using PVC storage (image-registry-test-claim-1778242485)                                                                                                      
     [SUCCESS] ✅ No crashing pods detected       
```

ricardomaraschini: Thanks for the due diligence. I don't think I have anything else to add, super.

PR #1341, opened 2026-05-07
NO-JIRA: fix flaky TestGlobalTLSCopy unit test

ricardomaraschini:
the fake client stored objects in-memory without json round-tripping, leaving RawExtension.Object set instead of converting to .Raw like the real api server does. ConfigObserver's UpdateOperatorSpec sets .Object, then ApplyOperatorStatus converts it to .Raw via server-side apply's json marshaling. The test raced on whether the registry controller read before or after the conversion. Add marshal/unmarshal in the reactor to match production behavior. if it read before we have a failure, if it read after a success.

to better understand the problem, the snippet below prints:
```
true false
false true
```

```go
package main

import (
        "encoding/json"
        "fmt"

        imageregistryv1 "github.com/openshift/api/imageregistry/v1"
        corev1 "k8s.io/api/core/v1"
        "k8s.io/apimachinery/pkg/runtime"
)

func main() {
        var config imageregistryv1.Config

        config.Spec.ObservedConfig = runtime.RawExtension{Object: &corev1.Namespace{}}
        fmt.Println(config.Spec.ObservedConfig.Raw == nil, config.Spec.ObservedConfig.Object == nil)

        data, err := json.Marshal(config)
        if err != nil {
                panic(err)
        }

        config = imageregistryv1.Config{}
        if err := json.Unmarshal(data, &config); err != nil {
                panic(err)
        }
        fmt.Println(config.Spec.ObservedConfig.Raw == nil, config.Spec.ObservedConfig.Object == nil)

}
```

## Summary by CodeRabbit

* **Tests**
  * Enhanced tests to better simulate Kubernetes API server normalization, fail fast on serialization issues, and more accurately exercise the tracker/resource-version update path.

### DIFF
--- a/pkg/operator/controller_test.go
+++ b/pkg/operator/controller_test.go
@@ -2,6 +2,7 @@ package operator
 
 import (
 	"context"
+	"encoding/json"
 	"strconv"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/operator/controller_test.go
```diff
@@ -65,6 +66,17 @@ func registryConfigResourceVersionBumper(t *testing.T, regcli *imageregistryfake
 		update := action.(clientgotesting.UpdateAction)
 		newcfg := update.GetObject().(*imageregistryv1.Config).DeepCopy()
 
+		// simulate real api server behavior: marshal/unmarshal to
+		// convert RawExtension.Object to RawExtension.Raw.
```
vincentdephily: Might be worth expanding a bit here, to point out that we sometimes run after update but before apply, which explains why we flakily get a `.Object` instead of a `.Raw`.

ricardomaraschini: Done, please take a look.

pkg/operator/controller_test.go
```diff
@@ -65,6 +66,24 @@ func registryConfigResourceVersionBumper(t *testing.T, regcli *imageregistryfake
...
+		// the object passing through the wire before hiting the
+		// storage.
+		data, err := json.Marshal(newcfg)
+		if err != nil {
+			t.Fatalf("failed to marshal config: %v", err)
+		}
+		newcfg = &imageregistryv1.Config{}
+		if err := json.Unmarshal(data, newcfg); err != nil {
+			t.Fatalf("failed to unmarshal config: %v", err)
+		}
```
ingvagabund, line 85: Also worth checking `.Raw` is not empty to make this more explicit.

ricardomaraschini: There is a time in which both .Raw and .Object are expected to be empty.

ricardomaraschini:
Unit tests are passing locally, let's try one more time.

/retest

PR #1344, opened 2026-05-14
OCPBUGS-99257: Improve GCS KMS encryption error handling and documentation

barbacbd:
Fixed inconsistent error handling in KMS encryption code where googleapi.Error was treated as fatal but other errors were not. Now both error types are handled consistently as best-effort, matching the S3 implementation pattern.

Added comprehensive documentation explaining:
- Encryption is best-effort and failures don't fail storage creation
- Data remains encrypted with Google-managed keys if customer KMS fails
- KMS encryption only applied to operator-created buckets
- User-provided existing buckets must have encryption configured externally

The KMS encryption feature has been functional since 2019 (commit 8c08cc6805). This change improves error handling consistency and documentation clarity.

## Summary by CodeRabbit

* **Bug Fixes**
  * Apply KMS encryption settings during creation of operator-managed GCS buckets for more reliable encryption and explicit logging of the configured KMS key.
  * Set the "encrypted" status based on whether a KMS key is configured (removing post-creation update dependency).

* **Documentation**
  * Clarify that encryption is best-effort and only applied to operator-created buckets; user-supplied buckets are not modified.

### DIFF
--- a/pkg/storage/gcs/gcs.go
+++ b/pkg/storage/gcs/gcs.go
@@ -231,7 +231,6 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 
 	// If a bucket name is supplied, and it already exists and we can access it
 	// just update the config
-	var bucket *gstorage.BucketHandle
 	var bucketExists bool
 	var bucketCreated bool
 	if len(d.Config.Bucket) != 0 {
@@ -249,7 +248,6 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 		}
 	}
 	if len(d.Config.Bucket) != 0 && bucketExists {
-		bucket = gclient.Bucket(d.Config.Bucket)
 		if cr.Spec.Storage.ManagementState == "" {
 			cr.Spec.Storage.ManagementState = imageregistryv1.StorageManagementStateUnmanaged
 		}
@@ -265,7 +263,7 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 			}
 		}
 		bucketAttrs := gstorage.BucketAttrs{Location: d.Config.Region}
-		bucket = gclient.Bucket(d.Config.Bucket)
+		bucket := gclient.Bucket(d.Config.Bucket)
 
 		labels, err := getUserLabels(d.Listers.Infrastructures)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


barbacbd:
This was executed using the cluster bot to create a cluster. The cluster came up and upon deep investigation

`gsutil ls -L` shows actual state of the bucket rather than `gcloud storage describe`.

pkg/storage/gcs/gcs.go
```diff
@@ -274,6 +274,16 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 		klog.V(1).Infof("createStorage: %v list of labels will be applied to %s bucket", labels, d.Config.Bucket)
 		bucketAttrs.Labels = labels
 
+		// Set KMS encryption at bucket creation time if configured.
+		// This is more efficient and reliable than creating the bucket first and updating it later,
+		// as it avoids requiring storage.buckets.update permission.
```
ricardomaraschini: > as it avoids requiring storage.buckets.update permission.

This seems to contradict the change on `manifests/01-registry-credentials-request-gcs.yaml`. We are giving more permissions on the latter, no ? Can you clarify ?

manifests/01-registry-credentials-request-gcs.yaml
```diff
@@ -23,6 +23,7 @@ spec:
       - storage.buckets.delete
       - storage.buckets.get
       - storage.buckets.list
+      - storage.buckets.update
```
ricardomaraschini: As per private discussions: you mentioned this is needed because we are updating buckets somewhere else. Could you elaborate more on this topic ? I mean, I can't understand how such "bucket updates somewhere else" were working without this.

barbacbd: There are two ways to approach this, it is add this permission or remove the redundant `bucket.Update` call. This call _could_ be redundant since encryption setting is at bucket creation time. If this ever changes in the future and encryption is not only at creation time then you will need this permission. I guess the question is should encryption setting only happen at creation time (and this means I should remove the bucket.Update call) or should we add this permission and all setting changes to buckets?

ricardomaraschini: How was the now "redundant bucket.Update call" working before without this permission ?

barbacbd: It could be that it was a silent failure (certainly not fatal), or it is a rarely used feature since KeyID is optional and that is required to hit this code I believe.

barbacbd: It looks like the more appropriate change is to not change permissions but to remove the bucket.Update call since we probably shouldn't change/add encryption.

ricardomaraschini: That is interesting. How have you seen it fail ?

barbacbd: I didn't actually. This was brought up from an analysis after I set the key information on creation. Claude analyzed this and determined that this is an issue that has actually been around since october 2023 when it was originally released.

barbacbd: I confirm that this is an issue without this PR:

barbacbd: Confirmed that this is an issue:
1. Create a cluster
2. Update the image registry bucket. 
3. No failure observed
4. Check the bucket using gcloud and the encryption was not used/set.

pkg/storage/gcs/gcs.go
```diff
@@ -249,7 +249,6 @@ func (d *driver) CreateStorage(cr *imageregistryv1.Config) error {
 		}
 	}
 	if len(d.Config.Bucket) != 0 && bucketExists {
-		bucket = gclient.Bucket(d.Config.Bucket)
```
ricardomaraschini, line 252: I think we don't need the `bucket` variable in a higher scope now.

ricardomaraschini (approved):

ricardomaraschini (approved):

barbacbd:
/verified by @barbacbd 

manual testing

PR #1347, opened 2026-06-03
OCPBUGS-73768: bugfix: podtemplatespec: update topology spread constraints

everettraven:
to honor node affinity and node taints in skew calculations to prevent deadlocks in deploying the image-registry pods on nodes that are worker nodes but are considered unschedulable because of other taints (like being edge nodes).

additionally, ensure skew calculations only take into account the current pod template revision to prevent deadlocks associated with rolling out a new revision of the deployment because of stale pods being included in skew calculations.

## Summary by CodeRabbit

* **Refactor**
  * Improved pod topology spread constraint configuration for more consistent/default behavior.

* **Tests**
  * Updated topology spread constraint test expectations to match the refactor.
  * Adjusted certificate sync test setup and expectations to cover copying and removal scenarios.

### DIFF
--- a/pkg/nodeca/nodeca_test.go
+++ b/pkg/nodeca/nodeca_test.go
@@ -746,12 +746,6 @@ func TestSyncCerts(t *testing.T) {
 					if err := os.WriteFile(filepath.Join(src, "new..90"), []byte("new"), 0o644); err != nil {
 						t.Fatal(err)
 					}
-					if err := os.Mkdir(filepath.Join(dst, "keep:80"), 0o755); err != nil {
-						t.Fatal(err)
-					}
-					if err := os.WriteFile(filepath.Join(dst, "keep:80", "ca.crt"), []byte("keep"), 0o644); err != nil {
-						t.Fatal(err)
-					}
 					if err := os.Mkdir(filepath.Join(dst, "remove:70"), 0o755); err != nil {
 						t.Fatal(err)
 					}
--- a/pkg/resource/podtemplatespec.go
+++ b/pkg/resource/podtemplatespec.go
@@ -457,40 +457,16 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
 	// some bare metal cluster nodes will not include the zone labels, in
 	// which case we just omit the related constraint.
 	// we constraint scheduling to workers because we want to reduce the
-	// scope of scheduling to workers only. we need this constraint
-	// because tainted nodes (such as control plane nodes) are not excluded
-	// from skew calculations, so if we don't limit scheduling to workers
-	// the maxSkew won't allow more than one pod to be scheduled per node.
+	// scope of scheduling to workers only.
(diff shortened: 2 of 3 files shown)


pkg/resource/podtemplatespec.go
```diff
@@ -457,40 +457,16 @@ func makePodTemplateSpec(coreClient coreset.CoreV1Interface, proxyLister configl
...
-		{
-			MaxSkew:           1,
-			TopologyKey:       "node-role.kubernetes.io/worker",
-			WhenUnsatisfiable: corev1.DoNotSchedule,
-			LabelSelector: &metav1.LabelSelector{
-				MatchLabels: defaults.DeploymentLabels,
-			},
-		},
+		topologySpreadConstraintForTopologyKey("kubernetes.io/hostname"),
+		topologySpreadConstraintForTopologyKey("node-role.kubernetes.io/worker"),
```
everettraven, line 465: Open question: While it doesn't hurt to have this here, the fact that the topology spread constraints now use the `Honor` policy for node affinity and node taints this topology key is likely unnecessary to keep around because skew should only ever be calculated against nodes that the image registry pods are able to be scheduled on. I kept this as-is to reduce the amount of change that is happening here, but it should be safe to remove. For those that may be more familiar with the intentions behind adding a spread constraint based on this topology key, is there any reasons I may be missing for why we _should not_ remove this constraint?

pkg/nodeca/nodeca_test.go
```diff
@@ -746,12 +746,6 @@ func TestSyncCerts(t *testing.T) {
 					if err := os.WriteFile(filepath.Join(src, "new..90"), []byte("new"), 0o644); err != nil {
 						t.Fatal(err)
 					}
-					if err := os.Mkdir(filepath.Join(dst, "keep:80"), 0o755); err != nil {
```
ricardomaraschini, line 749: This seems unrelated to the PR. Can you add some context on why this has been removed ?  Running locally, using `main`, works fine for me here.

everettraven: This was related to unit test failures that I was seeing both locally and in CI.

The run that failed for this PR: https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-image-registry-operator/1347/pull-ci-openshift-cluster-image-registry-operator-main-unit/2062198024473415680

This same failure appears for me when running locally after checking out the `main` branch and making sure it is up-to-date.

My understanding of the sync logic is that because we end up intentionally creating the destination file it has a newer modified timestamp resulting in it not being synced as expected (it is skipped instead).

I chose to assume that the expected counts were more indicative of intent of the test case so removing these lines seemed like the correct approach here.

ricardomaraschini: Alright, let's get this in. I will look closer to this "flake" and if needed I will open a follow up PR. This got me a little worried.

PR #1350, opened 2026-06-10
OCPBUGS-66225: Fix image-registry Degraded=True blips during upgrades

liouk:
Fixes two Degraded=True blips during OCP upgrades caught by the CVO monitor test:

  - **Unavailable (vSphere):** Increase degraded inertia from 2 to 5 minutes. On platforms using Recreate rollout with ReadWriteOnce PVC storage, CSI volume detach/reattach takes 2-4 minutes, exceeding the previous threshold.
  - **NodeCADaemonController (Azure):** Add 3-second degraded inertia. Transient API server errors (e.g. etcd timeouts) were immediately setting Degraded=True, even when cleared within milliseconds on the next sync.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevents brief sync glitches from immediately marking the node CA daemon as degraded; the controller now tolerates short failures and only reports degraded after the issue persists.
  * Avoids premature "unavailable/degraded" flips after short interruptions, reducing false-positive alerts and improving status stability.

* **Tests**
  * Adds unit tests validating inertia behavior and condition transitions to prevent regressions.

### DIFF
--- a/pkg/operator/nodecadaemon.go
+++ b/pkg/operator/nodecadaemon.go
@@ -27,6 +27,10 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/resource"
 )
 
+// nodeCADaemonControllerDegradedInertia is how long sync must keep failing (since the first
+// failure after the last success) before we set NodeCADaemonControllerDegraded.
+const nodeCADaemonControllerDegradedInertia = 2 * time.Minute
+
 type NodeCADaemonController struct {
 	eventRecorder   events.Recorder
 	appsClient      appsv1client.AppsV1Interface
@@ -34,8 +38,9 @@ type NodeCADaemonController struct {
 	daemonSetLister appsv1listers.DaemonSetNamespaceLister
 	serviceLister   corev1listers.ServiceNamespaceLister
 
-	cachesToSync []cache.InformerSynced
-	queue        workqueue.TypedRateLimitingInterface[any]
+	cachesToSync     []cache.InformerSynced
+	queue            workqueue.TypedRateLimitingInterface[any]
+	syncFailureSince time.Time
 }
 
 func NewNodeCADaemonController(
@@ -164,6 +169,21 @@ func (c *NodeCADaemonController) sync() error {
 
 	err = resource.ApplyMutator(gen)
 	if err != nil {
+		now := time.Now()
+		if c.syncFailureSince.IsZero() {
+			c.syncFailureSince = now
+		}
+		if now.Sub(c.syncFailureSince) < nodeCADaemonControllerDegradedInertia {
+			klog.V(2).Infof("NodeCADaemonController: sync error within degraded inertia window (%s), not reporting degraded yet: %v", nodeCADaemonControllerDegradedInertia, err)
+			_, _, updateError := v1helpers.UpdateStatus(
+				ctx,
+				c.operatorClient,
+				v1helpers.UpdateConditionFn(availableCondition),
+				v1helpers.UpdateConditionFn(progressingCondition),
+			)
+			return utilerrors.NewAggregate([]error{err, updateError})
+		}
+
 		_, _, updateError := v1helpers.UpdateStatus(
 			ctx,
 			c.operatorClient,
@@ -179,6 +199,7 @@ func (c *NodeCADaemonController) sync() error {
 		return utilerrors.NewAggregate([]error{err, updateError})
 	}
 
+	c.syncFailureSince = time.Time{}
 	_, _, err = v1helpers.UpdateStatus(
 		ctx,
 		c.operatorClient,
--- a/pkg/operator/status.go
+++ b/pkg/operator/status.go
@@ -21,6 +21,8 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/metrics"
 )
 
+const unavailableDegradedInertia = 5 * time.Minute
+
 func updateCondition(cr *imageregistryv1.Config, condtype string, condstate operatorapiv1.OperatorCondition) {
 	found := false
 	conditions := []operatorapiv1.OperatorCondition{}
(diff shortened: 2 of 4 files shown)


pkg/operator/status.go
```diff
@@ -349,7 +349,7 @@ func (c *Controller) syncStatus(
 		operatorDegraded.Reason = "Removed"
 	} else if operatorAvailable.Status != operatorapiv1.ConditionTrue {
 		updatedAvailableCondition := v1helpers.FindOperatorCondition(cr.Status.Conditions, operatorapiv1.OperatorStatusTypeAvailable)
-		if updatedAvailableCondition != nil && time.Since(updatedAvailableCondition.LastTransitionTime.Time) > 2*time.Minute {
+		if updatedAvailableCondition != nil && time.Since(updatedAvailableCondition.LastTransitionTime.Time) > 5*time.Minute {
```
liouk: If reviewers think that this inertia is too long for all cases, we can instead make inertia rollout-aware and set it to 5 minutes only during rollouts, using `isDeploymentStatusComplete()`.

pkg/operator/nodecadaemon.go
```diff
@@ -27,15 +27,20 @@ import (
 	"github.com/openshift/cluster-image-registry-operator/pkg/resource"
 )
 
+// nodeCADaemonControllerDegradedInertia is how long sync must keep failing (since the first
+// failure after the last success) before we set NodeCADaemonControllerDegraded.
+const nodeCADaemonControllerDegradedInertia = 3 * time.Second
```
liouk: Retries have an exponential backoff starting at 5ms via `AddRateLimited() -- therefore 3s should be plenty of time for any transient error to resolve, while being low enough to not delay degradation for too long.

I'm open to suggestions about this value though, as my experience with cluster-image-registry-operator is very limited.

liouk:
@coderabbitai review

liouk:
Testing the upgrade job for 5.0 is sufficient because CVO will rollout the new cluster-image-registry-operator first therefore the changes will be there when the new operand version gets rolled out.

Using the two jobs mentioned in the corresponding jira as failing:
/payload-job periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-vsphere-ovn-upgrade periodic-ci-openshift-release-main-ci-5.0-e2e-azure-ovn-upgrade

liouk:
Launching a payload aggregate given that the observed issues with Azure are intermittent in nature.

/payload-aggregate periodic-ci-openshift-release-main-ci-5.0-e2e-azure-ovn-upgrade 5

pkg/operator/nodecadaemon.go
```diff
@@ -164,6 +169,14 @@ func (c *NodeCADaemonController) sync() error {
...
 	err = resource.ApplyMutator(gen)
 	if err != nil {
+		now := time.Now()
+		if c.syncFailureSince.IsZero() {
+			c.syncFailureSince = now
+		}
+		if now.Sub(c.syncFailureSince) < nodeCADaemonControllerDegradedInertia {
+			klog.V(2).Infof("NodeCADaemonController: sync error within degraded inertia window (%s), not reporting degraded yet: %v", nodeCADaemonControllerDegradedInertia, err)
+			return err
+		}
```
everettraven, line 185: I noticed that this controller doesn't look to have any tests.

I don't want to unnecessarily hold this up since this a blocker, but I think it would be useful to have some kind of unit tests in place here to ensure we don't regress this behavior unintentionally somehow.

I'll leave it up to you what you'd like to do here.

pkg/operator/nodecadaemon.go
```diff
@@ -164,6 +169,14 @@ func (c *NodeCADaemonController) sync() error {
 
 	err = resource.ApplyMutator(gen)
 	if err != nil {
+		now := time.Now()
+		if c.syncFailureSince.IsZero() {
+			c.syncFailureSince = now
+		}
+		if now.Sub(c.syncFailureSince) < nodeCADaemonControllerDegradedInertia {
+			klog.V(2).Infof("NodeCADaemonController: sync error within degraded inertia window (%s), not reporting degraded yet: %v", nodeCADaemonControllerDegradedInertia, err)
+			return err
```
everettraven: Purely returning here looks like it would mean that any changes to the `Available` and `Progressing` conditions wouldn't be applied?

I think we would still want any of our observed states for those conditions to still be persisted.

pkg/operator/nodecadaemon.go
everettraven: IMO, 3 seconds feels quite short. Maybe for one-off errors this is fine, but for something that might cause request timeouts for a period of time (but still relatively short), this would likely cause the degraded state to happen on the very next request that times out.

From what I've seen in various other operators, ~5 minutes seems like a fairly reasonable amount of time to rule out most self-resolving transient issues and go degraded.

If we are concerned about going _too_ long here, maybe we settle on something like ~2 minutes to start with so that something like ~4 requests (assuming they've all gone the full ~30s timeout period) can happen before we decide to go degraded?

I'm not strongly opinionated here and am happy to start with whatever folks feel is reasonable and adjust as needed.

pkg/operator/status.go
everettraven: I think 5 minutes is fine.

everettraven: Nit: make the "inertia" here a constant? Would help make future updates automatically reflected in the unit tests here as well.

pkg/operator/nodecadaemon.go
liouk: Happy to go with 2min as well -- my initial mindset was to stay as close as possible to the existing behavior, but the 30s timeout is a good call. I'll bump this to 2min.

pkg/operator/nodecadaemon.go
liouk: Added some unit tests with the help of claude :+1:

ricardomaraschini (approved):

everettraven (approved):
/lgtm

liouk:
/payload-job periodic-ci-openshift-release-main-ci-5.0-upgrade-from-stable-4.22-e2e-vsphere-ovn-upgrade periodic-ci-openshift-release-main-ci-5.0-e2e-azure-ovn-upgrade

PR #1354, opened 2026-07-13
CORENET-7355: Remove redundant ingress rule from image-registry netpol

kyrtapz:
The namespaceSelector: {} peer is a strict subset of the ports-only rule (no `from` field) which already allows ingress from all sources on port 5000. It causes the CNI to maintain a cluster-wide pod IP address set that is recomputed on every pod event.

Keep the monitoring and router ingress rules as they document distinct intent.

## Summary by CodeRabbit

* **Bug Fixes / Access Changes**
  * Updated the image registry NetworkPolicy so TCP port **5000** ingress is permitted from **any source** for image pull/push. Previously, access was constrained by namespace/selector-based ingress rules.

* **Tests**
  * Updated e2e NetworkPolicy assertions for the image registry policy on **TCP/5000** to expect **universal allow-all ingress**, removing the prior namespace-scoped expectations.

### DIFF
--- a/bindata/image-registry-networkpolicy.yaml
+++ b/bindata/image-registry-networkpolicy.yaml
@@ -32,14 +32,8 @@ spec:
     - port: 5000
       protocol: TCP
 
-  # Allow ingress from all namespaces for image pull/push
-  - from:
-      - namespaceSelector: {}
-    ports:
-    - port: 5000
-      protocol: TCP
-
-  # Allow ingress from host network pods (nodes, kubelet)
+  # Allow ingress from all pod-network and host-network pods for image pull/push.
+  # Side effect: this rule does not restrict traffic by source.
   - ports:
     - port: 5000
       protocol: TCP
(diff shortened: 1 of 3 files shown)


kyrtapz:
@coderabbitai your auto generated pr summary is wrong, fix it.

bindata/image-registry-networkpolicy.yaml
```diff
@@ -14,32 +14,7 @@ spec:
...
-
-  # Allow ingress from all namespaces for image pull/push
-  - from:
-      - namespaceSelector: {}
-    ports:
-    - port: 5000
-      protocol: TCP
-
-  # Allow ingress from host network pods (nodes, kubelet)
+  # Allow ingress from all sources for image pull/push
```
danwinship: So 2 general principles of the NP work:

1. It's better to have redundant rules when that indicates the intent better.
2. If you do squash rules, or if you're forced to write a rule that allows more than you actually want to allow, then there needs to be a clear comment indicating the intent.

So in this case, I think it's fine to merge the "all namespaces" and "host network pods" rules together to avoid inefficiency in OVN-K, but the comment should still say something like "all pod-network and host-network pods", because the goal _isn't_ to allow ingress from outside the cluster here, that's just a side effect.

I'd be inclined to leave the separate monitoring and router rules, since they have a different intent than the generic pod rules.

kyrtapz: Thanks, I guess that makes sense.
I've updated the change to reflect that, ptal.

ricardomaraschini (approved):

PR #1356, opened 2026-07-21
CORS-4520: GCP Universe Domain Support

patrickdillon:
Sets the universe domain which is required in any GCP environment that does not use the default universe domain; particularly, Google Cloud Dedicated, the sovereign cloud offering. 

`WithAuthCredentialsJSON` takes a credential type parameter, which can be sed to limit accepted credentials types, which is useful for validating credentials provided by external third parties. For OpenShift, cluster credentials are coming from first party users running clusters in their project, so we do not need to limit which credentials are accepted; therefore we just pass the type through from the credential to the function. 

This commit also updates the deprecated `CredentialsFromJSON` to `CredentialsFromJSONWithType`, which has the same type pattern as `WithAuthCredentialsJSON`.

See https://github.com/openshift/enhancements/pull/1977 for more details. 

## Summary by CodeRabbit

* **New Features**
  * Added universe-domain-aware handling for Google credential JSON across storage and tagging.
  * Tag-binding and storage resource naming now include the universe domain.
  * Automatically disables redirect behavior when non-default universe-domain credentials are detected.
* **Bug Fixes**
  * Improved “bucket not found” handling using robust error matching.
* **Tests**
  * Updated GCS and tag-binding tests with a complete service-account fixture and new universe-domain expectations.
* **Chores**
  * Refreshed Go module dependencies, including cloud/auth/telemetry libraries.


pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -318,10 +319,23 @@ func getTagClientOptions(listers *regopclient.StorageListers, endpoint string) (
...
+		return nil, fmt.Errorf("failed to parse credentials JSON: %w", err)
+	}
 	opts := []option.ClientOption{
-		option.WithCredentialsJSON([]byte(cfg.KeyfileData)),
+		option.WithAuthCredentialsJSON(option.CredentialsType(f.Type), []byte(cfg.KeyfileData)),
 		option.WithEndpoint(endpoint),
 	}
+	if f.UniverseDomain != "" {
+		opts = append(opts, option.WithUniverseDomain(f.UniverseDomain))
+	}
```
barbacbd, line 343: With the bumped SDK (v0.247.0+), `WithEndpoint` overrides `WithUniverseDomain` for endpoint resolution, but the SDK now also validates that the universe domain matches the endpoint. On GDC, `WithUniverseDomain` would be set to the GDC universe domain while `WithEndpoint` still points at `cloudresourcemanager.googleapis.com` (hardcoded in `resourceManagerHostSubPath` on line 56). This mismatch could cause the SDK to reject the request with a universe domain validation error.

Similarly, `bucketParentPathFmt` (line 59) hardcodes `storage.googleapis.com`, which would not be correct on GDC.

This is a pre-existing issue — the tag client was never GDC-aware — but with `WithUniverseDomain` being conditionally added here, it may surface as a new runtime error on GDC rather than silently hitting the wrong endpoint. Worth considering whether these constants should be parameterized by the universe domain, or if tag bindings on GDC is out of scope for this PR.

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -318,10 +319,23 @@ func getTagClientOptions(listers *regopclient.StorageListers, endpoint string) (
...
+	if err := json.Unmarshal([]byte(cfg.KeyfileData), &f); err != nil {
+		return nil, fmt.Errorf("failed to parse credentials JSON: %w", err)
+	}
 	opts := []option.ClientOption{
-		option.WithCredentialsJSON([]byte(cfg.KeyfileData)),
+		option.WithAuthCredentialsJSON(option.CredentialsType(f.Type), []byte(cfg.KeyfileData)),
 		option.WithEndpoint(endpoint),
 	}
+	if f.UniverseDomain != "" {
+		opts = append(opts, option.WithUniverseDomain(f.UniverseDomain))
```
barbacbd: The universe domain here is extracted by parsing the raw JSON field, while `gcs.go` uses `creds.GetUniverseDomain()` on a `Credentials` object. These behave differently:

- `GetUniverseDomain()` always returns a value (falls back to `googleapis.com`), and on GCE it queries the metadata server if the JSON field is absent.
- The raw JSON parse returns empty string if the field is missing, which means `WithUniverseDomain` is skipped entirely.

This means on GCE where the metadata server provides the universe domain but the credentials JSON doesn't have the field, `gcs.go` would pick it up and `gcp_labels_tags.go` would miss it. Consider using `CredentialsFromJSONWithType` + `GetUniverseDomain()` here too for consistency with the storage client.

pkg/storage/gcs/gcs.go
```diff
@@ -63,12 +65,31 @@ func (d *driver) getGCSClient() (*gstorage.Client, error) {
...
+
+	// Parse the credential type to allow restricting which credential types are
+	// accepted from external sources. In this case, there are no restrictions
+	// so we simply pass the type through.
+	var f struct {
+		Type string `json:"type"`
+	}
+	if err := json.Unmarshal(jsonCreds, &f); err != nil {
+		return nil, fmt.Errorf("failed to parse credentials JSON: %w", err)
+	}
```
ricardomaraschini, line 106: Can we have this logic isolated on a function ? It is repeated twice, here and in `gcp_labels_tags.go`. I can't help but feeling sorry we have to do this `Unmarshal()` just to get the type :-)

pkg/storage/gcs/gcp_labels_tags.go
```diff
@@ -312,46 +315,65 @@ func (t *tagServiceManager) addTagsToStorageBucket(ctx context.Context, cr *imag
...
-func getTagClientOptions(listers *regopclient.StorageListers, endpoint string) ([]option.ClientOption, error) {
+func getTagClientOptions(listers *regopclient.StorageListers, location string) ([]option.ClientOption, string, error) {
 	cfg, err := GetConfig(listers)
 	if err != nil {
-		return nil, fmt.Errorf("failed to read GCS configuration for creating tag client: %w", err)
+		return nil, "", fmt.Errorf("failed to read GCS configuration for creating tag client: %w", err)
 	}
 
+	authJSON := []byte(cfg.KeyfileData)
+	creds, err := google.CredentialsFromJSON(context.Background(), authJSON)
```
ricardomaraschini: My linter says that `google.CredentialsFromJSON is deprecated`. I guess if we move the `json.Unmarshal()` earlier we can then use `CredentialsFromJSONWithType()` thus avoiding this deprecated function use.

pkg/storage/gcs/gcs.go
```diff
@@ -52,6 +52,24 @@ func NewDriver(ctx context.Context, c *imageregistryv1.ImageRegistryConfigStorag
...
 }
 
+// HasNonDefaultUniverseDomain returns true if the GCS credentials
+// are configured for a non-default universe domain (e.g. sovereign clouds).
+func HasNonDefaultUniverseDomain(listers *regopclient.StorageListers) bool {
+	cfg, err := GetConfig(listers)
+	if err != nil {
+		return false
+	}
+	creds, err := goauth2.CredentialsFromJSON(context.Background(), []byte(cfg.KeyfileData), gstorage.ScopeFullControl)
```
ricardomaraschini: Here I am seeing the same deprecated function in use.

ricardomaraschini (commented):
This looks good. I have left a few comments but none of those would be a blocker. Let me hear what you think.

_edit_: I just now noticed that the linter is failing on this PR as well. We need to address that deprecated function use.

pkg/storage/gcs/gcp_labels_tags.go
patrickdillon: fixed

pkg/storage/gcs/gcp_labels_tags.go
patrickdillon: fixed

pkg/storage/gcs/gcs.go
patrickdillon: Yes done!

ricardomaraschini:
@patrickdillon TLDR:
```
test/e2e/gcs_test.go:82:22: SA1019: goauth2.CredentialsFromJSON is deprecated: This function is deprecated because of a potential security risk. It does not validate the credential configuration. The security risk occurs when a credential configuration is accepted from a source that is not under your control and used without validation on your side. (staticcheck)
	credentials, err := goauth2.CredentialsFromJSON(ctx, []byte(gcscfg.KeyfileData), gstorage.ScopeFullControl)
```

patrickdillon:
@ricardomaraschini woops missed the deprecated function was hanging around in the test. Fixed now

ricardomaraschini:
Thanks @patrickdillon. I am glad to approve it. When it comes to px-approved and docs-approved tags:  maybe @tlwu2013 can help us with that. My understanding is that we only need an OK that the product is aware of the feature and that it is properly documented.

/approve
/lgtm

patrickdillon:
/verified by multi-pr testing 

https://prow.ci.openshift.org/view/gs/test-platform-results/logs/multi-pr-openshift-cloud-provider-gcp-131-openshift-gcp-pd-csi-driver-126-openshift-cluster-image-registry-operator-1356-openshift-cloud-network-config-controller-249-openshift-machine-api-provider-gcp-180-openshift-image-registry-474-openshift-cluster-cloud-controller-manager-operator-493-openshift-installer-10733-openshift-installer-10731-e2e-gcd-ovn-private-techpreview/2082899620337291264

PR #1363, opened 2026-08-21
OCPBUGS-112551: imageconfig: Preserve ImageStreamImportMode during upgrade race

tchap:
During an upgrade from a pre-5.x release, there is a brief window where the config-operator has already enabled the ImageStreamImportMode feature gate in featuregate/cluster but the CVO has not yet reconciled with those gates. Until the CVO picks up the change it actively clears Status.Desired.Architecture to "". syncImageStatus() was treating an empty architecture as a signal to write "Legacy", which is incorrect for multi-arch clusters and triggered a spurious double rollout of the openshift-apiserver deployment, ultimately causing the Available condition to flip False for one second.

Fix by extracting the import-mode decision into a pure helper imageStreamImportMode() and preserving the existing status value when Architecture is empty. The ClusterVersion informer will re-trigger the sync once the CVO has written the correct architecture. On any settled 5.x cluster with ImageStreamImportMode in the Default feature set, Architecture is always populated and the empty branch is unreachable.

Add unit tests covering all input combinations for the helper.

This actually doesn't have any effect on 5.1 as the field is always set, we need to backport to 5.0.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image stream import-mode selection based on configured settings and architecture.
  * Preserved the existing import mode when architecture information is unavailable.
  * Applied appropriate defaults for multi-architecture and single-architecture images.

* **Tests**
  * Added coverage for configured overrides, architecture-based defaults, and empty architecture values.

### DIFF
--- a/pkg/operator/imageconfig.go
+++ b/pkg/operator/imageconfig.go
@@ -187,22 +187,11 @@ func (icc *ImageConfigController) syncImageStatus() error {
 		modified = true
 	}
 	if icc.imageStreamImportModeEnabled {
-		var importmode configapi.ImportModeType
-		if cfg.Spec.ImageStreamImportMode != "" {
-			importmode = cfg.Spec.ImageStreamImportMode
-		} else {
-			cv, err := icc.clusterVersionLister.Get("version")
-			if err != nil {
-				return err
-			}
-			// If the clusterversion reports that the desired architecture (existing or desired) of the
-			// cluster is "Multi", set import mode to PreserveOriginal. Else set it to Legacy
-			if cv.Status.Desired.Architecture == configv1.ClusterVersionArchitectureMulti {
-				importmode = configapi.ImportModePreserveOriginal
-			} else {
-				importmode = configapi.ImportModeLegacy
-			}
+		cv, err := icc.clusterVersionLister.Get("version")
+		if err != nil {
+			return err
 		}
+		importmode := imageStreamImportMode(cfg.Spec.ImageStreamImportMode, cfg.Status.ImageStreamImportMode, cv.Status.Desired.Architecture)
 		if cfg.Status.ImageStreamImportMode != importmode {
 			cfg.Status.ImageStreamImportMode = importmode
 			modified = true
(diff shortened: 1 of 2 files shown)


ricardomaraschini (commented):
This looks correct but I have a couple of questions so I can better understand. Is the `ImageStreamImportMode` feature gate now enabled by default on 5.0 ?

> Until the CVO picks up the change it actively clears Status.Desired.Architecture to "".

Can you clarify why is it clearing it out ? What was the value set in there that is cleared ?

tchap:
@ricardomaraschini 

> Is the `ImageStreamImportMode` feature gate now enabled by default on 5.0?

Yes, it is now enabled by default.

> Can you clarify why is it clearing it out? What was the value set in there that is cleared?

The CVO has an internal flag `StatusReleaseArchitecture` that gates whether it writes `Status.Desired.Architecture` at all. This flag is enabled by `ImageStreamImportMode` being present in the enabled list of `featuregate/cluster`. When the flag is **off**, the CVO runs this on every reconciliation:

```go
if !enabledGates.StatusReleaseArchitecture() {
    desired.Architecture = configv1.ClusterVersionArchitecture("")
}
```

The 4.22 CVO ran with `ImageStreamImportMode` disabled (the gate was only promoted to `Default` in 5.x), so `Status.Desired.Architecture` was always `""` throughout the 4.22 lifetime.

The race this PR fixes is:

1. The 5.0 `config-operator` starts and writes the 5.0 feature gate set to `featuregate/cluster`, enabling `ImageStreamImportMode`.
2. The 5.0 `image-registry-operator` starts, reads the updated gates, sees `imageStreamImportModeEnabled=true`, and immediately calls `syncImageStatus()`.
3. At this point the 5.0 CVO has not yet reconciled with the new gate set — it is still actively clearing `Architecture` to `""`.
4. `syncImageStatus()` reads `""`, falls into the `default` branch, and writes `"Legacy"` to `image.config/cluster` status — wrong for a multi-arch cluster.
5. When the 5.0 CVO reconciles and enables `StatusReleaseArchitecture`, it starts writing `"Multi"` to `Architecture`. The image-registry-operator then re-syncs and correctly writes `"PreserveOriginal"`.

The fix makes `syncImageStatus()` treat `Architecture == ""` as "not yet known, hold steady" rather than "not Multi, so Legacy".

ricardomaraschini:
Thanks for the clarification.

> syncImageStatus() reads "", falls into the default branch, and writes "Legacy" to image.config/cluster status — wrong for a multi-arch cluster.

How does a blip on "Legacy" triggers a "spurious double rollout of the openshift-apiserver deployment" ? I think this is what I am missing here.

tchap:
@ricardomaraschini 

> How does a blip on "Legacy" trigger a "spurious double rollout of the openshift-apiserver deployment"?

`image.config.openshift.io/cluster` Status is not only written by this operator — it is also **read** by `cluster-openshift-apiserver-operator`. That operator has a config observer `ObserveImagestreamImportMode` which watches `image.config/cluster` and propagates `status.imageStreamImportMode` into the openshift-apiserver's own operator config.

The observed config is merged into the `config` ConfigMap in `openshift-apiserver` as `config.yaml` — the actual config file mounted into every apiserver pod. When the ConfigMap content changes, library-go's revision controller detects the diff, increments `LatestAvailableRevision`, and the workload sync updates the Deployment's pod template `revision` label. That label change is what Kubernetes sees as a spec change and uses to trigger the rolling update.

ricardomaraschini (commented):
/approve
/lgtm

